PR #13, opened 2019-12-30
karapace: add schema in subject check endpoint

vtainio:


karapace/karapace.py
```diff
@@ -390,6 +391,27 @@ def send_delete_subject_message(self, subject, version):
                     return master, master_url
                 await asyncio.sleep(1.0)
 
+    async def subjects_schema_post(self, content_type, *, subject, request):
+        body = request.json
+        try:
+            new_schema = avro.schema.Parse(body["schema"])
```
Ormod: What if there's no "schema" defined in the JSON body

PR #19, opened 2020-01-10
karapace: add metadata_max_age_ms to consumers and producers

vtainio:


karapace/karapace.py
```diff
@@ -157,6 +157,7 @@ def _create_producer(self):
             ssl_certfile=self.config["ssl_certfile"],
             ssl_keyfile=self.config["ssl_keyfile"],
             api_version=(1, 0, 0),
+            metadata_max_age_ms=60000,
```
Ormod: Could we take these out of the config file and default to 60k?

vtainio: Good idea :+1:

PR #24, opened 2020-02-07
schema_backup: add option for overriding topic from config

vtainio:


karapace/schema_backup.py
```diff
@@ -132,13 +132,14 @@ def parse_args():
     for p in {parser_get, parser_restore}:
         p.add_argument("--config", help="Configuration file path", required=True)
         p.add_argument("--location", help="File path for the backup file", required=True)
+        p.add_argument("--topic", help="Kafka topic name to be used")
```
Ormod: It's implicit but explicit is better so please add, required=False)

PR #33, opened 2020-03-24
kafka: move schema registry related logic to a separate file 

gabriel-tincu:
in order to distinguish from the upcoming update pertaining kafka rest

Delete unused karapace client creation function since all tests rely on fixtures
Run make format , add yapf to requirement list
Add reformat to test target prerequisites, fail on reformat diff found
Update setup.py entry_points with new main function location


tests/test_schema_backup.py
```diff
@@ -16,17 +15,6 @@
 baseurl = "http://localhost:8081"
 
 
-def create_karapace(config_path, topic_name, kafka_server):
```
Ormod: Hmm was this no longer needed/unused or why was it removed? (no mention of it in the commit msg)

gabriel-tincu: it was not used in any test file  / cases. The only `Karapace` object used in the tests came from a fixture defined in `conftest.py` called `fixture_karapace` . I'll update the commit message to explain why this was dropped

karapace/karapace.py
```diff
@@ -201,385 +69,5 @@ def r(body, content_type, status=200):
...
-
-    async def write_new_schema_remote(self, subject, body, master_url, content_type):
-        self.log.info("Writing new schema to remote url: %r since we're not the master", master_url)
-        response = await self.http_request(
-            url="{}/subjects/{}/versions".format(master_url, subject), method="POST", json=body, timeout=60.0
-        )
-        self.r(body=response.body, content_type=content_type, status=response.status)
-
-
-def main():
```
vtainio: Our `setup.py` looks for an entry point in this file so moving this without changing it breaks the console scripts.

And this is somewhat related to a bigger point that do we want to have a single entrypoint for starting the application as it is now with `karapace config.json` and make it possible to turn on/off the schema and REST features in the config file or do we want have two different command line tools for this e.g. `karapace_schema schema_config.json` and `karapace_rest rest_config.json`? I've previously thought of the first option but what are your thoughts?

Ormod: It would be ideal if we could serve both from the same process if needed but run them independently. 

I.e. we can have more entrypoints than one but I'd also like to be able to turn things on and off based on configuration.

gabriel-tincu: As a user i usually prefer tools that deal with a minimal number of things , just because the liberty offered by the composition opportunities . That's a personal preference though , and in this case i can't think of any good reason to do it one way versus the other other than the fact that we'd probably end up with less test code when all the logic is provided under one service

Ormod: So the reason why we want both options is that it'll be superbly memory efficient to run them from the same process but like you, many people would prefer them separately so we can cater to both crowds.

requirements-dev.txt
```diff
@@ -5,3 +5,4 @@ flake8==3.7.9
 pylint==2.4.4
 pytest==5.4.1
 requests==2.23.0
+yapf==0.29.0
```
vtainio: Could the yapf version be changed to `0.27.0`- That is the version we are running on the Fedora side of things so the output would match.

gabriel-tincu: sure thing

PR #34, opened 2020-03-25
kafka: temporarily update the kafka client repo ref to the aiven fork

gabriel-tincu:
To make use of zstd compression , until the parent repo is up to date


vtainio:
Github has trouble reporting the build status of this PR but I checked it manually from Travis and it passes. Therefore, merging this.

PR #35, opened 2020-03-25
kafka: support for schema registry ser / deser ops

gabriel-tincu:
Move config related functionality to config file in order to allow usage from different classes, add intellij config folder to `.gitignore.` Write ser/deser logic more or less in line with confluent client, drop any dependency on subject name strategy, in order for the clients to be easily embeddable.


karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+class SchemaRegistryClient:
+    def __init__(self, schema_registry_url: str = "http://localhost:8081", loop=None):
+        self.client = Client(server_uri=schema_registry_url, client=aiohttp.ClientSession(loop=loop))
+        self.base_url = schema_registry_url
+
+    async def post_new_schema(self, subject, schema):
+        payload = {"schema": json_encode(schema.to_json())}
+        result = await self.client.post("subjects/%s/versions" % quote(subject), json=payload)
+        if result.status_code != 200:
+            raise SchemaRetrievalError(result.json())
```
abn, line 74: doesn't `Response.json()` need to be awaited?

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+        return result.json()["id"]
+
+    async def get_latest_schema(self, subject):
+        result = await self.client.get("subjects/%s/versions/latest" % quote(subject))
+        if result.status_code != 200:
+            raise SchemaRetrievalError(result.json())
```
abn, line 80: This bit seems to be repeated.

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+        try:
+            return avro.io.schema.parse(json_result["schema"])
+        except avro.io.schema.SchemaParseException as e:
+            raise SchemaRetrievalError("Failed to parse schema string from response: %r" % json_result) from e
+
+
+class SchemaRegistrySerializerDeserializer:
+    def __init__(self, **config):
+        if not config.get("config_path"):
+            raise Exception("config_path is empty or missing")
```
abn, line 105: We might want to raise something specific here ValueError or RuntimeError perhaps? Depending on use case?

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+
+    async def get_schema_for_topic(self, subject):
+        # apparently we can't really implement the other 2 and keep in line with the current kafka client api
+        # due to the serialize / deserialize method signature only passing topic as a relevant indicator
+        if subject in self.subjects_to_schemas:
+            return self.subjects_to_schemas[subject]
+        log.info("schema for subject %r not found locally, attempting to retrieve latest value", subject)
+        schema_id, schema = await self.registry_client.get_latest_schema(subject)
+        self.subjects_to_schemas[subject] = schema
+        self.schemas_to_ids[self.serialize_schema(schema)] = schema_id
```
abn, line 128: Should we consider using a hash as the key?

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+        schema = await self.get_schema_for_topic(subject)
+        schema_id = self.schemas_to_ids[self.serialize_schema(schema)]
+        writer = DatumWriter(schema)
+        with io.BytesIO() as bio:
+            enc = BinaryEncoder(bio)
+            bio.write(struct.pack(HEADER_FORMAT, START_BYTE, schema_id))
+            try:
+                writer.write(value, enc)
+                enc_bytes = bio.getvalue()
+                return enc_bytes
```
abn, line 144:
```suggestion
                return bio.getvalue()
```

karapace/utils.py
```diff
@@ -7,8 +7,11 @@
 import datetime
 import decimal
 import json as jsonlib
+import logging
 import types
 
+log = logging.getLogger(__name__)
```
abn, line 13: Is this used?

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
+from avro.io import BinaryDecoder, BinaryEncoder, DatumReader, DatumWriter
+from karapace.config import read_config
+from karapace.utils import json_encode
+from tests.utils import Client
```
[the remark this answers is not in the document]
vtainio: Yeah I would move this code out of the test directory if we plan to use it here.

karapace/serialization.py
vtainio: @abn that is handled in the `Client` class

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+        self.schemas_to_ids = {}
+
+    @staticmethod
+    def serialize_schema(schema):
+        return json_encode(schema.to_json())
+
+    async def get_schema_for_topic(self, subject):
+        # apparently we can't really implement the other 2 and keep in line with the current kafka client api
+        # due to the serialize / deserialize method signature only passing topic as a relevant indicator
+        if subject in self.subjects_to_schemas:
```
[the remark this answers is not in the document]
gabriel-tincu: nice catch.... forgot about that :(

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+                writer.write(value, enc)
+                enc_bytes = bio.getvalue()
+                return enc_bytes
+            except avro.io.AvroTypeException as e:
+                raise InvalidMessageSchema("Object does not fit to stored schema") from e
+
+
+class SchemaRegistryDeserializer(SchemaRegistrySerializerDeserializer):
+    async def deserialize(self, subject, bytes_):
+        schema = await self.get_schema_for_topic(subject)
```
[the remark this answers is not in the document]
gabriel-tincu: what if a newer version exists in the registry? What's the desired behavior in that case?

karapace/config.py
```diff
@@ -25,3 +31,15 @@ def set_config_defaults(config):
     config.setdefault("topic_name", "_schemas")
     config.setdefault("metadata_max_age_ms", 60000)
     return config
+
+
+def read_config(config_path):
+    if os.path.exists(config_path):
+        try:
+            config = json.loads(open(config_path, "r").read())
```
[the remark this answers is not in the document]
gabriel-tincu: fixed

karapace/serialization.py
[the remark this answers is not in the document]
gabriel-tincu: updated

karapace/serialization.py
gabriel-tincu: fixed

karapace/serialization.py
gabriel-tincu: fixed

karapace/utils.py
gabriel-tincu: not any more

karapace/serialization.py
gabriel-tincu: it's repeated once for the normal `json()` method and then appears once more for the `message` member of the json body, which is only there __if__ the status code is not OK. Pulling this into a separate method for one repeat does not make much sense to me

karapace/config.py
```diff
@@ -25,3 +31,15 @@ def set_config_defaults(config):
     config.setdefault("topic_name", "_schemas")
     config.setdefault("metadata_max_age_ms", 60000)
     return config
+
+
+def read_config(config_path):
+    if os.path.exists(config_path):
```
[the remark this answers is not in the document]
gabriel-tincu: for a file exists check it's probably the same value, without any import / code update

karapace/serialization.py
gabriel-tincu: It's probably a toss up between log visibility (no __repr__ method yet but one would probably include those 2 dicts), and CPU vs memory consumption . Unless memory does turn out to be an issue i'd rather leave it like this and only put it behind getters / setters otherwise

karapace/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+NAME_STRATEGIES = {
+    "topic_name": topic_name_strategy,
+    "record_name": record_name_strategy,
+    "topic_record_name": topic_record_name_strategy
+}
+
+
+class SchemaRegistryClient:
+    def __init__(self, schema_registry_url: str = "http://localhost:8081", loop=None):
+        self.client = Client(server_uri=schema_registry_url, client=aiohttp.ClientSession(loop=loop))
```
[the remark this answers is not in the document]
gabriel-tincu: it should now, but all the dispose methods will / should be called by the owner object, which at the moment does not exist(will probably be an instance of the Karapace app)

karapace/utils.py
```diff
@@ -48,3 +50,98 @@ def json_encode(obj, *, compact=True, sort_keys=True, binary=False):
...
+            headers = {}
+        if self.client:
+            async with self.client.get(
+                path,
+                headers=headers,
+            ) as res:
+                json_result = await res.json()
+                return Result(res.status, json_result, headers=res.headers)
+        elif self.server_uri:
+            res = requests.get(os.path.join(self.server_uri, path), headers=headers)
```
abn, line 100: This might have be better done with either using `Path.as_posix()` or `urllib`.

karapace/utils.py
```diff
@@ -48,3 +50,98 @@ def json_encode(obj, *, compact=True, sort_keys=True, binary=False):
...
+            ) as res:
+                json_result = await res.json()
+                return Result(res.status, json_result, headers=res.headers)
+        elif self.server_uri:
+            res = requests.delete(os.path.join(self.server_uri, path), headers=headers)
+            return Result(status=res.status_code, json_result=res.json(), headers=res.headers)
+
+    async def post(self, path, json, headers=None):
+        if not headers:
+            headers = {"Content-Type": "application/vnd.schemaregistry.v1+json"}
```
abn, line 119: This should probably go into the client session itself perhaps?

karapace/utils.py
gabriel-tincu: urllib is basically a PITA with trailing/ leading present / missing slashes, which made me use os.path, since i'm always merging a base url with other components without having to worry a leading slash will cancel all previous components

karapace/utils.py
```diff
@@ -48,3 +50,98 @@ def json_encode(obj, *, compact=True, sort_keys=True, binary=False):
...
+        return "Result(status=%d, json_result=%r)" % (self.status_code, self.json_result)
+
+    def ok(self):
+        return self.status_code == 200
+
+
+class Client:
+    def __init__(self, server_uri=None, client=None):
+        self.server_uri = server_uri
+        self.session = requests.Session()
```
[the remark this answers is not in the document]
gabriel-tincu: thinking this was here so we can weight async vs sync implementation for all clients and components that depend on them. It's there partly there because i think the jury is still out on that one and because i did not really check weather removing it will actually break existing tests / functionality. On a more abstract note i'm pretty much trying to touch as few pieces as possible since fixing components that are not broken in a PR that addresses other issue is an easy way to get sidetracked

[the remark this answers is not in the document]
vtainio: Yes indeed removing requests here would break our remote target tests. I would also keep this file unchanged and just moved to a new place for now as it's not new code for this PR. We can come back to improving the client later.

karapace/utils.py
vtainio: Yes at some point we need to move this out of here especially as schema registry and kafka rest use different content-types.

gabriel-tincu: we could probably have a talk about this specific kind of tech debt and figure out a list along with priorities for each element on that list

karapace/utils.py
```diff
@@ -48,3 +50,98 @@ def json_encode(obj, *, compact=True, sort_keys=True, binary=False):
...
+        self.client = client
```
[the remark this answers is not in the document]
gabriel-tincu: Same as above, this appears in the diff because the class was moved from one file to another. Any changes to the test (util now) client implementation and general improvements really don't belong here as i already consider this code a blocker for other tasks

PR #36, opened 2020-03-27
schema_reader: copy global schema id value only on new schemas

vtainio:
Previously this was also assigned when using existing schemas which ended
up creating duplicate id values.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -189,7 +189,8 @@ class KafkaSchemaReader(Thread):
                 }
                 self.log.info("Setting schema_id: %r with schema: %r", value["id"], value["schema"])
                 self.schemas[value["id"]] = value["schema"]
-                self.global_schema_id = value["id"]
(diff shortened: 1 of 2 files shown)


karapace/schema_reader.py
```diff
@@ -208,7 +209,8 @@ def handle_msg(self, key, value):
                 }
                 self.log.info("Setting schema_id: %r with schema: %r", value["id"], value["schema"])
                 self.schemas[value["id"]] = value["schema"]
-                self.global_schema_id = value["id"]
+                if value["id"] > self.global_schema_id:  # Not an existing schema
+                    self.global_schema_id = value["id"]
```
gabriel-tincu: i was wondering actually if it was actually worth wrapping all update code to the global id inside a method , and locking on that, since we're still doing increments in the main thread, and comparisons + updates here

Ormod: I think that would be a good idea but can be done in a separate PR.

PR #42, opened 2020-04-03
Use pre-commit for linting

abn:
This change introduces pre-commit for managing linting config
and tool versions.


vendor/python_accept_types/accept_types.py
```diff
@@ -3,9 +3,9 @@
 See LICENSE for details
 """
 from __future__ import absolute_import, print_function, unicode_literals
+from decimal import Decimal
```
vtainio, line 6: Was the contents of this file accidentally run before as this should be excluded from linting?

abn: Think this was linted with the pylint upgrade; which has been rolled back now.

PR #49, opened 2020-04-29
Update container related files

gabriel-tincu:
Add dockerfile that bundles all karapace needed utilities into one
Update the test dockerfile to also run the rest proxy
Update makefile rules

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -1,23 +1,33 @@
 FROM fedora:30
-
 ARG APACHE_MIRROR=http://www.nic.funet.fi/pub/mirrors/apache.org
 ARG SCALA_VERSION=2.12
 ARG KAFKA_VERSION=2.4.1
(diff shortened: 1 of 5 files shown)


container/Dockerfile
```diff
@@ -1,23 +1,33 @@
...
-ARG CONLFUENT_PATCH_VERSION=3
+ARG CONFLUENT_MAJOR_VERSION=5
+ARG CONFLUENT_MINOR_VERSION=5
+ARG CONFLUENT_PATCH_VERSION=0
 ARG CONFLUENT_RPM_REPO=https://packages.confluent.io/rpm
 
 RUN dnf install -y java-openjdk
+RUN dnf install -y python3
+RUN dnf install -y wget
+RUN dnf install -y git
```
OneCricketeer, line 13: It'd be better to install all these at once, as one layer

gabriel-tincu: done

PR #51, opened 2020-05-03
Parameterize the Dockerfile

OneCricketeer:
You only should need to update a few locations and you can now add `--build-arg` to your CI/CD system


OneCricketeer:
Also, why `fedora` instead of `centos`?

OneCricketeer:
Fedora is used as a Centos / RHEL replacement?

That's fine, but I think the base Docker image for Centos is smaller than Fedora, and RedHat maintains Universal Build Images (UBI) for common use cases like Java / Python, etc

PR #57, opened 2020-05-11
Update Dockerfile

OneCricketeer:
Whoops. Typo


gabriel-tincu:
there's actually one more typo here https://github.com/aiven/karapace/pull/57/files#diff-d08aac623c69515cfe5945737b1b90ecL8 . But they are both being fixed in the latest re-base here
https://github.com/aiven/karapace/pull/49. Thank you for your effort and for noticing this though !

Issue #58, opened 2020-05-12
Evolving schemas not working properly


simonwahlgren:
Hey!

I can't get the following schema to evolve with Karapace but it works as expected with Schema Registry:

Initial schema:
```json
{
  "name": "bar",
  "namespace": "foo",
  "type": "record",
  "fields": [
    {
      "name": "foobar",
      "type": [
        {
          "type": "array",
          "name": "foobar_items",
          "items": {
            "type": "record",
            "name": "foobar_fields",
            "fields": [
              {
                "name": "foo",
                "type": "string"
              }
            ]
          }
        }
      ]
    }
  ]
}
```
Initial registration:
```
〉 http http://localhost:8082/subjects/test/versions schema=@schemas/foo.bar.events-value.avsc 
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Authorization, Content-Type
Access-Control-Allow-Methods: DELETE, GET, OPTIONS, POST, PUT
Access-Control-Allow-Origin: *
Content-Length: 8
Content-Type: application/json
Date: Tue, 12 May 2020 15:17:55 GMT
Server: Karapace/1.1.0-42-gfd4a04c
access-control-expose-headers: etag
etag: "848ec8509f030dd4e57263de2ccbb6f5"

{
    "id": 1
}
```
Evolved schema:
```json
{
  "name": "bar",
  "namespace": "foo",
  "type": "record",
  "fields": [
    {
      "name": "foobar",
      "type": [
        {
          "type": "array",
          "name": "foobar_items",
          "items": {
            "type": "record",
            "name": "foobar_fields",
            "fields": [
              {
                "name": "foo",
                "type": "string"
              },
              {
                "name": "bar",
                "type": [
                  "null",
                  "string"
                ],
                "default": null
              }
            ]
          }
        }
      ]
    }
  ]
}
```
And this is the response I get when trying to register it again:
```
〉 http http://localhost:8082/subjects/test/versions schema=@schemas/foo.bar.events-value.avsc
HTTP/1.1 409 Conflict
Access-Control-Allow-Headers: Authorization, Content-Type
Access-Control-Allow-Methods: DELETE, GET, OPTIONS, POST, PUT
Access-Control-Allow-Origin: *
Content-Length: 93
Content-Type: application/json
Date: Tue, 12 May 2020 15:19:49 GMT
Server: Karapace/1.1.0-42-gfd4a04c

{
    "error_code": 409,
[...]

PR #64, opened 2020-06-02
compatibility: extend supported types

vtainio:
Compatibility checks are extended to support Union, Array, Map and Fixed
types both directly and nested. Primitive and Enum schemas can now also
be used directly and not only in a nested record structure. The simple
presentation for types in maps and arrays is also made to look at
compatibility promotions correctly. In addition, the test suite is
extended with more schema checks to increase coverage.


vtainio:
Fixes #58 additionally

PR #71, opened 2020-06-23
Add more config options to be forwarded to rest producers

gabriel-tincu:


gabriel-tincu:
Addresses https://github.com/aiven/karapace/issues/70

PR #75, opened 2020-07-02
Subject compatibility fix

gabriel-tincu:


tests/test_schema.py
```diff
@@ -134,6 +134,15 @@ async def union_to_union_check(c):
     assert res.status == 200
 
 
+async def missing_subject_compatibility_check(c):
```
vtainio, line 137: This should be added to `run_schema_tests`

tests/test_schema.py
```diff
@@ -134,6 +134,15 @@ async def union_to_union_check(c):
...
+    subject = os.urandom(16).hex()
+    res = await c.get(f"config/{subject}")
+    assert res.status == 404
+    res = await c.get(f"config/{subject}?defaultToGlobal")
```
vtainio, line 141: The `defaultToGlobal` should include a value (either `true` or `false`) and we should test it here.

PR #80, opened 2020-07-07
Handle missing and miscased content headers

gabriel-tincu:
Header search will try to ignore case. Missing Content-Type
headers will default in application/json.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -9,6 +9,7 @@ from karapace.utils import Client
 import json as jsonlib
 import os
 import pytest
+import requests
 
 pytest_plugins = "aiohttp.pytest_plugin"
 baseurl = "http://localhost:8081"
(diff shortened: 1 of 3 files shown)


tests/test_schema.py
```diff
@@ -1462,6 +1463,14 @@ async def check_http_headers(c):
     assert res.status == 200
     assert res.headers["Content-Type"] == "application/vnd.schemaregistry.v1+json"
     assert res.json()["compatibility"] == "NONE"
+    # Subjects request with no headers will return a valid response
+    # using requests directly as aiohttp test client seems to inject a header
+    # thus you can only test this remotely
+    base_url = os.getenv("SERVER_URI", "")
+    if baseurl:
+        path = os.path.join(base_url, "subjects/unknown_subject")
+        res = requests.request("POST", path, data=b"{}", headers={})
```
Ormod, line 1472: Hmm this is without any content-type, could you also add a something like cOnTeNt-TyPe: as one to make sure we have an explicit test for the casing too.

PR #93, opened 2020-08-25
client: more aggresive idle conn recycling

gabriel-tincu:
The default 9 minute idle connection closing is too lenient for
scenarios like maintaining a valid connection pool through a cluster
update.

Injecting a configurable lower value "should" lower the
possibility of clients with a connection pool where the majority of
connections are broken


Ormod (approved):

gabriel-tincu:
closing this down , as the final fix, should there be one, will probably still happen in the kafka client

Issue #98, opened 2020-09-23
Compatibility check for union types doesn't work


nick-zh:
Having a schema like this in the registry
```
{
  "type": "record",
  "name": "order",
  "namespace": "example",
  "fields": [
    {
      "name": "someField",
      "type": [
        "null",
        {
          "type": "record",
          "name": "someEmbeddedRecord",
          "namespace": "example",
          "fields": [
            {
              "name": "name",
              "type": "string"
            }
          ]
        }
      ],
      "default": null
    }
  ]
}
```
and trying to update to this
```
{
  "type": "record",
  "name": "order",
  "namespace": "example",
  "fields": [
    {
      "name": "someField",
      "type": [
        "null",
        {
          "type": "record",
          "name": "someEmbeddedRecord",
          "namespace": "example",
          "fields": [
            {
              "name": "name",
              "type": "string"
            },
            {
              "name": "price",
              "type": "int"
            }
          ]
        }
      ],
      "default": null
    }
  ]
}
```
Karapace rest call for compatibility claims, that it is compatible.
For `full_transitive` only this version should be allowed:
```
            {
              "name": "price",
              "type": [
                "null",
                "int"
              ],
              "default": null
            }
```

Curl call to register the schema:
```
curl --location --request POST 'localhost:9081/subjects/test/versions' \
--header 'Accept: application/vnd.schemaregistry.v1+json' \
--header 'Content-Type: application/vnd.schemaregistry.v1+json' \
--data-raw '{"schema":"{\"type\":\"record\",\"name\":\"order\",\"namespace\":\"example\",\"fields\":[{\"name\":\"someField\",\"type\":[\"null\",{\"type\":\"record\",\"name\":\"someEmbeddedRecord\",\"namespace\":\"example\",\"fields\":[{\"name\":\"name\",\"type\":\"string\"}]}],\"default\":null}]}"}'
```
Curl call to check compatibility:
```
curl --location --request POST 'localhost:9081/compatibility/subjects/test/versions/latest' \
--header 'Accept: application/vnd.schemaregistry.v1+json' \
--header 'Content-Type: application/json' \
[...]

PR #99, opened 2020-09-23
compatibility: add test case for full transitive

gabriel-tincu:
As discussed here https://github.com/aiven/karapace/issues/97


nick-zh:
#98

vtainio (changes requested):
The test should set the `FULL_TRANSITIVE` config for the subject explicitly.

vtainio (approved):

PR #101, opened 2020-09-24
compatibility: config msg offsets while not ready

gabriel-tincu:
Allow storing message offsets for config type messages no matter the
node state, since all node types will accept config type requests
and send relevant messages on the kafka topic, thus in certain scenarios
(single node not ready, but server is up) causing writes to
succeed but offsets to be lost, and thus requests to hang indefinitely

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -85,8 +85,9 @@ class TypedSchema:
 
 
 class KafkaSchemaReader(Thread):
-    def __init__(self, config):
+    def __init__(self, config, master_coordinator=None):
         Thread.__init__(self)
+        self.master_coordinator = master_coordinator
         self.log = logging.getLogger("KafkaSchemaReader")
         self.timeout_ms = 200
         self.config = config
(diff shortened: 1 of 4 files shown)


gabriel-tincu:
This could have been done in several ways:
- this one, which seems to be more spammy when it comes to log output
- ignore message offset reading, which would introduce a race condition, and fail some tests
- handle the config messages as the schema messages (only write to kafka if node is master and ready), which is a bit more code, probably the safest as far as functionality goes, but could also introduce unexpected behavior for some clients (not sure yet)

gabriel-tincu:
addresses https://github.com/aiven/karapace/issues/100

Ormod:
I'm thinking we might want to actually do your option:

"handle the config messages as the schema messages (only write to kafka if node is master and ready), which is a bit more code, probably the safest as far as functionality goes, but could also introduce unexpected behavior for some clients (not sure yet)"

vtainio:
Agreed that we need to have only the master write these values. The other problem that we have is that the offset are stored in the queue for all nodes but they are only removed for the master node. And when the master node fails, the next master will have to process those offset values unnecessarily from the queue and it fills the logs with warnings. We could fix this issue at the same time.

Ormod (approved):

PR #105, opened 2020-10-08
deps: update package versions

gabriel-tincu:
The rpm target requires new package versions in order to properly work


Excds (approved):
Tested building the packages.

PR #109, opened 2020-10-12
deps: remove rpm dependencies

gabriel-tincu:
Cleanup the dependency folder, relying instead on the requirements file
for dependency resolving


dependencies/Makefile
```diff
@@ -1,10 +1,6 @@
 
 LOCAL_PACKAGES = \
-	python3-aiohttp_socks \
-	python3-avro \
-	python3-kafka \
-	python3-aiokafka \
-	python3-jsonschema \
+	python3-aiohttp_socks
```
Excds, line 3: Why not remove this as well?

Excds (changes requested):
requirements-dev.txt should be renamed to requirements.txt

dependencies/Makefile
Excds: Got answers out-of-band.

Excds (approved):

PR #113, opened 2020-10-22
logic: metadata requests to bootstrap nodes

gabriel-tincu:
Patch the karapace kafka client to only make metadata calls to the
bootstrap node. This will fix a bug where metadata calls would end up
continuously being made to no longer valid connections, resulting in a
stale metadata, log spam and probably decreased performance

### DIFF
--- a/karapace/utils.py
+++ b/karapace/utils.py
@@ -5,6 +5,7 @@ Copyright (c) 2019 Aiven Ltd
 See LICENSE for details
 """
 from functools import partial
+from kafka.client_async import KafkaClient, MetadataRequest
 from urllib.parse import urljoin
 
 import datetime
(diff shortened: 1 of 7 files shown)


karapace/utils.py
```diff
@@ -15,6 +16,7 @@
 import types
 
 log = logging.getLogger("KarapaceUtils")
+BLACKLIST_DURATION = 300
```
apisarenco, line 19: Where is this used?

gabriel-tincu: nowhere.... artefact from alternate solution, will remove

apisarenco (approved):
Only issue that has me thinking is the use of a protected member, but I see no good way around it.

PR #115, opened 2020-10-23
logic: update master election logic

gabriel-tincu:
Having the highest cluster node instead of the lowest one be the master
will in theory ensure that newer cluster members are immediately
eligible, thus preventing slaves pointing to dead / killed masters

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -26,6 +26,7 @@ def get_identity_url(scheme, host, port):
 
 
 class SchemaCoordinator(BaseCoordinator):
+    election_strategy = "lowest"
     hostname = None
     port = None
     scheme = None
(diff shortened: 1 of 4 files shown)


karapace/master_coordinator.py
```diff
@@ -57,17 +58,17 @@ def _perform_assignment(self, leader_id, protocol, members):
                 urls[get_identity_url(member_identity["scheme"], member_identity["host"],
                                       member_identity["port"])] = (member_id, member_data)
 
-        lowest_url = sorted(urls)[0]
-        schema_master_id, member_data = urls[lowest_url]
+        master_url = sorted(urls, reverse=self.election_strat.lower() == "highest")[0]
```
NARSimoes, line 61: `self.election_strat.lower() == "highest"`
Is that always False, right? If yes, Why?

gabriel-tincu: For now we would like to inject the possibility of configuring the strategy for master election. Since this is passed from the config, in the future new services can have an updated config , that would allow for a different strategy

karapace/master_coordinator.py
```diff
@@ -56,18 +57,17 @@ def _perform_assignment(self, leader_id, protocol, members):
             if member_identity["master_eligibility"] is True:
                 urls[get_identity_url(member_identity["scheme"], member_identity["host"],
                                       member_identity["port"])] = (member_id, member_data)
-
-        lowest_url = sorted(urls)[0]
-        schema_master_id, member_data = urls[lowest_url]
+        master_url = sorted(urls, reverse=self.election_strategy.lower() == "highest")[0]
```
Excds, line 60: This assignment could be to self.master_url instead.

Excds (changes requested):
I would like for there to be a test verifying the master selection based on strategy.

karapace/master_coordinator.py
gabriel-tincu: fixed

Excds (approved):

PR #116, opened 2020-10-23
logic: clean stale connections

gabriel-tincu:
Metadata updates will ensure we have access to an up to date broker list
for sending requests, but the client connection list can still hold
stale connections. This update just ensures that the connection list
mirrors the cluster state

### DIFF
--- a/karapace/utils.py
+++ b/karapace/utils.py
@@ -191,6 +191,19 @@ def convert_to_int(object_: dict, key: str, content_type: str):
 
 
 class KarapaceKafkaClient(KafkaClient):
+    def close_invalid_connections(self):
+        with self._lock:
+            for node_id in self._conns:
+                if not self.cluster.broker_metadata(node_id):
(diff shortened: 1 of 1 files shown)


karapace/utils.py
```diff
@@ -5,17 +5,19 @@
...
 import decimal
 import json as jsonlib
+import kafka.client_async
 import logging
 import requests
+import time
 import types
 
 log = logging.getLogger("KarapaceUtils")
+BLACKLIST_DURATION = 60
```
NARSimoes, line 20: What is the unit (s or ms)?
Why 60, is it based in what?

gabriel-tincu: It is seconds, and based on nothing specifically.... I just wanted the DNS errors to be silenced but not gone forever, so a minute seemed like a good idea

gabriel-tincu:
Converted to draft until we figure out whether the change actually belongs to the python client library (in a different form)

PR #118, opened 2020-10-26
logic: block ns failed connections

gabriel-tincu:
Metadata updates will ensure we have access to an up to date broker list
for sending requests, but the client connection list can still hold
stale connections.
- Make sure we blackout requests failing with DNS errors
- If we encounter any such connections, close them and then force a
  metadata update
- Actually pass the metadata_max_age parameter to the SR consumer, as
  previously only the producer would make use of it

### DIFF
--- a/karapace/utils.py
+++ b/karapace/utils.py
@@ -5,17 +5,20 @@ Copyright (c) 2019 Aiven Ltd
 See LICENSE for details
 """
 from functools import partial
-from kafka.client_async import KafkaClient, MetadataRequest
+from kafka.client_async import BrokerConnection, KafkaClient, MetadataRequest
 from urllib.parse import urljoin
 
 import datetime
 import decimal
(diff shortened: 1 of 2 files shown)


karapace/utils.py
```diff
@@ -191,6 +194,27 @@ def convert_to_int(object_: dict, key: str, content_type: str):
 
 
 class KarapaceKafkaClient(KafkaClient):
+    def __init__(self, **configs):
+        kafka.client_async.BrokerConnection = KarpaceBrokerConnection
```
hackaugusto, line 198: can this be done without monkey patching?

gabriel-tincu: yes it can, but it will require copy pasting a very large method, so i figured this is the lesser of 2 evils

hackaugusto (approved):
LGTM

PR #123, opened 2020-10-29
compatibility: more union comparison rules

gabriel-tincu:
Add support for comparing unions to simple types when the compatibility
configured would allow the promotion.


gabriel-tincu:
Closing this, as new avro compatibility code makes it obsolete

PR #125, opened 2020-10-31
compat: refactor avro compatibility code

gabriel-tincu:
Keeping in line with the same logic present in the avro java library
that handles type promotions, this aims to be a faithful adaptation of
that code in python.

See https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java


tests/test_avro_schema.py
```diff
@@ -0,0 +1,168 @@
...
+    res = ReaderWriterCompatibilityChecker().get_compatibility(reader, writer)
+    assert res.compatibility is SchemaCompatibilityType.compatible, res
+    res = ReaderWriterCompatibilityChecker().get_compatibility(writer, reader)
+    assert res.compatibility is SchemaCompatibilityType.incompatible, res
+
+    writer = Parse(
+        json.dumps({
+            "type": "record",
+            "name": "CarrierAssignment",
+            "namespace": "lastmile.data.model",
```
elongeau, line 36: Hello,
I'm part of the team that discover this bug and provides this schema, can you _anonymize_ it ? Like removing _lastmile_

Best regards

gabriel-tincu: done

elongeau: thanks :+1

karapace/avro_compatibility.py
```diff
@@ -0,0 +1,268 @@
...
+    def __str__(self):
+        return f"{self.compatibility}: {self.message}"
+
+
+class ReaderWriter:
+    def __init__(self, reader: Schema, writer: Schema) -> None:
+        self.reader, self.writer = reader, writer
+
+    def __hash__(self) -> int:
+        return id(self.reader) ^ id(self.writer)
```
ivanyu, line 72: I think this is incorrect. `hash(a)` should be equal `hash(b)` if `a` equals `b`. However, `id` won't satisfy this property. Try e.g. this:

```python
schema = json.dumps({"name": "foo", "type": "record", "fields": [{"type": "int", "name": "f1"}]})

rw1 = ReaderWriter(Parse(schema), Parse(schema))
rw2 = ReaderWriter(Parse(schema), Parse(schema))

print(rw1 == rw2)
print(hash(rw1) == hash(rw2))
```

```
True
False
```

ivanyu (dismissed):

karapace/avro_compatibility.py
ivanyu: `hash((self.reader, self.writer))` might have been a way, but `Schema` is not hashable.

gabriel-tincu: I lifted the code from these lines https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java#L163-L166. My java is a bit rusty, but i figured the closest i can get to `System.identityHashCode` is by using the `id` function

karapace/avro_compatibility.py
```diff
@@ -0,0 +1,268 @@
...
+        if reader.type == LONG:
+            if writer.type == INT:
+                return result
+            return result.merged_with(self.type_mismatch(reader, writer, location))
+        if reader.type == FLOAT:
+            if writer.type in {INT, LONG}:
+                return result
+            return result.merged_with(self.type_mismatch(reader, writer, location))
+        if reader.type == DOUBLE:
+            if writer.type in {INT, LONG, DOUBLE}:
```
ivanyu, line 170: Shouldn't it be `{INT, LONG, FLOAT}`?

karapace/avro_compatibility.py
ivanyu: Aha, the trick is here: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java#L175

They use pointer comparison in `equals`. In this case `identityHashCode` makes sense and doesn't violate the `hashCode` - `equals` properties. (Though I don't immediately understand why they need this whole thing instead of normal equality.)

To follow this, we need to use `is` instead of `==` in `__eq__`.

karapace/avro_compatibility.py
```diff
@@ -0,0 +1,268 @@
...
+    @staticmethod
+    def check_reader_enum_contains_writer_enum(
+        reader: EnumSchema, writer: EnumSchema, location: List[str]
+    ) -> SchemaCompatibilityResult:
+        result = SchemaCompatibilityResult.compatible()
+        location.append("symbols")
+        writer_symbols, reader_symbols = set(writer.symbols), set(reader.symbols)
+        extra_symbols = writer_symbols.difference(reader_symbols)
+        # TODO-> it looks like this branch is never actually reached
+        if extra_symbols and False:
```
ivanyu, line 233: TBH, I don't get this piece. `xxx and False` will always be `False`.

gabriel-tincu: basically, with the way enums are implemented in python, the entire method will always return a compatible result, which means it can be done with entirely (it's only used in one location). I started writing it, then realised i can't reach that particular branch without defaults, then kind of only ignored the if. On second thought it's probably better if i delete all of it

ivanyu:
@gabriel-tincu there are many interesting test cases in https://github.com/apache/avro/blob/8c15c1241fb52f3ce6eb480bbe16ff70d3bc426a/lang/java/avro/src/test/java/org/apache/avro/TestSchemaCompatibility.java#L42 Any plans porting them?

gabriel-tincu:
> @gabriel-tincu there are many interesting test cases in https://github.com/apache/avro/blob/8c15c1241fb52f3ce6eb480bbe16ff70d3bc426a/lang/java/avro/src/test/java/org/apache/avro/TestSchemaCompatibility.java#L42 Any plans porting them?

Actually i do plan on porting as many of the tests as possible, given the minor api differences

karapace/avro_compatibility.py
ivanyu: Hm...

```json
{
    "type": "enum",
    "name": "Suit",
    "symbols": ["SPADES", "HEARTS", "DIAMONDS", "CLUBS"],
    "default": "SPADES"
}
```

and

```json
{
    "type": "enum",
    "name": "Suit",
    "symbols": ["SPADES", "HEARTS", "DIAMONDS", "CLUBS"]
}
```

are different, right? `Parse` recognizes the difference (`"default"`).

I can't find the place where an `EnumSchema` turns into a Python enum (or vice versa). Could you please point me to it, maybe we'll figure something out.

gabriel-tincu: @ivanyu 
https://github.com/apache/avro/blob/8c15c1241fb52f3ce6eb480bbe16ff70d3bc426a/lang/py3/avro/schema.py#L681-L735

gabriel-tincu: As for the actual initialization :
Java does it like so: https://github.com/apache/avro/blob/942ec8e0cc011ec9fcfff7ba9f5c656bb54d51d1/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1696-L1700
Python like so: https://github.com/apache/avro/blob/8c15c1241fb52f3ce6eb480bbe16ff70d3bc426a/lang/py3/avro/schema.py#L1125-L1128

It seems it would be pretty straight forward to add enum support, but we would have to patch avro

karapace/avro_compatibility.py
gabriel-tincu: fixed

karapace/avro_compatibility.py
gabriel-tincu: fixed

vtainio (approved):

ivanyu (approved):

PR #127, opened 2020-11-19
rest: admin client metadata request refactor

gabriel-tincu:
Due to the fact that the patched KafkaClient class will retrieve
metadata in a more reliable fashion, our patched admin client no longer
needs to make metadata requests directed to the bootstrap node, as the
client logic will guarantee an up to date node id list

### DIFF
--- a/karapace/kafka_rest_apis/admin.py
+++ b/karapace/kafka_rest_apis/admin.py
@@ -5,7 +5,6 @@ from kafka.future import Future
 from kafka.protocol.admin import DescribeConfigsRequest
 from kafka.protocol.metadata import MetadataRequest
 from kafka.protocol.offset import OffsetRequest, OffsetResetStrategy
-from threading import Lock
 from typing import List
 
 import logging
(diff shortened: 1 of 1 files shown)


karapace/kafka_rest_apis/admin.py
```diff
@@ -40,36 +40,23 @@ def get_topic_config(self, topic: str) -> dict:
...
             )
-        # update_metadata on cluster only works with v1 type metadata
         request = MetadataRequest[1](topics=topics)
-        with self.client_lock:
-            node_id = self._client.least_loaded_node()
-            # normal admin code will call poll on client not ready which will sometime block indefinitely on select
-            if not node_id or not self.first_boot or not self._client.ready(node_id):
-                #  pylint: disable=protected-access
-                node_id = list(self._client.cluster._bootstrap_brokers.keys())[0]
-                self.first_boot = False
```
vtainio, line 59: We could remove `first_boot` from `__init__` as well.

vtainio (changes requested):

karapace/kafka_rest_apis/admin.py
gabriel-tincu: nice catch, done !

vtainio (approved):

PR #128, opened 2020-11-19
backup: allow stdout location

gabriel-tincu:
This will make the backup tool more flexible and easier to pipe into
other utilities, should the need arise

### DIFF
--- a/karapace/schema_backup.py
+++ b/karapace/schema_backup.py
@@ -151,10 +151,14 @@ class SchemaBackup:
                     except json.JSONDecodeError:
                         self.log.debug("Invalid JSON in message.value: %r, key: %r", message.value, message.key)
                 values.append((key, value))
-
-        with open(self.backup_location, "w") as fp:
-            fp.write(json.dumps(values))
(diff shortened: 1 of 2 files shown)


karapace/schema_backup.py
```diff
@@ -200,7 +204,7 @@ def parse_args():
     parser_restore = subparsers.add_parser("restore", help="Restore the schema backup from a file")
     for p in {parser_get, parser_restore}:
         p.add_argument("--config", help="Configuration file path", required=True)
-        p.add_argument("--location", help="File path for the backup file", required=True)
+        p.add_argument("--location", default=None, help="File path for the backup file")
```
vtainio, line 207: For restore this should probably still be required.

vtainio (changes requested):

karapace/schema_backup.py
gabriel-tincu: We're doing a check for restore, asserting that the location actually exists. I'll just update the default value to empty string, ensuring that check will fail with the proper error message

vtainio (approved):

PR #129, opened 2020-12-04
tests: ported avro compatibility tests

hackaugusto:
This is the initial port of the tests, commented out tests are:

- Relying in the schema8, which is not parseable by the avro library (it looks like it is a sequence of items separated by a single `,`)
- Doing transitivity tests. These need to be reviewed in more detail to understand what is the proper behaviour. It is not clear which of the "legs" in the chain is breaking the schema compatibility
- There seems to be a few bugs with backwards and forwards compatibility that were commented out

### DIFF
--- a/karapace/avro_compatibility.py
+++ b/karapace/avro_compatibility.py
@@ -1,10 +1,32 @@
 from avro.schema import (
     ARRAY, ArraySchema, BOOLEAN, BYTES, DOUBLE, ENUM, EnumSchema, Field, FIXED, FixedSchema, FLOAT, INT, LONG, MAP,
-    MapSchema, NamedSchema, NULL, RECORD, RecordSchema, Schema, STRING, UNION, UnionSchema
+    MapSchema, NamedSchema, Names, NULL, RECORD, RecordSchema, Schema, SchemaFromJSONData, STRING, UNION, UnionSchema
 )
 from copy import copy
 from enum import Enum
 from typing import Any, cast, Dict, List, Optional, Set
+
+import json
+
+
+def parse_json_ignore_trailing(s: str) -> Any:
+    """ Compatibility function with Avro which ignores trailing data in JSON
+    strings.
+
+    The Python stdlib `json` module doesn't allow to ignore trailing data. If
+    parsing fails because of it, the extra data can be removed and parsed
+    again.
+    """
+    try:
+        json_data = json.loads(s)
+    except json.JSONDecodeError as e:
+        if e.msg != 'Extra data':
+            raise
+
+        json_data = json.loads(s[:e.pos])
+
+    names = Names()
+    return SchemaFromJSONData(json_data, names)
 
 
 class SchemaCompatibilityType(Enum):
--- a/karapace/avro_compatibility.py
+++ b/karapace/avro_compatibility.py
@@ -4,7 +4,7 @@ from avro.schema import (
 )
 from copy import copy
 from enum import Enum
-from typing import cast, Dict, List, Optional, Set
+from typing import Any, cast, Dict, List, Optional, Set
 
 
 class SchemaCompatibilityType(Enum):
@@ -71,6 +71,15 @@ class SchemaCompatibilityResult:
             messages={message},
         )
         return ret
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, SchemaCompatibilityResult):
+            return False
+
+        return (
+            self.locations == other.locations and self.messages == other.messages
+            and self.compatibility == other.compatibility and self.incompatibilities == other.incompatibilities
+        )
 
     def __str__(self):
         return f"{self.compatibility}: {self.messages}"
--- a/tests/test_avro_schema.py
+++ b/tests/test_avro_schema.py
@@ -96,25 +96,26 @@ def test_schemaregistry_basic_forwards_compatibility():
     schema.
     """
     msg = "adding a field is a forward compatible change"
-    res = ReaderWriterCompatibilityChecker().get_compatibility(schema2, schema1)
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema1, schema2)
     assert res == SchemaCompatibilityResult.compatible(), msg
 
-    # msg = "adding a field is a forward compatible change"
-    # res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema1)
-    # assert res == SchemaCompatibilityResult.compatible(), msg
-
     msg = "adding a field is a forward compatible change"
-    res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema2)
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema1, schema3)
     assert res == SchemaCompatibilityResult.compatible(), msg
 
     msg = "adding a field is a forward compatible change"
     res = ReaderWriterCompatibilityChecker().get_compatibility(schema2, schema3)
     assert res == SchemaCompatibilityResult.compatible(), msg
 
-    msg = "removing a default is not a transitively compatible change"
-    res = ReaderWriterCompatibilityChecker().get_compatibility(schema1, schema3)
+    msg = "adding a field is a forward compatible change"
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema2)
     assert res == SchemaCompatibilityResult.compatible(), msg
-    res = ReaderWriterCompatibilityChecker().get_compatibility(schema1, schema2)
+
+    msg = "removing a default is not a transitively compatible change"
+    # Only schema 2 is checked!
+    # res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema1)
+    # assert res != SchemaCompatibilityResult.compatible(), msg
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema2, schema1)
     assert res == SchemaCompatibilityResult.compatible(), msg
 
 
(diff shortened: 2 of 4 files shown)


tests/test_avro_schema.py
```diff
@@ -8,6 +8,206 @@
...
+schema3 = parse('{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1"},{"type":"string","name":"f2"}]}')
+schema4 = parse('{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1_new","aliases":["f1"]}]}')
+schema6 = parse('{"type":"record","name":"myrecord","fields":[{"type":["null","string"],"name":"f1","doc":"doc of f1"}]}')
+schema7 = parse(
+    '{"type":"record","name":"myrecord","fields":[{"type":["null","string","int"],"name":"f1","doc":"doc of f1"}]}'
+)
+# schema8 = (
+#     '{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1"},{"type":"string",'
+#     '"name":"f2","default":"foo"}]},{"type":"string","name":"f3","default":"bar"}]}'
+# )
```
gabriel-tincu, line 26: this JSON is invalid, which is probably causing the tests related to it to fail. As a rule of thumb, i find that `parse(dumps..` is safer than `parse`

gabriel-tincu (changes requested):

tests/test_avro_schema.py
```diff
@@ -8,6 +8,206 @@
...
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema2, schema1)
+    assert res.compatibility is SchemaCompatibilityType.compatible, msg
+
+    msg = "adding a field w/o default is NOT a backward compatible change"
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema1)
+    assert res.compatibility is SchemaCompatibilityType.incompatible, msg
+
+    # msg = "changing field name with alias is a backward compatible change"
+    # res = ReaderWriterCompatibilityChecker().get_compatibility(schema4, schema1)
+    # assert res.compatibility is SchemaCompatibilityType.compatible, msg
```
gabriel-tincu, line 48: Would be nice if you included the `locations` and `messages` in the failure message, so we could pinpoint the problem

gabriel-tincu (dismissed):

tests/test_avro_schema.py
```diff
@@ -8,6 +8,206 @@
...
 import pytest
 
+# Schemas defined in AvroCompatibilityTest.java. Used here to ensure compatibility with the schema-registry
+schema1 = parse('{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1"}]}')
+schema2 = parse(
+    '{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1"},{"type":"string",'
+    '"name":"f2","default":"foo"}]}'
+)
+schema3 = parse('{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1"},{"type":"string","name":"f2"}]}')
+schema4 = parse('{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1_new","aliases":["f1"]}]}')
```
gabriel-tincu, line 18: Any failures related to this could make sense, as we only compare field names at the moment, instead of aliases

tests/test_avro_schema.py
hackaugusto: > this JSON is invalid, which is probably causing the tests related to it to fail. 

Exactly. I can't call `parse` here because an exception is raised.

The problem is that this is how the tests are originally defined [ref](https://github.com/confluentinc/schema-registry/blob/e8f89f8e55ac3f2a3c32abbd9f4de0affff238c8/core/src/test/java/io/confluent/kafka/schemaregistry/avro/AvroCompatibilityTest.java#L76). I commented it out until I have time to dig into the java implementation and see how it is handling the invalid JSON.

tests/test_avro_schema.py
hackaugusto: I added an automatic formatter for this. It will looks like this:

```
    assert res != SchemaCompatibilityResult.compatible(), msg                                                                                                                                                                                 
E   AssertionError: adding a field with default is a backward compatible change                                                                                                                                                               
E   assert Comparing SchemaCompatibilityResult instances:                                                                                                                                                                                     
E       Left:                                                                                                                                                                                                                                 
E         compatibility SchemaCompatibilityType.compatible                                                                                                                                                                                    
E         locations:                                                                                                                                                                                                                          
E           /                                                                                                                                                                                                                                 
E       Right:                                                                                                                                                                                                                                
E         compatibility SchemaCompatibilityType.compatible                                                                                                                                                                                    
E         locations:... 
```

tests/test_avro_schema.py
hackaugusto: Okay, so it seems the tests don't care if there is trailing data when parsing. I added a test to print the contents of the `schema8` and this is what I got:

```java
  @Test
  public void testPrintSchemaEight() {
    System.out.println(schema8);
  }
```
> {"type":"record","name":"myrecord","fields":[{"name":"f1","type":"string"},{"name":"f2","type":"string","default":"foo"}]}

Now I have to check if this is also done for production code.

hackaugusto: Okay, so:

- [The tests uses `AvroUtils.parseSchema`](https://github.com/confluentinc/schema-registry/blob/e8f89f8e55ac3f2a3c32abbd9f4de0affff238c8/core/src/test/java/io/confluent/kafka/schemaregistry/avro/AvroCompatibilityTest.java#L77)
- [Which just forwards the call to the `AvroSchema` constructor](https://github.com/confluentinc/schema-registry/blob/e8f89f8e55ac3f2a3c32abbd9f4de0affff238c8/core/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroUtils.java#L32)
- [The `AvroSchema` will instantiate a avro `Schema.Parser`](https://github.com/confluentinc/schema-registry/blob/e8f89f8e55ac3f2a3c32abbd9f4de0affff238c8/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroSchema.java#L119-L123), and [call use it to parser the schema](https://github.com/confluentinc/schema-registry/blob/e8f89f8e55ac3f2a3c32abbd9f4de0affff238c8/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroSchema.java#L74)
- [Apache avro will instantiate a jackson JsonParser](https://github.com/apache/avro/blob/89c03cfe5fe8a0ad1aacf910e9b94c57a41ea67e/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1413), and [use it with a ObjectMaper](https://github.com/apache/avro/blob/89c03cfe5fe8a0ad1aacf910e9b94c57a41ea67e/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1425)
- [The ObjectMapper will deserialized the object, and only throw if the flag `FAIL_ON_TRAILING_TOKENS` is set](https://github.com/FasterXML/jackson-databind/blob/39fab67429c0f23399f031e0c0dbd4a13a73622d/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L2421-L2423p), which happens [to be off by default](https://github.com/FasterXML/jackson-databind/blob/39fab67429c0f23399f031e0c0dbd4a13a73622d/src/main/java/com/fasterxml/jackson/databind/DeserializationFeature.java#L238-L254)

So the above JSON is *not* invalid from the perspective of the test tooling. The next question is if it is ever invalid from the API perspective.

hackaugusto: It looks like this is the case, the [`AvroSchemaProvider` uses the same `AvroSchema` class to parse the json](https://github.com/confluentinc/schema-registry/blob/e8f89f8e55ac3f2a3c32abbd9f4de0affff238c8/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroSchemaProvider.java#L53-L55)

karapace/avro_compatibility.py
```diff
@@ -1,10 +1,32 @@
...
 from enum import Enum
-from typing import cast, Dict, List, Optional, Set
+from typing import Any, cast, Dict, List, Optional, Set
+
+import json
+
+
+def parse_json_ignore_trailing(s: str) -> Any:
+    """ Compatibility function with Avro which ignores trailing data in JSON
+    strings.
```
ivanyu, line 14: Out of curiosity: when does this happen?

karapace/avro_compatibility.py
```diff
@@ -72,6 +94,15 @@ def incompatible(incompat_type: SchemaIncompatibilityType, message: str, locatio
...
         return ret
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, SchemaCompatibilityResult):
+            return False
+
+        return (
+            self.locations == other.locations and self.messages == other.messages
+            and self.compatibility == other.compatibility and self.incompatibilities == other.incompatibilities
+        )
```
ivanyu, line 104: In in the [Java version](https://github.com/apache/avro/blob/eda435931b8dbd913b98567814494ddfe4bea016/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java#L610), the class doesn't have `locations` and `messages` and hence they are not used for comparison. I wonder what might be the implications of including them into comparison in our implementation.

tests/test_avro_schema.py
```diff
@@ -2,16 +2,270 @@
...
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema2, schema1)
+    assert res == SchemaCompatibilityResult.compatible(), msg
+
+    msg = "removing a default is a compatible change, but not transitively"
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema2)
+    assert res == SchemaCompatibilityResult.compatible(), msg
+
+    msg = "removing a default is not a transitively compatible change"
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema2)
+    assert res == SchemaCompatibilityResult.compatible(), msg
```
ivanyu, line 96: (Not implying it's bad and should be changed,) are these the same thing?

ivanyu: If it's this way in the SR tests, we'd better leave it as is.

tests/test_avro_schema.py
```diff
@@ -2,16 +2,270 @@
...
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema3, schema1)
+    assert res != SchemaCompatibilityResult.compatible(), msg
+
+
+def test_schemaregistry_basic_forwards_compatibility():
+    """
+    Forward compatibility: A new schema is forward compatible if the previous schema can read data written in this
+    schema.
+    """
+    msg = "adding a field is a forward compatible change"
```
ivanyu, line 106:
```suggestion
    msg = "adding a field with default is a forward compatible change"
```
maybe?

tests/test_avro_schema.py
```diff
@@ -2,16 +2,270 @@
...
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema1, schema2)
+    assert res == SchemaCompatibilityResult.compatible(), msg
+
+    msg = "adding a field is a forward compatible change"
+    res = ReaderWriterCompatibilityChecker().get_compatibility(schema1, schema3)
+    assert res == SchemaCompatibilityResult.compatible(), msg
+
+    msg = "adding a field is a forward compatible change"
```
ivanyu, line 114: Does this and the next one really check adding a field or it's more about defaults?

karapace/avro_compatibility.py
hackaugusto: AFAIK `locations` and `messages` were added to aid debugging. What could happen is the tests is that if the code is not deterministic we would see more failures with our code, e.g. if the messages themselves change across runs because a timestamp is added, then we would see more failures. For testing ATM this seems fine though.

tests/test_avro_schema.py
hackaugusto: That is how it is defined in the original test suite: https://github.com/confluentinc/schema-registry/blob/v5.0.3/core/src/test/java/io/confluent/kafka/schemaregistry/avro/AvroCompatibilityTest.java#L118-L121

tests/test_avro_schema.py
hackaugusto: sure, it makes it more understandable, the messages were just copied from the original https://github.com/confluentinc/schema-registry/blob/v5.0.3/core/src/test/java/io/confluent/kafka/schemaregistry/avro/AvroCompatibilityTest.java#L131-L138

karapace/avro_compatibility.py
hackaugusto: In the test case it would be here: https://github.com/confluentinc/schema-registry/blob/v5.0.3/core/src/test/java/io/confluent/kafka/schemaregistry/avro/AvroCompatibilityTest.java#L69-L75

And it seems the production code also uses the same parsing facilities: https://github.com/confluentinc/schema-registry/blob/v5.0.3/core/src/main/java/io/confluent/kafka/schemaregistry/storage/KafkaSchemaRegistry.java#L821

I did not check, however, if the configuration can be globally changed.

tests/test_avro_schema.py
ivanyu: Gotcha

tests/test_avro_schema.py
ivanyu: Yep, I realized down the road you've copied them. Let them be as is, it'd be easier to keep them in sync.

ivanyu (approved):

PR #132, opened 2020-12-09
Added typing info to conftest.py

hackaugusto:
As the title says, this just adds type information to the conftest.

While adding the typing annotations I noticed some fixtures are conditional in respect to the env variables `REST_URI` and `REGISTRY_URI`. From the commit history it looks like these are just artifacts from how the tests used to be executed. I guess they should eventually be removed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,14 +4,17 @@ karapace - conftest
 Copyright (c) 2019 Aiven Ltd
 See LICENSE for details
 """
+from dataclasses import dataclass
 from kafka import KafkaAdminClient, KafkaProducer
 from karapace.avro_compatibility import SchemaCompatibilityResult
 from karapace.config import write_config
 from karapace.kafka_rest_apis import KafkaRest, KafkaRestAdminClient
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -21,7 +21,7 @@
     version=version_for_setup_py,
     zip_safe=False,
     packages=find_packages(exclude=["test"]),
-    install_requires=["aiohttp"],
+    install_requires=["aiohttp", "dataclasses;python_version<'3.7'"],
```
hackaugusto, line 24: fedora 31 is on 3.7 already [ref](https://docs.fedoraproject.org/en-US/fedora/f31/release-notes/developers/Development_Python/#_python_3_8_is_now_available), so the `karapace.spec` doesn't need a new dependency

ivanyu: Considering we're going to drop 3.6 (https://github.com/aiven/karapace/pull/134), do we still need to add this?

hackaugusto: Nope :) I will update this PR

ivanyu (approved):

PR #133, opened 2020-12-09
karapace: read configuration only once

hackaugusto:
This change the code so that the configuration file is read only once at the `main`. I'm doing this for a few reasons:

- I want to eventually add validation for the configuration and make it a dataclass. This will make it possible to have type checking and auto completion for the config, have faster errors on improper configuration, and make it easier to document all the available options. (some of that, but not all, is done by the [`set_config_defaults`](https://github.com/aiven/karapace/blob/master/karapace/config.py#L68-L81))
- This should also make testing easier, since files won't be necessary any longer. ATM I couldn't remove the file path because the fixtures are piggybacking into the class state and getting the path.
- This also makes sure the config is consistent, in the unlikely event of the file changing while the servers are starting up.

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -5,7 +5,7 @@ Copyright (c) 2019 Aiven Ltd
 See LICENSE for details
 """
 
-from typing import Dict, Union
+from typing import Dict, IO, Union
 
 import json
 import os
@@ -86,14 +86,13 @@ def write_config(config_path: str, custom_values: Dict[str, Union[str, int, bool
         fp.write(json.dumps(custom_values))
 
 
-def read_config(config_path: str) -> Dict[str, Union[str, int, bool]]:
-    with open(config_path, "r") as cf:
-        try:
-            config = json.loads(cf.read())
-            config = set_config_defaults(config)
-            return config
-        except Exception as ex:
-            raise InvalidConfiguration(ex)
+def read_config(config_handler: IO) -> Dict[str, Union[str, int, bool]]:
+    try:
+        config = json.load(config_handler)
+        config = set_config_defaults(config)
+        return config
+    except Exception as ex:
+        raise InvalidConfiguration(ex)
 
 
 def create_ssl_context(config: Dict[str, Union[str, int, bool]]) -> ssl.SSLContext:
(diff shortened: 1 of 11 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -35,13 +35,13 @@ class FormatError(Exception):
 
 class KafkaRest(KarapaceBase):
     # pylint: disable=attribute-defined-outside-init
-    def __init__(self, config_path: str) -> None:
-        super().__init__(config_path)
+    def __init__(self, config_file_path: str, config: dict) -> None:
```
hackaugusto, line 38: I wanted to remove `config_file_path` and have only `config`. But the `KarapaceBase` needs it because a few tests rely on the availability of the config path.

karapace/kafka_rest_apis/__main__.py
```diff
@@ -1,29 +1,36 @@
...
+from karapace.config import read_config
 from karapace.kafka_rest_apis import KafkaRest
 
 import argparse
 import os
 import sys
 
 
-def main():
+def main() -> int:
```
hackaugusto, line 11: All the scripts are doing `sys.exit(main())` but most of the functions were missing a return with the error code.

karapace/config.py
```diff
@@ -86,14 +86,13 @@ def write_config(config_path: str, custom_values: Dict[str, Union[str, int, bool
         fp.write(json.dumps(custom_values))
 
 
-def read_config(config_path: str) -> Dict[str, Union[str, int, bool]]:
-    with open(config_path, "r") as cf:
```
hackaugusto, line 90: I changed the command line parser to use the built-in `FileType`. Which will verify if the file exists and give back a opened handler. The code path with a file name is not necessary any more after this change.

karapace/karapace.py
```diff
@@ -67,10 +66,6 @@ def close(self):
         self.producer.close()
         self.producer = None
 
-    @staticmethod
-    def read_config(config_path):
-        return read_config(config_path)
```
hackaugusto, line 72: The places that need to read the config are just calling `read_config` directly.

karapace/config.py
ivanyu: Neat!

ivanyu (changes requested):
LGTM, but conflicts need fixing.

ivanyu (approved):

PR #135, opened 2020-12-10
bugfix: fetch all schemas from a large topic

hackaugusto:
Because poll works in batches, topics with more than `max_poll_records`
(default of 500) have to be polled multiple times, otherwise only the
first `max_poll_records` are saved by the backup tool.

This assumes there are two timeouts, one for the request and another for
the socket read. A request timeout would occur when the server does not
have additional data to send the client after the timeout, so an empty
response is returned by the server. A socket read timeout happens when
the client is waiting for the server but a response never arrives.

The above is necessary because it allows to use an empty response as an
indicator of "no more data".

### DIFF
--- a/karapace/schema_backup.py
+++ b/karapace/schema_backup.py
@@ -134,23 +134,31 @@ class SchemaBackup:
         if not self.consumer:
             self.init_consumer()
         self.log.info("Starting schema backup read for topic: %r", self.topic_name)
+
         values = []
-        raw_msg = self.consumer.poll(timeout_ms=self.timeout_ms)
-        for _, messages in raw_msg.items():
-            for message in messages:
-                key = message.key.decode("utf8")
-                try:
-                    key = json.loads(key)
-                except json.JSONDecodeError:
-                    self.log.debug("Invalid JSON in message.key: %r, value: %r", message.key, message.value)
-                value = None
-                if message.value:
-                    value = message.value.decode("utf8")
+        topic_fully_consumed = False
+
+        while not topic_fully_consumed:
+
+            raw_msg = self.consumer.poll(timeout_ms=self.timeout_ms)
+            topic_fully_consumed = len(raw_msg) == 0
+
+            for _, messages in raw_msg.items():
+                for message in messages:
+                    key = message.key.decode("utf8")
                     try:
-                        value = json.loads(value)
+                        key = json.loads(key)
                     except json.JSONDecodeError:
-                        self.log.debug("Invalid JSON in message.value: %r, key: %r", message.value, message.key)
-                values.append((key, value))
+                        self.log.debug("Invalid JSON in message.key: %r, value: %r", message.key, message.value)
+                    value = None
+                    if message.value:
+                        value = message.value.decode("utf8")
+                        try:
+                            value = json.loads(value)
+                        except json.JSONDecodeError:
+                            self.log.debug("Invalid JSON in message.value: %r, key: %r", message.value, message.key)
+                    values.append((key, value))
+
         ser = json.dumps(values)
         if self.backup_location:
             with open(self.backup_location, "w") as fp:


hackaugusto:
This was tested with:

```yaml
---
version: '2'
services:
  zookeeper:
    image: confluentinc/cp-zookeeper:6.0.0
    hostname: zookeeper
    container_name: zookeeper
    ports:
      - "2181:2181"
    environment:
      ZOOKEEPER_CLIENT_PORT: 2181
      ZOOKEEPER_TICK_TIME: 2000

  broker:
    image: confluentinc/cp-server:6.0.0
    hostname: broker
    container_name: broker
    depends_on:
      - zookeeper
    ports:
      - "9092:9092"
      - "9101:9101"
    environment:
      KAFKA_BROKER_ID: 1
      KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://broker:29092,PLAINTEXT_HOST://localhost:9092
      KAFKA_METRIC_REPORTERS: io.confluent.metrics.reporter.ConfluentMetricsReporter
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
      KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: 0
      KAFKA_CONFLUENT_LICENSE_TOPIC_REPLICATION_FACTOR: 1
      KAFKA_CONFLUENT_BALANCER_TOPIC_REPLICATION_FACTOR: 1
      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
      KAFKA_JMX_PORT: 9101
      KAFKA_JMX_HOSTNAME: localhost
      KAFKA_CONFLUENT_SCHEMA_REGISTRY_URL: http://schema-registry:8081
      CONFLUENT_METRICS_REPORTER_BOOTSTRAP_SERVERS: broker:29092
      CONFLUENT_METRICS_REPORTER_TOPIC_REPLICAS: 1
      CONFLUENT_METRICS_ENABLE: 'true'
      CONFLUENT_SUPPORT_CUSTOMER_ID: 'anonymous'

  schema-registry:
    image: confluentinc/cp-schema-registry:6.0.0
    hostname: schema-registry
    container_name: schema-registry
    depends_on:
      - broker
    ports:
      - "8081:8081"
    environment:
      SCHEMA_REGISTRY_HOST_NAME: schema-registry
      SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS: 'broker:29092'
      SCHEMA_REGISTRY_LISTENERS: http://0.0.0.0:8081

  rest-proxy:
    image: confluentinc/cp-kafka-rest:6.0.0
    depends_on:
      - broker
      - schema-registry
    ports:
      - 8082:8082
    hostname: rest-proxy
    container_name: rest-proxy
    environment:
      KAFKA_REST_HOST_NAME: rest-proxy
      KAFKA_REST_BOOTSTRAP_SERVERS: 'broker:29092'
      KAFKA_REST_LISTENERS: "http://0.0.0.0:8082"
      KAFKA_REST_SCHEMA_REGISTRY_URL: 'http://schema-registry:8081'
```
```sh
podman-compose schema-registry.yml
```
```sh
for i in $(seq 1000); do curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"schema": "{\"type\":\"record\",\"name\":\"Payment'$i'\",\"namespace\":\"io.confluent.examples.clients.basicavro\",\"fields\":[{\"name\":\"id\",\"type\":\"string\"},{\"name\":\"amount\",\"type\":\"double\"}]}"}' http://localhost:8081/subjects/test-value/versions; done
python -m karapace.schema_backup get --config /tmp/karapace.json --location /tmp/backup.json
jq '. | length' /tmp/backup.json
```

karapace/schema_backup.py
```diff
@@ -134,23 +134,31 @@ def request_backup(self):
...
-                values.append((key, value))
+                        self.log.debug("Invalid JSON in message.key: %r, value: %r", message.key, message.value)
+                    value = None
+                    if message.value:
+                        value = message.value.decode("utf8")
+                        try:
+                            value = json.loads(value)
+                        except json.JSONDecodeError:
+                            self.log.debug("Invalid JSON in message.value: %r, key: %r", message.value, message.key)
+                    values.append((key, value))
```
gabriel-tincu, line 160: Can we end up adding an invalid json byte blob as a value?

vtainio (approved):
Excellent :+1:

karapace/schema_backup.py
hackaugusto: It looks so, for both keys and values. What are the expectations here? Should that ever happen? If it does, how should it be handled?

vtainio: For now I think we can leave it as it is for debugging purposes. The restore part of this tool needs some love before it's going to be used for real.

PR #144, opened 2021-01-14
bugfix: fix infinite loop

hackaugusto:
Upstream issue: https://github.com/dpkp/kafka-python/issues/2193

Note: This PR should not be merged before dependency `kafka-python` is updated, or we merge the patch to our repo https://github.com/aiven/kafka-python/pull/20

Calling `_send_request_to_node` with the result of `least_loaded_node`
results in an infinite loop if the metadata is out-dated and
`least_loaded_node` return the `node_id` of a broker that was removed
from the cluster. The method `_send_request_to_least_loaded_node`
handles this race and fixes the infinite loop.


mts-aiven (approved):
LGTM

PR #145, opened 2021-01-26
security: add security policy

tapioaiven:


Myrcurial (approved):
Approve adding security policy documentation to the repo.

PR #146, opened 2021-01-26
removed six

hackaugusto:
Karapace [requires `python>=3.7`](https://github.com/aiven/karapace/blob/4d9524812a9a0ee7be4a26c8f146d0cdd70bdaf0/setup.py#L41), using `six` doesn't make sense anymore


mts-aiven (approved):
ok

PR #148, opened 2021-01-27
depedency: using accept-types from pypi

hackaugusto:


hackaugusto:
There is a small patch to the dependency, so we can not use the official package. The difference is that the package does not accept `*` as a valid, but we have to for compatibility reasons.

PR #149, opened 2021-01-27
Updated Dockerfile

hackaugusto:
This introduces a Dockerfile for Karapace based on `debian:stable-slim`, this base image was chosen because of its size and compatibility with glibc.

Things taken care by this PR:

- Images with th [opencontainers pre-defined labels](https://github.com/opencontainers/image-spec/blob/79b036d80240ae530a8de15e1d21c7ab9292c693/annotations.md#pre-defined-annotation-keys)
- The container's entrypoint was fixed to have the main application as PID 1, so that signals are properly delivered to the app (and not to the spawning shell).
- The image is running with a user named `karapace` instead of root.
- The image is using multi-stage setup to produce the wheels in a builder image, which are then installed on the runtime image. This reduces the end image size.
- There are separate Dockerfiles for the REST and the schema-registry. These images are only for convenience and completely reuse the existing cached layers. 
- The `RUN` commands have been chosen to optimized layer cache reuse. Dependencies are handled independently from the main app because they change less often, and the caching layers can be reused.
- The idea here is to set the ground work for reusable builds. The direct dependencies are pinned which is a start, next step is to pin the full dependency closure.

I tested this with:

```shell
./container/build_image.sh HEAD latest
(cd container; sudo docker-compose up -d)
```
Unfortunately I didn't manage to run this with `podman-compose`, it seems the tool can't use images from local repositories.


hackaugusto:
I did a quick check on the available base images. Here are some notes:

- [Alpine](https://hub.docker.com/_/alpine/): This [seems like a bad idea](https://pythonspeed.com/articles/alpine-docker-python/). tl;dr the whells won't work because alpine uses musl instead of glibc
- [Debian](https://hub.docker.com/_/debian): [stable](https://hub.docker.com/layers/debian/library/debian/stable/images/sha256-8fe3e2d7237fb716015b846524f08e97c2a7f626e1466e013ed43ff2d3aa2090?context=explore) (currently Debian 10 [buster](https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64/buster)) is around 48MB, buster slim is 27MB.
- [Fedora](https://hub.docker.com/_/fedora): [33](https://github.com/fedora-cloud/docker-brew-fedora/blob/33/x86_64/Dockerfile) is at 64MB. There is no slim version.
- [Ubuntu](https://hub.docker.com/_/ubuntu): [groovy](https://github.com/tianon/docker-brew-ubuntu-core/blob/dist-amd64/groovy/Dockerfile) is at 30MB.

and then we have the python specific versions of each of the base images mentioned above:

- python:alpine is around 16MB, but again, it uses musl
- [python:slim-buster](https://hub.docker.com/layers/python/library/python/slim-buster/images/sha256-88625be04f61b3c2cefcbbcaa49c1f5a2106cc4f4b85557f2ce9e787b05e7625?context=explore) is around 42MB.

container/build_image.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# HACK: Delete files from the `.git` which change on every operation.
+#
+# - .git/logs/HEAD - command history
+# - .git/index - binary file for the current index, very important for a
+#    working repository, not interesting for our image
+#
+# Being extra cautious here, to make sure only the temporary copy is modified.
+rm ${code_checkout}/.git/logs/HEAD
+rm ${code_checkout}/.git/index
```
hackaugusto, line 49: This is the most questionable change introduced by the PR. This improves cache reuse by removing the files that change when git is used, for instance, a git checkout will be logged into `.git/logs/HEAD` making the contents of the repo different for each run.

An alternative here would be to have this script run the `version.py` script outside of the container, and skip the copy of the repo into the container. I guess both work and I just didn't want to re implement this in the interest of time.

container/Dockerfile
```diff
@@ -1,18 +1,76 @@
...
+
+RUN apt-get update && \
+    apt-get -y install --no-install-recommends python3-pip && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /build/dependencies-wheels/*.whl /build/dependencies-wheels/
+RUN pip3 install --no-deps /build/dependencies-wheels/*.whl && rm -rf /build/dependencies-wheels/
+
+COPY --from=builder /build/karapace-wheel/*.whl /build/karapace-wheel/
+RUN pip3 install --no-deps /build/karapace-wheel/*.whl && rm -rf /build/karapace-wheel/
```
tvainika, line 63: It is also option to have three stages, where one is builder, second is for installing & generating runtime dependencies, and then final image just copies from runtime-deps images without installing pip or anything into resulting container. That would make the resulting image smaller (not containing pip).

container/build_image.sh
tvainika: Would it work to list those files in `.dockerignore` to solve caching etc. and skip not wanted files from ending up into container image?

tvainika (commented):
I didn't quite catch why you generated image both as karapace_registry and karapace_rest as those images have equal contents? Would it work to just launch the image with different args depending on what mode you want to run?

Also I'd suggest to include some usage notes into main level README how to launch the container image.

container/Dockerfile
```diff
@@ -1,18 +1,76 @@
...
-RUN mkdir /app
-ADD . /app
-WORKDIR /app
-RUN make
-ENTRYPOINT ["/bin/bash", "/opt/start.sh"]
+# Builder image contains header files and additional dependencies necessary to
+# generate wheel files.
+FROM debian:stable-slim AS builder
+
+LABEL org.opencontainers.image.authors=hannu.valtonen@aiven.io \
```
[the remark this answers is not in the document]
hackaugusto: > Adding labels to intermediate builder container might be somewhat useless, but you had that too for resulting container

I added here too to make sure all layers have the LICENSE metadata.

hackaugusto:
> Would it work to list those files in .dockerignore to solve caching etc. and skip not wanted files from ending up into container image?

That is a good question, I didn't check that.

container/Dockerfile
hackaugusto: > then final image just copies from runtime-deps images

it is worth it though? what about I just delete the pip package before the commit?

hackaugusto:
> I didn't quite catch why you generated image both as karapace_registry and karapace_rest as those images have equal contents? 

> There are separate Dockerfiles for the REST and the schema-registry. These images are only for convenience and completely reuse the existing cached layers.

> Would it work to just launch the image with different args depending on what mode you want to run?

Yes.

container/Dockerfile
hackaugusto: Adding:

```Dockerfile
RUN apt-get -y remove --prune python3-pip
```
The new image had size `53572016`, against `53583060`. Doesn't seem worth it

tvainika: A later command cannot remove files from underlying layer, so the way you tested it basically has no effect. The underlying layer still contains pip and pulling docker image would download that layer as part of the container image.

Regardless of that, I'm fine having pip included in the image, but if you want to remove pip from the image, it cannot be done that way adding that removal command later.

hackaugusto: Ah, you're right, I forgot that the format included all layers.

container/build_image.sh
```diff
@@ -0,0 +1,61 @@
...
+    echo "Invalid commit provided '${COMMIT}'"
+    echo ""
+    echo "$0 <commitish_object> [:tag]"
+    exit 1
+fi
+
+code_checkout=$(mktemp --directory --suffix=-karapace-image)
+trap "rm -rf ${code_checkout}" EXIT
+
+git clone $(pwd) "${code_checkout}"
```
tvainika, line 36:
```suggestion
git clone $(dirname $0)/.. "${code_checkout}"
```
would allow this script to be called from anywhere, and not fail badly if not called from repository root as `./container/build_image.sh`

tvainika (dismissed):
Looks good to me, also tested locally building with podman-docker. In my setup I did not need sudo however, so I was thinking should it be included in shell script or not...

hackaugusto:
> Looks good to me, also tested locally building with podman-docker. In my setup I did not need sudo however, so I was thinking should it be included in shell script or not...

Yeah, I was not sure how to about that either. I preferred to keep the `sudo` in the script because I didn't want to run the whole script with sudo, specially because the previous implementation was not using the `.dockerignore` file and had a few calls to `rm`. I suppose we should see how that will work when we integrate that with the CI. Otherwise I don't think it is a big deal.

ivanyu:
Why do we need `Dockerfile.f31`?

container/start.sh
```diff
@@ -1,149 +1,80 @@
...
     "ssl_cafile": null,
     "ssl_certfile": null,
-    "ssl_keyfile": null,
-    "topic_name": "_schemas"
+    "ssl_keyfile": null
 }
 EOF
-  /opt/wait-for-it.sh "${kafka_host}"
-  python3 -m karapace.schema_registry_apis /opt/karapace.config.json 2>&1 | tee /var/log/karapace-registry.log
+  exec python3 -m karapace.schema_registry_apis /opt/karapace/registry.config.json 2>&1 | tee /var/log/karapace/registry.log
```
ivanyu, line 44: Do we really need to output logs to a file? We also don't have any log rotation, etc.

ivanyu: Also if removing this, I'm not 100% sure about `2>&1` - Docker should merge stdout and stderr itself.

hackaugusto:
> Why do we need Dockerfile.f31?

We don't actually, I had it there because it was my first iteration based on f31. Removing it from the source tree, if we ever need it again it should be in the git history.

container/build_image.sh
ivanyu: @hackaugusto do you plan to apply this?

hackaugusto: Yep, I think it makes sense.

ivanyu (approved):

PR #158, opened 2021-02-08
depedency: using accept-types from pypi

hackaugusto:
Currently `accept-types` is vendorized. This removes the code from our repo and install it from pypi.


karapace/rapu.py
```diff
@@ -195,7 +195,7 @@ def check_schema_headers(self, request: HTTPRequest):
             http_error("HTTP 415 Unsupported Media Type", response_default_content_type, 415)
         accept_val = request.get_header("Accept")
         if accept_val:
-            if accept_val == "*/*" or accept_val.startswith("*/"):
+            if accept_val in ("*/*", "*") or accept_val.startswith("*/"):
```
hackaugusto, line 198: This is an incompatibility with `schema-registry`.

Doing requests with these headers:

- `Content-Type: */*`
- `Content-Type: *`

Will result in a `{"error_code":415,"message":"HTTP 415 Unsupported Media Type"}`. karapace in the other hand will accept it.

This small change is because of what seems to be a problem with `accept-types` releases. The code we have in the vendorized folder is equal to the projects `master` branch code, which accepts `*`. The last release on PyPI is `0.4.1` which accepts `*/*`. This small patch is just to maintain compatibility with the previous karapace releases.

ivanyu (approved):

PR #159, opened 2021-02-08
Split requirements

hackaugusto:
This splits the development from the runtime requirements, the idea is to have a way to install only the runtime dependencies.


requirements.txt
```diff
@@ -0,0 +1,18 @@
...
+aiokafka==0.6.0
+jsonschema==3.2.0
+lz4==3.0.2
+requests==2.23.0
+
+# Patched dependencies
+#
+# Note: It is important to use commits to reference patched dependencies. This
+# has two advantages:
+# - Reproducible builds
```
hackaugusto, line 13: Note: The dependencies of our dependencies are not pinned. So this is not 100% reproducible, what I meant here is just to avoid issues with branch updates.

To fully pin the dependencies I would have to add another dependency to generate the `.lock` file, I have done this yet because I just want to get the docker image done.

ivanyu (approved):

PR #162, opened 2021-02-15
Mypy/Pylint update and `make clean` deleting python cache files 

hackaugusto:
- add mypy and update pylint
   - the new version of pylint caught a few more issues, namely calling `super` with arguments and a `raise` without a `from`
   - I also fixed some issues that mypy caught, but there are many left, so it can't be enabled by default yet.  
- clean also removing python cache files

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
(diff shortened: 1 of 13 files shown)


hackaugusto:
Note: I'm not sure why pylint is failing. I did experience the problem locally too, but after fixing the warnings the error went away.

hackaugusto:
closing this for now, I didn't investigate why the serialization is failing

PR #163, opened 2021-02-15
Socket port improvements

hackaugusto:
These are just some small cosmetic changes, I started doing this with the idea of randomizing the ports `1234, 5678` so that multiple tests runs could be done in parallel. (It is not guaranteed to be conflict free, as explained in the comments below, but the changes of collision are low enough that it works for me locally).

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -293,9 +293,9 @@ def fixture_registry_async_pair(session_tmpdir: TempDirCreator, kafka_server: Op
     )
     master_process = subprocess.Popen(["python", "-m", "karapace.karapace_all", master_config_path])
     slave_process = subprocess.Popen(["python", "-m", "karapace.karapace_all", slave_config_path])
-    wait_for_port(master_port)
-    wait_for_port(slave_port)
     try:
+        wait_for_port(master_port)
+        wait_for_port(slave_port)
         yield f"http://127.0.0.1:{master_port}", f"http://127.0.0.1:{slave_port}"
     finally:
         master_process.kill()
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -27,6 +27,38 @@ KAFKA_CURRENT_VERSION = "2.4"
 BASEDIR = "kafka_2.12-2.4.1"
 
 
+@dataclass(frozen=True)
+class PortRangeInclusive:
+    start: int
+    end: int
+
+    PRIVILEGE_END = 2 ** 10
+    MAX_PORTS = 2 ** 16 - 1
+
+    def __post_init__(self):
+        # Make sure the range is valid and that we don't need to be root
+        assert self.end > self.start, "there must be at least one port available"
+        assert self.end <= self.MAX_PORTS, f"end must be lower than {self.MAX_PORTS}"
+        assert self.start > self.PRIVILEGE_END, "start must not be a privileged port"
+
+    def next_range(self, number_of_ports: int) -> "PortRangeInclusive":
+        next_start = self.end + 1
+        next_end = next_start + number_of_ports - 1  # -1 because the range is inclusive
+
+        return PortRangeInclusive(next_start, next_end)
+
+
+# To find a good port range use the following:
+#
+#   curl --silent 'https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt' | \
+#       egrep -i -e '^\s*[0-9]+-[0-9]+\s*unassigned' | \
+#       awk '{print $1}'
+#
+KAFKA_PORTS = PortRangeInclusive(48700, 48800)
+ZK_PORT_RANGE = KAFKA_PORTS.next_range(100)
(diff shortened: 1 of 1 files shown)


tests/integration/conftest.py
```diff
@@ -82,43 +128,62 @@ def port_is_listening(hostname: str, port: int, ipv6: bool) -> bool:
...
+        This function is *not* aware of the ports currently open in the system,
+        the blacklist only prevents two services of the same type to randomly
+        get the same ports for *a single test run*.
+
+        Because of that, the port range should be chosen such that there is no
+        system service in the range. Also note that running two sessions of the
+        tests with the same range is not supported and will lead to flakiness.
+    """
+    value = random.randint(port_range.start, port_range.end)
+    while value in blacklist:
```
HelenMel, line 184: [Small note] It looks for me that in a scenarios I saw later, blacklist is not so necessary: As far as I know two adjacent calls or random.randint() return the same value with a very low probability (if range is wide, of course).

This approach is totally fine, but I would rather write a global generator in `BeforeAll` and  withdraw next available port each time. Totally up to you.

HelenMel:
Looks good. Left one tiny comment.

tests/integration/conftest.py
```diff
@@ -82,43 +128,62 @@ def port_is_listening(hostname: str, port: int, ipv6: bool) -> bool:
...
         try:
-            client = KafkaAdminClient(bootstrap_servers=bootstrap_server)
-            _ = client.list_topics()
-            return
-        # pylint: disable=broad-except
-        except Exception as e:
+            KafkaAdminClient(bootstrap_servers=bootstrap_server).list_topics()
+        except Exception as e:  # pylint: disable=broad-except
             print(f"Error checking kafka cluster: {e}")
             time.sleep(2.0)
```
HelenMel, line 151: What do you think about making 2.0 a variable? I had a feeling that with a tests amount increased it will change

hackaugusto: Do you mean the amount of time we wait for kafka in total? This is the time we wait in between checks. The total amount of time we wait for kafka is the `wait_time` variable.

HelenMel: I mean between checks. I mean why 2 seconds? Could we wait less? If we have more tests this could make whole suit very slow, no?

HelenMel (approved):

tests/integration/conftest.py
HelenMel: Sorry! Wrong place, I was going to leave it wait_for_port, because it is called often, and 2.0 seconds could be too much. (If it is not ready for the first time - 4 seconds... etc) So, it could be tuned

tests/integration/conftest.py
```diff
@@ -82,43 +128,62 @@ def port_is_listening(hostname: str, port: int, ipv6: bool) -> bool:
...
-        if time.monotonic() - start_time > wait_time:
-            raise Timeout("Timeout waiting for port {} on host {}".format(port, hostname))
+    expiration = Expiration(
+        msg=f"Timeout waiting for `{hostname}:{port}`",
+        deadline=start_time + wait_time,
+    )
+
+    while not port_is_listening(hostname, port, ipv6):
+        expiration.raise_if_expired()
         time.sleep(2.0)
```
HelenMel, line 165: I was going to leave a comment here

tests/integration/conftest.py
hackaugusto: > As far as I know two adjacent calls or random.randint() return the same value with a very low probability (if range is wide, of course).

Using a deterministic series with global states breaks with parallelization, and I'm trying to go towards parallelization to speed up the tests.

This code is far from perfect, there is actually a non-negligible chance of getting a conflict, specially because of parallelization and some repeated runs. Ideally the start port number should be provided externally to the test runner, then we would draw the ports in the fashion you described, and even then try to check if the port is bound before using it.

I just didn't want to introduce these additional changes in this PR that was getting big already.

tests/integration/conftest.py
hackaugusto: > I mean why 2 seconds?

I suppose this is totally arbitrary. I have the impression it takes something like 6 seconds for kafka to become available, perhaps that is why this value was chosen.

> Could we wait less?

Most definitely.

> If we have more tests this could make whole suit very slow, no?

At most by `2.0` seconds.

> Because wait_for_kafka will probably used in every test

This is used only for the session fixture, which runs once before any tests. It just starts a kafka process and waits for it to be available before running the tests that depend on it.

> That is why, maybe, make this value tunable

Making it tunable for this function is not as useful as it first seems.

The problem is that this function is called by fixtures only. Configuring fixtures with tunable values is really confusing for someone that is not used to it because the call graph is implicit. Fixtures describe dependencies, these is the used to build a dependency graph, the execution of these fixtures is not shown in the code but it is rather determined at run-time. Using fixtures to tune the behavior of the build becomes rather quickly untenable. I can just decrease this to another arbitrary value, say `0.3`.

PR #164, opened 2021-02-16
tests: added test for accept header regex

hackaugusto:
As the PR title says. The goal here is more to have examples of the matching strings to make it easier to understand what to expect from the regex groups


tvainika (approved):
LGTM

PR #165, opened 2021-02-17
Added typing to the SchemaCompatibilityResult class

hackaugusto:
This does a few things

- Simplifies the code of `Compatibility`, it can be just a single function call
- Adds typing info and some docs to `SchemaCompatibilityResult`, I want to reuse this class with JSON schema

The goal was not to change any behavior just to document and simplify. I'm saying this because `check_compatibility` will need some refactoring. ATM this function is used only for the exceptions it raises, IOW, the return value is ignored by every caller. It makes a bit hard to reason if the callers are working as intended because `False` is returned is for unsupported formats (e.g. json schema and protobuf). I believe the code is behaving correctly because there are checks earlier in the code, but still, local reasoning is not great here.

### DIFF
--- a/karapace/avro_compatibility.py
+++ b/karapace/avro_compatibility.py
@@ -2,7 +2,6 @@ from avro.schema import (
     ARRAY, ArraySchema, BOOLEAN, BYTES, DOUBLE, ENUM, EnumSchema, Field, FIXED, FixedSchema, FLOAT, INT, LONG, MAP,
     MapSchema, NamedSchema, Names, NULL, RECORD, RecordSchema, Schema, SchemaFromJSONData, STRING, UNION, UnionSchema
 )
-from copy import copy
 from enum import Enum
 from typing import Any, cast, Dict, List, Optional, Set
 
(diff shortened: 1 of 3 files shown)


karapace/avro_compatibility.py
```diff
@@ -52,19 +51,23 @@ class SchemaCompatibilityResult:
     def __init__(
         self,
         compatibility: SchemaCompatibilityType = SchemaCompatibilityType.recursion_in_progress,
-        incompatibilities: List[SchemaIncompatibilityType] = None,
+        incompatibilities: Optional[List[SchemaIncompatibilityType]] = None,
```
hackaugusto, line 54: Note that the annotation for `Optional` is not required when used with a default value `= None`, ref:

https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling

I usually prefer to be explicit, this way I can apply just one heuristic on annotations. This also has the side benefit of allowing me to easily copy the type to datastructures (e.g. if I'm refactoring a class to a dataclass, or creating a namedtuple to call a function in a threadpool)

tvainika (approved):
LGTM

PR #166, opened 2021-02-17
Schema registry APIs using enum for error codes

hackaugusto:
My idea here is to remove the magic constants that we have through out the code and give them a name, the end goal is to use these variables in the tests so that when a test fails the enum name gives a hint on what went wrong.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -26,6 +26,21 @@ class CompatibilityModes(Enum):
     NONE = "NONE"
 
 
+@unique
+class SchemaErrorCodes(Enum):
+    HTTP_NOT_FOUND = 404
+    HTTP_CONFLICT = 409
+    HTTP_UNPROCESSABLE_ENTITY = 422
+    HTTP_INTERNAL_SERVER_ERROR = 500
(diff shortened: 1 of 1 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -284,7 +286,14 @@ def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
                     self._cluster_metadata = self.admin_client.cluster_metadata(topics)
                 except NodeNotReadyError:
                     self.log.exception("Could not refresh cluster metadata")
-                    self.r({"message": "Kafka node not ready", "code": 500}, "application/json", 500)
+                    self.r(
+                        body={
+                            "message": "Kafka node not ready",
+                            "code": RESTErrorCodes.HTTP_INTERNAL_SERVER_ERROR.value,
```
hackaugusto, line 292: I avoided using `class RESTErrorCodes(int, Enum)` because that is a very leaky abstraction. Having a few more characters to type seems an okay trade off to avoid unexpected errors.

ivanyu (approved):

PR #175, opened 2021-03-12
rest api integrate jsonschema compatibility

hackaugusto:
~Depends on https://github.com/aiven/karapace/pull/171~ merged.

This integrates the jsonschema compatibility check with the existing rest apis.


karapace/avro_compatibility.py
```diff
@@ -2,13 +2,15 @@
     ARRAY, ArraySchema, BOOLEAN, BYTES, DOUBLE, ENUM, EnumSchema, Field, FIXED, FixedSchema, FLOAT, INT, LONG, MAP,
     MapSchema, NamedSchema, Names, NULL, RECORD, RecordSchema, Schema, SchemaFromJSONData, STRING, UNION, UnionSchema
 )
-from enum import Enum
-from typing import Any, cast, Dict, List, Optional, Set
+from enum import Enum, unique
+from typing import Any, cast, Dict, Generic, List, Optional, Set, TypeVar
```
HelenMel, line 6: mmm....Generics...things become interesting

karapace/avro_compatibility.py
```diff
@@ -2,13 +2,15 @@
...
 
 import json
 
+E = TypeVar("E", bound=Enum)
 
-def parse_json_ignore_trailing(s: str) -> Any:
+
+def parse_avro_schema_definition(s: str) -> Any:
```
HelenMel, line 13: Can we narrow down Any, to something a bit more specific?

karapace/avro_compatibility.py
```diff
@@ -28,12 +30,24 @@ def parse_json_ignore_trailing(s: str) -> Any:
     return SchemaFromJSONData(json_data, names)
 
 
+def is_incompatible(result: "SchemaCompatibilityResult") -> bool:
+    return result.compatibility is SchemaCompatibilityType.incompatible
+
+
+def is_compatible(result: "SchemaCompatibilityResult") -> bool:
```
HelenMel, line 37: Do you actually need both? Does is_compatible != not is_incompatible?

karapace/avro_compatibility.py
```diff
@@ -83,14 +97,21 @@ def merged_with(self, that: "SchemaCompatibilityResult") -> "SchemaCompatibility
...
+
     @staticmethod
     def compatible() -> "SchemaCompatibilityResult":
         return SchemaCompatibilityResult(SchemaCompatibilityType.compatible)
 
     @staticmethod
-    def incompatible(
-        incompat_type: SchemaIncompatibilityType, message: str, location: List[str]
-    ) -> "SchemaCompatibilityResult":
+    def incompatible(incompat_type: E, message: str, location: List[str]) -> "SchemaCompatibilityResult":
```
HelenMel, line 114: Could from __future__ import annotations be useful here?

karapace/compatibility/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+            location=[],
+        )
+
+    if compatibility_mode is CompatibilityModes.NONE:
+        LOG.info("Compatibility level set to NONE, no schema compatibility checks performed")
+        return SchemaCompatibilityResult.compatible()
+
+    if source.schema_type is SchemaType.AVRO:
+        if compatibility_mode in {CompatibilityModes.BACKWARD, CompatibilityModes.BACKWARD_TRANSITIVE}:
+            result = check_avro_compatibility(reader_schema=target.schema, writer_schema=source.schema)
```
HelenMel, line 71: Maybe just use a return statement? Here and below?

karapace/avro_compatibility.py
hackaugusto: yes, there are more than two states:

https://github.com/aiven/karapace/blob/4737f82cb685444aaa49d6aabb5edc0a6e0392ea/karapace/avro_compatibility.py#L31-L34

I'm not sure how important the state `recursion_in_progress` is for the Avro compatibility check, so I have not changed it.

karapace/avro_compatibility.py
hackaugusto: I tend to use `__future__` imports only for forwards compatibility, which is not an issue anymore since python 2.7 reached its EOL. I don't see any problems with using it ... I also don't see any great advantages for the change here, so usually I just don't bother.

karapace/compatibility/__init__.py
hackaugusto: I usually employ a poor man's design-by-contract approach. Were I have checks on the top of the function for the preconditions with early-returns, followed by the main code. This usually makes the indentation level shallow and caught issues early. 

For the main code I don't have a hard rule. Here I just didn't use many `return` statements to reduce the cyclomatic complexity ... not that important but it keeps the linting tools happy. (pylint complains if there are too many return statements)

karapace/avro_compatibility.py
hackaugusto: Done :)

karapace/avro_compatibility.py
```diff
@@ -10,7 +10,7 @@
 E = TypeVar("E", bound=Enum)
 
 
-def parse_json_ignore_trailing(s: str) -> Any:
+def parse_avro_schema_definition(s: str) -> Schema:
```
hackaugusto, line 13: renamed both this and `parse_schema_definition` to avoid confusiong, now the names have the format name in it.

karapace/compatibility/__init__.py
```diff
@@ -34,36 +37,62 @@ def is_transitive(self) -> bool:
...
 
 
-def check_compatibility(source: TypedSchema, target: TypedSchema, compatibility_mode: CompatibilityModes) -> None:
+def check_jsonschema_compatibility(reader: Draft7Validator, writer: Draft7Validator) -> SchemaCompatibilityResult:
+    return jsonschema_compatibility(reader, writer)
+
+
+def check_compatibility(
+    source: TypedSchema, target: TypedSchema, compatibility_mode: CompatibilityModes
+) -> SchemaCompatibilityResult:
```
hackaugusto, line 57: Instead of raising, now the error is returned. This allows the API to report the error message.

karapace/compatibility/jsonschema/parse.py
```diff
@@ -1,14 +0,0 @@
-from jsonschema import Draft7Validator
-
-import json
-
-
-def parse_schema_definition(schema_definition: str) -> Draft7Validator:
```
hackaugusto, line 6: I had to move the function to avoid circular imports

tests/integration/conftest.py
```diff
@@ -173,10 +181,10 @@ def fixture_zkserver(session_tmppath: Path) -> Iterator[Optional[ZKConfig]]:
     # transfer_file (primarily the server's port number)
     with FileLock(str(lock_path_for(zk_dir))):
         if transfer_file.exists():
-            config = pickle.loads(transfer_file.read_bytes())
```
hackaugusto, line 176: this was breaking when running the tests with `pytest tests`, it only worked if `pytests tests/integration` was used because of how `pytest` messes with the import paths.

karapace/compatibility/__init__.py
```diff
@@ -34,36 +37,62 @@ def is_transitive(self) -> bool:
...
+            result = check_avro_compatibility(reader_schema=target.schema, writer_schema=source.schema)
+
+        elif compatibility_mode in {CompatibilityModes.FORWARD, CompatibilityModes.FORWARD_TRANSITIVE}:
+            result = check_avro_compatibility(reader_schema=source.schema, writer_schema=target.schema)
+
+        elif compatibility_mode in {CompatibilityModes.FULL, CompatibilityModes.FULL_TRANSITIVE}:
+            result = check_avro_compatibility(reader_schema=target.schema, writer_schema=source.schema)
+            result = result.merged_with(check_avro_compatibility(reader_schema=source.schema, writer_schema=target.schema))
+
+    elif source.schema_type is SchemaType.JSONSCHEMA:
```
hackaugusto, line 80: Adding this is pretty much all that was needed to support jsonschema in the APIs, the rest of the code was already in place.

karapace/compatibility/jsonschema/utils.py
```diff
@@ -86,11 +79,18 @@ def maybe_get_subschemas_and_type(schema: Any) -> Optional[Tuple[List[Any], Subs
 
         return (normalized_schemas, Subschema.ANY_OF)
 
-    for type_ in (Subschema.ALL_OF, Subschema.ANY_OF, Subschema.ONE_OF, Subschema.NOT):
+    for type_ in (Subschema.ALL_OF, Subschema.ANY_OF, Subschema.ONE_OF):
         subschema = schema.get(type_.value)
         if subschema is not None:
+            # https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.2.1
+            assert isinstance(subschema, list), "allOf/anyOf/oneOf must be an array"
```
hackaugusto, line 86: This was caught by mypy, not by a test.

I had to add the `isinstance' check to narrow the type, however that assert fails for `NOT`, so the two cases had to be split and it fixed the bug with `not` (I probably need to also add a unit test for that)

karapace/avro_compatibility.py
HelenMel: That is unusual. I would say that recursion_in_progress is another type, that just wrapped around compatibility type. But it is not important right now.

karapace/compatibility/__init__.py
HelenMel: Ok. Lint would be surprised, that his function is complex even without multiple return statements. and it is not trivial to make it simpler.

HelenMel:
LGTM

hackaugusto:
> LGTM. Do you have an intent to add something more?

Nope

HelenMel (approved):

PR #177, opened 2021-03-16
tests: replaced pickle with json

hackaugusto:
- pickle encodes the object's absolute module location while encoding, and
  use it while decoding.
- pytest will change the top-level module depending on the arguments it
  is provided, breaking pickle decoding.

This changes to use json, since that does not depends on the above.


tvainika (approved):
I'd personally use `from __future__ import annotations` to avoid quoting typenames, but I saw you already had discussion about such preferences in another PR, so merging as it is.

PR #179, opened 2021-03-23
Sync kafka settings

hackaugusto:
- update kafka/zookeeper container versions
- change from cp-server to cp-kafka which is the community edition
- made sure the docker-compose and the pytest fixtures use the same settings


tests/integration/conftest.py
```diff
@@ -23,8 +23,10 @@
 import socket
 import time
 
+# Keep these in sync with the Makefile
 KAFKA_CURRENT_VERSION = "2.4"
-BASEDIR = "kafka_2.12-2.4.1"
+BASEDIR = "kafka_2.13-2.7.0"
```
tvainika, line 28: The line above has `KAFKA_CURRENT_VERSION = "2.4"`, but this PR switches to Kafka 2.7. I see that KAFKA_CURRENT_VERSION is used only for protocol version related settings, but reflecting to comment line before is confusing. What software versions and protocol versions should be defined and where?

tvainika (changes requested):
Version sync comments and version usage needs a bit clarification

tests/integration/conftest.py
hackaugusto: I'm not sure why `KAFKA_CURRENT_VERSION` is necessary, I just synchronized the files. Unfortunately setting it to `2.7.0` crashes Kafka

tvainika: I'm fine with updating the comments here (and possibly in Makefile) about what versions need to be in sync. Also what about renaming  `KAFKA_CURRENT_VERSION` to something like `KAFKA_PROTOCOL_VERSION`?

hackaugusto: I must have used an outdated version of the container image. I did check the documentation and tried it again, and the protocol version set as the current version works.

tvainika (approved):

PR #183, opened 2021-03-23
Fix test master selection and tests with timings issues for topics

hackaugusto:
~Depends on  #180~

The test for master selection couldn't be run in parallel because the port numbers, group ids, and clients ids conflicted. This changes the test to use random values fixing the flakiness.

I tested this with:

```
(cd containers; sudo docker-compose --project-name karapace up -d)
while [[ $? == 0 ]]; do pytest tests -vvv -x --registry-url http://localhost:8081 --rest-url http://localhost:8082/ --kafka-bootstrap-servers localhost:9092 -k test_master_selection; done
```

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -15,13 +15,15 @@ from karapace.schema_registry_apis import KarapaceSchemaRegistry
 from karapace.utils import Client
 from pathlib import Path
 from subprocess import Popen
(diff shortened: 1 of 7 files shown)


tvainika:
`FAILED tests/integration/test_master_coordinator.py::test_master_selection[lowest]` was only failed test, supposed to be fixed by this PR, but apparently not

tvainika (approved):

PR #184, opened 2021-03-23
Fix test flakiness because of timming issues

hackaugusto:
depends on #180

When running the tests with external services it may take a little while for the services to pick up freshly created topics.


hackaugusto:
moved this commit to #183 because it needs the Expiration class.

PR #186, opened 2021-03-23
tests: cosmetic change, consistently using new_random_name

hackaugusto:
This is only a cosmetic change, the only benefit is to have a more descriptive prefix for the subjects.


tvainika (changes requested):
Lint failures

tvainika (approved):

PR #188, opened 2021-03-25
Early log level configuration

hackaugusto:
Configuring the logging in the base class constructor interferes with
the testing runner configuration, it effectively overwrites pytest's
configuration making the flag `--log-cli-level` unusable.


tests/integration/conftest.py
```diff
@@ -177,11 +177,7 @@ async def fixture_rest_async(tmp_path: Path, kafka_servers: KafkaServers,
                              registry_async_client: Client) -> AsyncIterator[KafkaRest]:
     config_path = tmp_path / "karapace_config.json"
 
-    config = set_config_defaults({
-        "log_level": "WARNING",
```
hackaugusto, line 181: This setting doesn't have an effect any longer. They can be configured with `--log-cli-level`

tvainika (approved):

PR #190, opened 2021-03-25
rest: fix error code

hackaugusto:
it should be 40403 and not 40401


tvainika (changes requested):
There are set of error codes in karapace/kafka_rest_apis/error_codes.py here https://github.com/aiven/karapace/blob/09d426d1db3a41e7009371abffb5da77311e142f/karapace/kafka_rest_apis/error_codes.py#L10

I think using those enum error codes would make sense here, also this PR should explain why 40403 is correct and not 40401 for referencing purposes.

hackaugusto:
> I think using those enum error codes would make sense here

I agree, but ATM I'm not sure how to resolve the naming conflict. The existing names for the error codes were based on how Karapace was using these codes, now we have a `CONSUMER_NOT_FOUND` with the value 40403:

https://github.com/aiven/karapace/blob/09d426d1db3a41e7009371abffb5da77311e142f/karapace/kafka_rest_apis/error_codes.py#L12

However that name does not make sense in the context of a `UnknownTopicOrPartitionError`.

>  PR should explain why 40403 is correct and not 40401 for referencing purposes.

Ran the test using `--registry-url http://localhost:8081 --rest-url http://localhost:8082/` against a schema registry node.

HelenMel (approved):

tvainika:
> I agree, but ATM I'm not sure how to resolve the naming conflict. The existing names for the error codes were based on how Karapace was using these codes, now we have a `CONSUMER_NOT_FOUND` with the value 40403:
> 
> However that name does not make sense in the context of a `UnknownTopicOrPartitionError`.

Would it make sense to remove `@unique` and just define `UNKNOWN_TOPIC_OR_PARTITION = 40403`?

> > PR should explain why 40403 is correct and not 40401 for referencing purposes.
> 
> Ran the test using `--registry-url http://localhost:8081 --rest-url http://localhost:8082/` against a schema registry node.
 
Yes, I can understand it should be same as Confluent's Schema Registry, but do we have any reference or specifications to follow? Probably not that nor automated test suite to compare.

hackaugusto:
> Yes, I can understand it should be same as Confluent's Schema Registry, but do we have any reference or specifications to follow?

AFAIK only the HTTP response codes are documented. https://docs.confluent.io/platform/current/schema-registry/develop/api.html#subjects 

> Probably not that nor automated test suite to compare

I'm not sure if that is what you meant, but I used our integration tests to compare our expected results against their API. Unfortunately we are most likely not covering every case, but that is what caught the difference here.

> Would it make sense to remove @unique and just define UNKNOWN_TOPIC_OR_PARTITION = 40403?

sure, will do :)

tests/integration/test_rest.py
```diff
@@ -199,7 +199,7 @@ async def test_topics(rest_async_client, admin_client):
     assert data["partitions"][0]["replicas"][0]["in_sync"], "Replica should be in sync"
     res = await rest_async_client.get(f"/topics/{topic_foo}")
     assert res.status_code == 404, f"Topic {topic_foo} should not exist, status_code={res.status_code}"
-    assert res.json()["error_code"] == 40401, "Error code does not match"
+    assert res.json()["error_code"] == 40403, "Error code does not match"
```
hackaugusto, line 202: I didn't want to update the whole `test_rest.py` right now, there is lots of places using the hardcoded constant and not the enum there.

tvainika (approved):

PR #193, opened 2021-03-29
Documentation for compatibility modes

hackaugusto:
This just adds a bit of documentation in the code about the direction of the compatibility modes.

### DIFF
--- a/karapace/compatibility/__init__.py
+++ b/karapace/compatibility/__init__.py
@@ -20,6 +20,16 @@ LOG = logging.getLogger(__name__)
 
 @unique
 class CompatibilityModes(Enum):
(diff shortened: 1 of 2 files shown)


karapace/compatibility/__init__.py
```diff
@@ -53,45 +63,74 @@ def check_jsonschema_compatibility(reader: Draft7Validator, writer: Draft7Valida
 
 
 def check_compatibility(
-    source: TypedSchema, target: TypedSchema, compatibility_mode: CompatibilityModes
```
hackaugusto, line 56: I did this change mostly because of these names, I always had to double check what was the source and target schemas.

tvainika (approved):

PR #199, opened 2021-04-16
schema_backup: do not treat empty files as an error

hackaugusto:
It is possible to generate empty backup files, this should not be
treated as an error.


vtainio (approved):
:+1:

PR #205, opened 2021-04-28
JSON Schema pattern properties tests

tvainika:

Add separate `patternProperties` tests before implementing `propertyNames` compatibility and tests.  Also rename test schema objects for clarification.


hackaugusto (approved):
LGTM

PR #206, opened 2021-04-29
JSON Schema propertyNames compatibility

tvainika:

Fix `propertyNames` support: An object with `propertyNames` setting constraints on property names is still open ie. there are no limitations on property values for those.

Add separate test for object properties without `propertyNames` and fix test to match existing integration test.

### DIFF
--- a/tests/unit/test_json_schema.py
+++ b/tests/unit/test_json_schema.py
@@ -25,8 +25,6 @@ from tests.schemas.json_schemas import (
     TUPLE_OF_INT_SCHEMA, TUPLE_OF_INT_WITH_ADDITIONAL_INT_SCHEMA, TYPES_STRING_INT_SCHEMA, TYPES_STRING_SCHEMA
 )
 
-import pytest
-
 COMPATIBLE = SchemaCompatibilityResult.compatible()
 
 COMPATIBILIY = "compatibility with schema registry"
@@ -1323,7 +1321,6 @@ def test_object_properties():
     )
 
 
-@pytest.mark.skip("not implemented yet")
 def test_property_names():
     schemas_are_compatible(
         reader=OBJECT_SCHEMA,
--- a/tests/unit/test_json_schema.py
+++ b/tests/unit/test_json_schema.py
@@ -1302,7 +1302,7 @@ def test_object_properties():
     not_schemas_are_compatible(
         reader=A_OBJECT_SCHEMA,
         writer=OBJECT_SCHEMA,
-        msg=COMPATIBILIY,
+        msg=INCOMPATIBLE_READER_RESTRICTED_ACCEPTED_VALUES,
     )
     schemas_are_compatible(
         reader=OBJECT_SCHEMA,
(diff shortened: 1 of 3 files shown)


tests/unit/test_json_schema.py
```diff
@@ -1300,8 +1298,30 @@ def test_pattern_properties():
...
+    )
+    not_schemas_are_compatible(
+        reader=A_INT_OBJECT_SCHEMA,
+        writer=OBJECT_SCHEMA,
+        msg=INCOMPATIBLE_READER_RESTRICTED_ACCEPTED_VALUES,
+    )
+    not_schemas_are_compatible(
+        reader=B_INT_OBJECT_SCHEMA,
+        writer=OBJECT_SCHEMA,
+        msg=INCOMPATIBLE_READER_IS_CLOSED_AND_REMOVED_FIELD,
```
hackaugusto, line 1320: here the message should be `INCOMPATIBLE_READER_RESTRICTED_ACCEPTED_VALUES`

tests/unit/test_json_schema.py
```diff
@@ -1300,8 +1298,30 @@ def test_pattern_properties():
     )
 
 
-@pytest.mark.skip("not implemented yet")
-def test_property_name():
+def test_object_properties():
+    not_schemas_are_compatible(
+        reader=A_OBJECT_SCHEMA,
+        writer=OBJECT_SCHEMA,
+        msg=COMPATIBILIY,
```
hackaugusto, line 1305: here the message should be `INCOMPATIBLE_READER_RESTRICTED_ACCEPTED_VALUES`

tests/unit/test_json_schema.py
```diff
@@ -1300,8 +1298,30 @@ def test_pattern_properties():
...
+    )
+    schemas_are_compatible(
+        reader=OBJECT_SCHEMA,
+        writer=A_OBJECT_SCHEMA,
+        msg=COMPATIBILIY,
```
hackaugusto, line 1310: here the message should be `INCOMPATIBLE_READER_RESTRICTED_ACCEPTED_VALUES`

hackaugusto (changes requested):

tvainika:
> Open is used to say if unknown properties are allowed, not if the existing ones have constraints in their values, i.e. if `additionalProperties` is not `false`

Maybe I used wording about openess wrongly, but yet schema with `propertyNames` constraint only makes backward compatible changes ie. adding _optional_ fields or deleting fields, not adding required fields. Thus the logic is fine, right?

hackaugusto:
> but yet schema with propertyNames constraint only makes backward compatible changes

I think adding `propertyNames` is backwards *incompatible*. Say the schema for the writer allows anything:
```
{}
```
This would allow a producer to generate `{"a":1}` and `{"b":2}`. Now the schema is updated with a `propertyNames`:
```
{
  "propertyNames": {
    "pattern": "^a+$"
  }
}
```
Now the value `{"b":2}` is not accepted anymore, thus a incompatible change.

hackaugusto (approved):

tvainika:
And to document for later usage, the implementation and the interpretation of schema compatibility now follows Confluent's Schema Registry's compatibility table even if it is written from Avro point of view (Avro does not have similar `propertyNames` pattern concept):
https://docs.confluent.io/platform/current/schema-registry/avro.html#compatibility-types

PR #209, opened 2021-05-05
docker: add script to publish image

hackaugusto:
This script can be used to publish an image based on a tag or commit:

> ./container/publish_image.sh 863bf4a7faec3b13eba57a0864e63f5e416cd6e3

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -141,20 +141,9 @@ Set up the configuration file in ``karapace.config.json`` to include connection 
 Docker setup for development
 ----------------------------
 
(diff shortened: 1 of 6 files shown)


hackaugusto:
Converted to draft because docker-compose needs some work

tvainika (approved):
Looks fine. I think main level README could contain `docker run` or `docker-compose up` commands in more detail to show how to pull pre-build container image and get it up and running faster. This can be addressed in separate PR though.

PR #210, opened 2021-05-06
Support for hard delete versions and schemas

### LINKED ISSUE #196: Add support for hard delete of schemas through the REST API
When the user sends a `DELETE` request with the get parameter `?permanent=true`, the schema should be hard deleted by sending `null` as value, instead of setting the flag `deleted` to true. Support for handling these messages is already in place:

https://github.com/aiven/karapace/blob/61b700e92fbf459e84a57995b95a230bfb80bb26/karapace/schema_reader.py#L287

- [ ] Add support for the flag to the `DELETE /subject/<subject>/version/<version>` endpoint (the hard delete should only allowed after a soft delete) https://github.com/aiven/karapace/blob/61b700e92fbf459e84a57995b95a230bfb80bb26/karapace/schema_registry_apis.py#L447
- [ ] Write an integration test for it

tvainika:

Support hard delete (`permanent=true`) flag for deleting versions and schemas.  This has some memory impact as soft deleted schemas are stored in memory for longer period of time now, but it is estimated to be negligible.  Fixes #196.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -361,7 +361,9 @@ class KafkaSchemaReader(Thread):
             schema["deleted"] = True
         return schema
 
-    def get_schemas(self, subject):
+    def get_schemas(self, subject, *, include_deleted=False):
+        if include_deleted is True:
+            return self.subjects[subject]["schemas"]
         non_deleted_schemas = {
             key: val
             for key, val in self.subjects[subject]["schemas"].items()
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -10,6 +10,7 @@ from karapace.master_coordinator import MasterCoordinator
 from karapace.rapu import HTTPRequest
 from karapace.schema_reader import InvalidSchema, KafkaSchemaReader, SchemaType, TypedSchema
 from karapace.utils import json_encode
+from typing import Optional
 
 import argparse
 import asyncio
@@ -27,6 +28,10 @@ class SchemaErrorCodes(Enum):
     SUBJECT_NOT_FOUND = 40401
     VERSION_NOT_FOUND = 40402
     SCHEMA_NOT_FOUND = 40403
+    SUBJECT_SOFT_DELETED = 40404
+    SUBJECT_NOT_SOFT_DELETED = 40405
+    SCHEMAVERSION_SOFT_DELETED = 40406
+    SCHEMAVERSION_NOT_SOFT_DELETED = 40407
     INVALID_VERSION_ID = 42202
     INVALID_COMPATIBILITY_LEVEL = 42203
     INVALID_AVRO_SCHEMA = 44201
@@ -86,7 +91,9 @@ class KarapaceSchemaRegistry(KarapaceBase):
             "/subjects/<subject:path>/versions/<version:path>",  # needs
             callback=self.subject_version_delete,
             method="DELETE",
-            schema_request=True
+            schema_request=True,
+            with_request=True,
+            json_body=False,
         )
         self.route(
             "/subjects/<subject:path>/versions/<version>/schema",
@@ -94,7 +101,14 @@ class KarapaceSchemaRegistry(KarapaceBase):
             method="GET",
             schema_request=True
         )
-        self.route("/subjects/<subject:path>", callback=self.subject_delete, method="DELETE", schema_request=True)  # needs
+        self.route(
+            "/subjects/<subject:path>",
+            callback=self.subject_delete,
+            method="DELETE",
+            schema_request=True,
+            with_request=True,
+            json_body=False,
+        )
 
     def close(self):
         super().close()
@@ -114,29 +128,30 @@ class KarapaceSchemaRegistry(KarapaceBase):
         self.mc = MasterCoordinator(config=self.config)
         self.mc.start()
 
-    def _subject_get(self, subject, content_type):
+    def _subject_get(self, subject, content_type, include_deleted=False):
         subject_data = self.ksr.subjects.get(subject)
         if not subject_data:
             self.r(
                 body={
                     "error_code": SchemaErrorCodes.SUBJECT_NOT_FOUND.value,
-                    "message": "Subject not found.",
+                    "message": f"Subject '{subject}' not found.",
                 },
                 content_type=content_type,
                 status=HTTPStatus.NOT_FOUND,
             )
 
-        schemas = self.ksr.get_schemas(subject)
+        schemas = self.ksr.get_schemas(subject, include_deleted=include_deleted)
         if not schemas:
             self.r(
                 body={
                     "error_code": SchemaErrorCodes.SUBJECT_NOT_FOUND.value,
-                    "message": "Subject not found.",
+                    "message": f"Subject '{subject}' not found.",
                 },
                 content_type=content_type,
                 status=HTTPStatus.NOT_FOUND,
             )
 
+        subject_data = subject_data.copy()
         subject_data["schemas"] = schemas
         return subject_data
 
@@ -209,16 +224,26 @@ class KarapaceSchemaRegistry(KarapaceBase):
         self,
         *,
         subject: str,
-        schema: TypedSchema,
+        schema: Optional[TypedSchema],
         schema_id: int,
         version: int,
         deleted: bool,
     ):
         key = '{{"subject":"{}","version":{},"magic":1,"keytype":"SCHEMA"}}'.format(subject, version)
-        value = {"subject": subject, "version": version, "id": schema_id, "schema": schema.schema_str, "deleted": deleted}
-        if schema.schema_type is not SchemaType.AVRO:
-            value["schemaType"] = schema.schema_type
-        return self.send_kafka_message(key, json_encode(value, compact=True))
+        if schema is not None:
+            valuedict = {
+                "subject": subject,
+                "version": version,
+                "id": schema_id,
+                "schema": schema.schema_str,
+                "deleted": deleted
+            }
+            if schema.schema_type is not SchemaType.AVRO:
+                valuedict["schemaType"] = schema.schema_type
+            value = json_encode(valuedict, compact=True)
+        else:
+            value = ""
+        return self.send_kafka_message(key, value)
 
     def send_config_message(self, compatibility_level: CompatibilityModes, subject=None):
         if subject is not None:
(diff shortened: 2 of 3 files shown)


karapace/schema_registry_apis.py
```diff
@@ -114,29 +128,30 @@ def _create_master_coordinator(self):
...
                 body={
                     "error_code": SchemaErrorCodes.SUBJECT_NOT_FOUND.value,
-                    "message": "Subject not found.",
+                    "message": f"Subject '{subject}' not found.",
                 },
                 content_type=content_type,
                 status=HTTPStatus.NOT_FOUND,
             )
 
+        subject_data = subject_data.copy()
```
juha-aiven, line 154: Why is this now needed?

tvainika: It is in function `_subject_get` which without this modified internal data structures by eliminating soft deleted schema versions from memory permanently.  Ie. this is the one causing memory impact mentioned in the description.

karapace/schema_registry_apis.py
```diff
@@ -395,16 +420,39 @@ async def subjects_list(self, content_type):
...
 
-    async def subject_delete(self, content_type, *, subject):
-        self._subject_get(subject, content_type)
-        version_list = list(self.ksr.get_schemas(subject))
+    async def subject_delete(self, content_type, *, subject, request: HTTPRequest):
+        permanent = request.query.get("permanent", "false").lower() == "true"
+
+        subject_data = self._subject_get(subject, content_type, include_deleted=permanent)
+
+        if permanent is True and [
```
juha-aiven, line 428: `if permanent  and`

No need for `is True`.

karapace/schema_reader.py
```diff
@@ -361,7 +361,9 @@ def _delete_schema_below_version(schema, version):
             schema["deleted"] = True
         return schema
 
-    def get_schemas(self, subject):
+    def get_schemas(self, subject, *, include_deleted=False):
```
juha-aiven, line 364: Or just change only this line 

```python
if include_deleted or val.get("deleted", False) is False
```

There'd be only a single execution path.

karapace/schema_registry_apis.py
```diff
@@ -209,16 +224,26 @@ def send_schema_message(
...
         schema_id: int,
         version: int,
         deleted: bool,
     ):
         key = '{{"subject":"{}","version":{},"magic":1,"keytype":"SCHEMA"}}'.format(subject, version)
-        value = {"subject": subject, "version": version, "id": schema_id, "schema": schema.schema_str, "deleted": deleted}
-        if schema.schema_type is not SchemaType.AVRO:
-            value["schemaType"] = schema.schema_type
-        return self.send_kafka_message(key, json_encode(value, compact=True))
+        if schema is not None:
```
juha-aiven, line 233: `if not schema`

karapace/schema_registry_apis.py
```diff
@@ -395,16 +420,39 @@ async def subjects_list(self, content_type):
...
-            self.send_delete_subject_message(subject, latest_schema_id)
+            if permanent:
+                for version, value in list(subject_data["schemas"].items()):
+                    schema_id = value.get("id")
+                    self.log.info("Permanently deleting subject '%s' version %s (schema id=%s)", subject, version, schema_id)
+                    self.send_schema_message(
+                        subject=subject, schema=None, schema_id=schema_id, version=version, deleted=True
+                    )
+            else:
+                self.send_delete_subject_message(subject, latest_schema_id)
```
juha-aiven, line 455: Don't we need to `send_delete_subject_message` if `permanent`?

juha-aiven (commented):
A few minor comments. Looks fine otherwise, as far as I can tell.

karapace/schema_registry_apis.py
juha-aiven: I see, so if called `include_deleted=true` this method would up pushing the deleted schemas to `ksr.subjects`? It was strange behaviour that this method has a mutating side effect. Sure that nothing depends on that? Now this method works differently, i.e. the side effect has been removed.

karapace/schema_registry_apis.py
```diff
@@ -114,29 +128,30 @@ def _create_master_coordinator(self):
         self.mc = MasterCoordinator(config=self.config)
         self.mc.start()
 
-    def _subject_get(self, subject, content_type):
+    def _subject_get(self, subject, content_type, include_deleted=False):
```
hackaugusto, line 131: Please add type annotation. I would suggest that we always add type annotations to functions that we modify. Eventually we will have the whole codebase typed.

karapace/schema_registry_apis.py
```diff
@@ -395,16 +420,39 @@ async def subjects_list(self, content_type):
         subjects_list = [key for key, val in self.ksr.subjects.items() if self.ksr.get_schemas(key)]
         self.r(subjects_list, content_type, status=HTTPStatus.OK)
 
-    async def subject_delete(self, content_type, *, subject):
-        self._subject_get(subject, content_type)
-        version_list = list(self.ksr.get_schemas(subject))
+    async def subject_delete(self, content_type, *, subject, request: HTTPRequest):
```
hackaugusto, line 423: There are lots of copies going and iteration over the entries here.

- `_subject_get` calls `get_schemas` which iterates on all entries to filter the deleted schemas, and returns a new dict
- `_subject_get` makes a copy of the result above (@juha-aiven [suggestion](https://github.com/aiven/karapace/pull/210/files#r627156298) would eliminate this second copy, since it would always be a new dict)
- the check for soft deletes below iterate over all the schemas again
- then `get_schemas` is called a second time, which may make another copy, and then a new list of the keys is copied out
- and then there is another copy in the loop which actually send the hard delete message

It would be nice to improve this.

karapace/schema_registry_apis.py
```diff
@@ -395,16 +420,39 @@ async def subjects_list(self, content_type):
...
+                status=HTTPStatus.NOT_FOUND,
+            )
+
+        version_list = list(self.ksr.get_schemas(subject, include_deleted=permanent))
         if version_list:
             latest_schema_id = version_list[-1]
         else:
             latest_schema_id = 0
         are_we_master, master_url = await self.get_master()
         if are_we_master:
```
hackaugusto, line 446: The flag for permanent delete is not forwarded if we are *not* the master. (couldn't comment on the line because GH UI doesn't allow it)

karapace/schema_registry_apis.py
hackaugusto: This method needs synchronization with the `subject_post`, if the whole subject is being deleted, a new schema must not be registered.

karapace/schema_registry_apis.py
hackaugusto: please move this to a validation method, lets try to end up with something like this:

https://github.com/aiven/karapace/blob/ac9a811c9af08afc86237d18d894a071edd25584/karapace/schema_registry_apis.py#L618-L631

Otherwise it gets convoluted and hard to read.

karapace/schema_registry_apis.py
```diff
@@ -458,25 +506,55 @@ async def subject_version_get(self, content_type, *, subject, version, return_di
...
+        permanent = request.query.get("permanent", "false").lower() == "true"
+
+        subject_data = self._subject_get(subject, content_type, include_deleted=True)
 
         subject_schema_data = subject_data["schemas"].get(version, None)
+        self.log.info("subject scjema data %s", subject_schema_data)
+        self.log.info("subject  data %s", subject_data)
         if not subject_schema_data:
-            self.r(
+            return self.r(
```
hackaugusto, line 519: This is not needed, `self.r` raises an exception

karapace/schema_registry_apis.py
```diff
@@ -458,25 +506,55 @@ async def subject_version_get(self, content_type, *, subject, version, return_di
...
+                    ),
                 },
                 content_type=content_type,
                 status=HTTPStatus.NOT_FOUND,
             )
+
         schema_id = subject_schema_data["id"]
         schema = subject_schema_data["schema"]
         are_we_master, master_url = await self.get_master()
         if are_we_master:
```
hackaugusto, line 554: here too the `permanent` is not being forwarded to the master.

hackaugusto (changes requested):

karapace/schema_registry_apis.py
tvainika: If called with `include_deleted=` any value, it no longer mutates. Now it this method is free from side effects. Earlier any get method caused soft deleted schema versions in `ksr.subjects` to be removed, ie. `ksr.subjects` was modified both outside `KafkaSchemaReader` (`ksr`) and inside it.

I think the test suite should capture if this removal of side effect had any extra dependency. Of course beside the memory consumption mentioned.

karapace/schema_registry_apis.py
tvainika: No need to send new delete subject message, soft delete message is enough.

karapace/schema_reader.py
tvainika: Do you think it would be beneficial as current implementation avoids copying of that list iff `include_deleted=true`?

karapace/schema_registry_apis.py
tvainika: Actually `if schema:` but fixed

karapace/schema_registry_apis.py
tvainika: Validation method for delete case is quite different from others as it needs to read soft deleted versions and `permanent` flag so I failed to see how mixing this into another method really improves readability of the code...

karapace/schema_registry_apis.py
tvainika: Fixed second reading of `get_schemas` as it was not needed. When working on this I had a bit hard time to understand original code with `_subject_get` and `ksr.get_schemas` both going into ksr's internals.

About locking for me it looks there has not been any locking to synchronize earlier for posting or deleting. Only lock is for id reservation. So if this locking is an issue, should it be addressed separately for good as I fail to see that this PR would introduce any new synchronization issues?

hackaugusto: > About locking for me it looks there has not been any locking to synchronize earlier for posting or deleting. Only lock is for id reservation. So if this locking is an issue, should it be addressed separately for good as I fail to see that this PR would introduce any new synchronization issues?

This PR is not introducing a new issue, but it is touching the code that needs and lacks synchronization. For me the important thing is to fix the issue, I would prefer to do it now instead of delaying it.

tests/integration/test_schema.py
```diff
@@ -1813,3 +1813,154 @@ async def test_invalid_schemas(registry_async_client):
...
+
+    # Hard delete whole schema
+    res = await registry_async_client.delete("subjects/{}?permanent=true".format(subject))
+    assert res.status_code == 200
+    assert res.json() == [1, 2]
+
+    res = await registry_async_client.get("subjects/{}/versions".format(subject))
+    assert res.status_code == 404
+    assert res.json()["error_code"] == 40401
+    assert res.json()["message"] == f"Subject '{subject}' not found."
```
hackaugusto, line 1966: Please add code to create a new schema for the deleted subject. I don't know what should be the behavior then, if the IDs are reset or they continue from the previous value.

hackaugusto: We should also double check if there is such a test for soft deletes

karapace/schema_registry_apis.py
hackaugusto: I'm not proposing mixing code. It is not about code reuse, but about code organization, making sure the methods are short and giving names to code so that it is easier to follow. i.e. creating a method `_validate_existing_schemas_are_soft_deleted` with this if condition. That makes the method shorter, and the function name is good enough documentation for what is the intent of that piece of code.

karapace/schema_reader.py
hackaugusto: please add types

karapace/schema_registry_apis.py
```diff
@@ -395,21 +420,42 @@ async def subjects_list(self, content_type):
         subjects_list = [key for key, val in self.ksr.subjects.items() if self.ksr.get_schemas(key)]
         self.r(subjects_list, content_type, status=HTTPStatus.OK)
 
-    async def subject_delete(self, content_type, *, subject):
-        self._subject_get(subject, content_type)
-        version_list = list(self.ksr.get_schemas(subject))
+    async def subject_delete(self, content_type, *, subject, request: HTTPRequest):
```
hackaugusto, line 423: please add types

karapace/schema_registry_apis.py
```diff
@@ -458,30 +504,58 @@ async def subject_version_get(self, content_type, *, subject, version, return_di
             return ret
         self.r(ret, content_type)
 
-    async def subject_version_delete(self, content_type, *, subject, version):
+    async def subject_version_delete(self, content_type, *, subject, version, request: HTTPRequest):
```
hackaugusto, line 507: please add types

tests/integration/test_schema.py
```diff
@@ -1813,3 +1813,154 @@ async def test_invalid_schemas(registry_async_client):
     )
     assert res.status != 500, "an invalid schema should not cause a server crash"
     assert not is_success(HTTPStatus(res.status)), "an invalid schema must not be a success"
+
+
+async def test_schema_hard_delete_version(registry_async_client):
```
hackaugusto, line 1818: typing

tests/integration/test_schema.py
```diff
@@ -1813,3 +1813,154 @@ async def test_invalid_schemas(registry_async_client):
...
+    assert res.status_code == 200
+
+    # Cannot hard delete twice
+    res = await registry_async_client.delete("subjects/{}/versions/1?permanent=true".format(subject))
+    assert res.status_code == 404
+    assert res.json()["error_code"] == 40402
+    assert res.json()["message"] == "Version 1 not found."
+
+
+async def test_schema_hard_delete_whole_schema(registry_async_client):
```
hackaugusto, line 1897: typing

PR #213, opened 2021-05-06
Schema versions

juha-aiven:
Added support for a missing API call

https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--schemas-ids-int-%20id-versions

Note the test `test_schema_versions_multiple_subjects_with_same_schema` doesn't work (yet) against Schema Registry. There are other tests too that don't work against it. The idea is to fix them all in a separate PR.

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -7,7 +7,11 @@ See LICENSE for details
 from http import HTTPStatus
 from kafka import KafkaProducer
 from karapace.rapu import is_success
-from tests.utils import new_random_name, repeat_until_successful_request
+from karapace.utils import Client
+from tests.utils import (
+    create_field_name_factory, create_subject_name_factory, new_random_name, repeat_until_successful_request
+)
+from typing import List, Tuple
 
 import json as jsonlib
 import os
@@ -20,7 +24,10 @@ baseurl = "http://localhost:8081"
 @pytest.mark.parametrize("trail", ["", "/"])
 @pytest.mark.parametrize("compatibility", ["FORWARD", "BACKWARD", "FULL"])
 async def test_enum_schema_compatibility(registry_async_client, compatibility, trail):
-    subject = new_random_name("subject")
+    subject_name_factory = create_subject_name_factory(f"test_enum_schema_compatibility-{trail}")
+
+    subject = subject_name_factory()
+
     res = await registry_async_client.put(f"config{trail}", json={"compatibility": compatibility})
     assert res.status == 200
     schema = {
@@ -97,7 +104,9 @@ async def test_enum_schema_compatibility(registry_async_client, compatibility, t
 
 @pytest.mark.parametrize("trail", ["", "/"])
 async def test_union_to_union(registry_async_client, trail):
-    subject = new_random_name("subject")
+    subject_name_factory = create_subject_name_factory(f"test_union_to_union-{trail}")
+    subject = subject_name_factory()
+
     res = await registry_async_client.put(f"config/{subject}{trail}", json={"compatibility": "BACKWARD"})
     assert res.status == 200
     init_schema = {"name": "init", "type": "record", "fields": [{"name": "inner", "type": ["string", "int"]}]}
@@ -131,7 +140,7 @@ async def test_union_to_union(registry_async_client, trail):
     )
     assert res.status == 200
     # fw compat check
-    subject = new_random_name("subject")
+    subject = subject_name_factory()
     res = await registry_async_client.put(f"config/{subject}{trail}", json={"compatibility": "FORWARD"})
     assert res.status == 200
     res = await registry_async_client.post(
@@ -149,7 +158,9 @@ async def test_union_to_union(registry_async_client, trail):
 
 @pytest.mark.parametrize("trail", ["", "/"])
 async def test_missing_subject_compatibility(registry_async_client, trail):
-    subject = new_random_name("subject")
+    subject_name_factory = create_subject_name_factory(f"test_missing_subject_compatibility-{trail}")
+    subject = subject_name_factory()
+
     res = await registry_async_client.post(
         f"subjects/{subject}/versions{trail}", json={"schema": jsonlib.dumps({"type": "string"})}
     )
@@ -166,7 +177,9 @@ async def test_missing_subject_compatibility(registry_async_client, trail):
 
 @pytest.mark.parametrize("trail", ["", "/"])
 async def test_record_union_schema_compatibility(registry_async_client, trail):
-    subject = new_random_name("subject")
+    subject_name_factory = create_subject_name_factory(f"test_record_union_schema_compatibility-{trail}")
+    subject = subject_name_factory()
+
     res = await registry_async_client.put(f"config/{subject}{trail}", json={"compatibility": "BACKWARD"})
     assert res.status == 200
     original_schema = {
@@ -245,7 +258,9 @@ async def test_record_union_schema_compatibility(registry_async_client, trail):
 
 @pytest.mark.parametrize("trail", ["", "/"])
 async def test_record_nested_schema_compatibility(registry_async_client, trail):
-    subject = new_random_name("subject")
+    subject_name_factory = create_subject_name_factory(f"test_record_nested_schema_compatibility-{trail}")
+    subject = subject_name_factory()
+
     res = await registry_async_client.put("config", json={"compatibility": "BACKWARD"})
     assert res.status == 200
     schema = {
@@ -289,10 +304,12 @@ async def test_record_nested_schema_compatibility(registry_async_client, trail):
 
 @pytest.mark.parametrize("trail", ["", "/"])
 async def test_compatibility_endpoint(registry_async_client, trail):
+    subject_name_factory = create_subject_name_factory(f"test_compatibility_endpoint-{trail}")
+    subject = subject_name_factory()
+
     res = await registry_async_client.put(f"config{trail}", json={"compatibility": "BACKWARD"})
     assert res.status == 200
 
-    subject = new_random_name("subject")
     schema = {
         "type": "record",
         "name": "Objct",
@@ -384,8 +401,9 @@ async def test_type_compatibility(registry_async_client, trail):
             yield "FULL", target, source, False
             yield "FULL", source, target, False
 
+    subject_name_factory = create_subject_name_factory(f"test_type_compatibility-{trail}")
     for compatibility, source_type, target_type, expected in _test_cases():
-        subject = new_random_name("subject")
+        subject = subject_name_factory()
         res = await registry_async_client.put(f"config/{subject}{trail}", json={"compatibility": compatibility})
         schema = {
             "type": "record",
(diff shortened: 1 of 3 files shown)


karapace/schema_reader.py
```diff
@@ -50,7 +50,7 @@ class SchemaType(str, Enum):
 
 
 class TypedSchema:
-    def __init__(self, schema, schema_type: SchemaType, schema_str: str):
+    def __init__(self, schema: any, schema_type: SchemaType, schema_str: str):
```
hackaugusto, line 53: `schema` should be typed as `Union[avro.schema.Schema, jsonschema.Draft7Validator, dict]`. I'm not sure if the third option is necessary, we would need to investigate why we have this:

https://github.com/aiven/karapace/blob/ff348266f67517c644d8c4314c662599edb444a2/karapace/schema_reader.py#L307-L310

juha-aiven: And even type the `dict` in the `Union` should be typed.

hackaugusto: > And even type the dict in the Union should be typed

Maybe we should just remove it. Here is the PR that introduce the `json.loads`:

https://github.com/aiven/karapace/pull/91

I think we don't need it, just to handle the exception and log the message

juha-aiven: I removed all these typings I added. Out of scope of this PR.

hackaugusto (changes requested):
Please fix the commit history. Also it seems you don't have the `pre-commit` hooks installed, it would have caught the formatting issues at commit time.

tests/integration/test_schema.py
```diff
@@ -1186,7 +1186,7 @@ async def assert_schema_versions(client: Client, url_trail: str, schema_id: int,
     assert res.json() == [3]  # Version number generation should now begin at 3
 
     # Check the return format on a more complex schema for version get
-    subject_2 = new_random_name("subject")
+    s2 = new_random_name("subject")
```
hackaugusto: I would argue that we should prefer longer names instead of shorter lines. Also none of the lines here were long enough for this to be necessary

tests/integration/test_schema.py
```diff
@@ -20,7 +22,10 @@
 @pytest.mark.parametrize("trail", ["", "/"])
 @pytest.mark.parametrize("compatibility", ["FORWARD", "BACKWARD", "FULL"])
 async def test_enum_schema_compatibility(registry_async_client, compatibility, trail):
-    subject = new_random_name("subject")
+    nf = subject_name_factory(f"test_enum_schema_compatibility-{trail}")
```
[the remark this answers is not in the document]
juha-aiven: Will do.

tests/integration/test_schema.py
```diff
@@ -975,25 +1002,96 @@ async def test_transitive_compatibility(registry_async_client):
     assert res_json["error_code"] == 409
 
 
+async def assert_schema_versions(
```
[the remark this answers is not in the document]
juha-aiven: Makes total sense. I'll do this.

tests/integration/test_schema.py
```diff
@@ -975,25 +1002,96 @@ async def test_transitive_compatibility(registry_async_client):
...
+    schema_str1 = '{"type": "string", "just-making-this-unique" : "csab789-%s"}' % trail
+    schema_str2 = '{"type": "string", "just-making-this-unique" : "dsukybu-%s"}' % trail
+
+    async def subject_post(subject: str, schema_str: str):
+        res = await registry_async_client.post(
+            f"subjects/{subject}/versions{trail}",
+            json={"schema": schema_str},
+        )
+        assert res.status == 200
+        assert "id" in res.json()
```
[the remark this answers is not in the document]
juha-aiven: I guess both ways can be justified. If getting a key that doesn't exist, AFAIK a `KeyError` is raised. The explicit `assert` gives a nicer error. I'll remove this.

tests/integration/test_schema.py
```diff
@@ -975,25 +1002,96 @@ async def test_transitive_compatibility(registry_async_client):
...
+
+
 @pytest.mark.parametrize("trail", ["", "/"])
 async def test_schema(registry_async_client, trail):
-    subject = new_random_name("subject")
-    schema_str = '{"type": "string"}'
+    nf = subject_name_factory(f"test_schema-{trail}")
+
+    subject = nf()
+    schema_str = '{"type": "string", "just-making-this-unique" : "xyz-%s"}' % trail
```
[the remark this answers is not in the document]
juha-aiven: `test_schema` is way too large. It's very hard to say what's the point of it. I think refactoring that to smaller, more focused test methods should be done next. Then we can throw stuff away if needed.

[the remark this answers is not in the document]
hackaugusto: that is my point `test_schema` is too large, we shouldn't add new things to it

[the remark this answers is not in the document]
juha-aiven: I can address that in another PR that splits `test_schema` into multiple parts.

tests/integration/test_schema.py
```diff
@@ -990,23 +1102,29 @@ async def test_schema_types(registry_async_client, trail):
...
     assert res.status_code == 200
     assert res.json()["schema"] == schema_str
 
+    await assert_schema_versions(registry_async_client, trail, schema_id, [(subject, 1)])
+
     # repost same schema again to see that a new id is not generated but an old one is given back
     res = await registry_async_client.post(
         f"subjects/{subject}/versions{trail}",
-        json={"schema": '{"type": "string"}'},
+        json={"schema": '{"type": "string", "unique" : "%s"}' % unique_field_factory()},
```
hackaugusto, line 1127: This change is breaking the test, this schemas is not supposed to be different, as per the comment above.

hackaugusto (changes requested):

tests/integration/test_schema.py
```diff
@@ -1008,7 +1008,12 @@ async def assert_schema_versions(client: Client, trail: str, schema_id: int, exp
     """
     res = await client.get(f"/schemas/ids/{schema_id}/versions{trail}")
     assert res.status_code == 200
-    assert [{"subject": e[0], "version": e[1]} for e in expected] == res.json()
+
+    # Schema Registry doesn't return an ordered list, Karapace does.
+    # Need to check equality ignoring ordering.
+    assert len(res.json()) == len(expected)
+    for e in ({"subject": e[0], "version": e[1]} for e in expected):
```
hackaugusto, line 1015: Please change the signature of the function to receive the dictionary values. This will then be:
```python
assert expected == sorted(res.json(), key=lambda result: result["version"])
```

tests/utils.py
```diff
@@ -211,14 +212,21 @@ def new_random_name(prefix):
     return f"{prefix}{suffix}"
 
 
-def create_subject_name_factory(prefix):
-    subject_index = 1
+def create_id_factory(prefix=None):
```
hackaugusto, line 215: Please add the types to the signature.

Also make the prefix required, we use random ids for subjects, schema fields, topics. A meaningful prefix is useful while debugging

hackaugusto: Also, I just noticed that all the subject schemas lost the prefix that said they were a subject, can you please add it back?

tests/integration/test_schema.py
juha-aiven: Yes, a bug in the code. Fixing.

tests/utils.py
juha-aiven: True both. Fixing.

tests/integration/test_schema.py
juha-aiven: The called would then need to create the `dict` always? I think it's easier to write tuples - no need to repeat the keys all the time? What would be the benefit of passing a `dict`?

Also, I think `key` in the `lambda` above should take into account also `subject`. The list in Karapace is ordered like this `key=lambda s: (s["subject"], s["version"]`.

tests/integration/test_schema.py
```diff
@@ -975,25 +1002,96 @@ async def test_transitive_compatibility(registry_async_client):
...
+    Tests case where there are multiple subjects with the same schema.
+    The schema/versions endpoint returns all these subjects.
+    """
+    nf = subject_name_factory(f"test_schema-{trail}")
+
+    schema_str1 = '{"type": "string", "just-making-this-unique" : "csab789-%s"}' % trail
+    schema_str2 = '{"type": "string", "just-making-this-unique" : "dsukybu-%s"}' % trail
+
+    async def subject_post(subject: str, schema_str: str):
+        res = await registry_async_client.post(
```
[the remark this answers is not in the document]
juha-aiven: This is fixed now. The reason was actually the fact that the subject name contained a `/` causing the request URL in the case `trail == "/"` to be "corrupt". SR didn't recognize the URL. That's why the random subject name thing now changes `/` in the prefix to `_`.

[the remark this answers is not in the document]
juha-aiven: And the attempt to decode was there because the _response_ in the error case wasn't `application/json` and aiohttp didn't like that.

tests/integration/test_schema.py
hackaugusto: > The called would then need to create the dict always?

Yes. I don't understand if the problem you're pointing out is the instantiation itself, because this loop is instantiating the dicts anyways.

> I think it's easier to write tuples - no need to repeat the keys all the time?

Both are extremely easy to read and write:
```python
[
  ("subject1", 1),
  ("subject1", 2),
  ("subject1", 3),
]
```
```python
[
  {"subject": "subject1", "version": 1},
  {"subject": "subject1", "version": 2},
  {"subject": "subject1", "version": 3},
]
```

> What would be the benefit of passing a dict?

It is the actual expected result, not a list of tuples.

> Also, I think key in the lambda above should take into account also subject. The list in Karapace is ordered like this key=lambda s: (s["subject"], s["version"].

Sure, that was just an example to remove the `o^2` complexity and streamline the code

tests/utils.py
```diff
@@ -205,11 +207,35 @@ async def new_consumer(c, group, fmt="avro", trail=""):
...
 
+def create_subject_name_factory(prefix: str) -> str:
+    return create_id_factory(f"subject-{prefix}")
+
+
+def create_field_name_factory(prefix: str) -> str:
+    return create_id_factory(f"field-{prefix}")
+
+
+def create_id_factory(prefix: str) -> str:
```
hackaugusto, line 223: This type is not correct, this function creates a factory, the factory is a function:
```suggestion
def create_id_factory(prefix: str) -> Callable[[], str]:
```

tests/utils.py
```diff
@@ -205,11 +207,35 @@ async def new_consumer(c, group, fmt="avro", trail=""):
...
+    """
+    Creates unique ids prefixed with prefix..
+    The resulting ids are safe to embed in URLs.
+    """
+    index = 1
+
+    def create_name():
```
hackaugusto, line 230: This is missing the return type.

tests/integration/test_schema.py
juha-aiven: I'm leaving like it is, if it causes problems, can be changed later. The effective functionality is the same in both approaches.

tests/utils.py
juha-aiven: Fixing!

tests/utils.py
juha-aiven: Fixing this one.

hackaugusto (approved):
LGTM

Issue #108, opened 2020-10-09
add a `/schemas/types` endpoint


gabriel-tincu:
It should list the distinct types of schemas currently stored in the registry

juha-aiven:
Spec in https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--schemas-types-.

hackaugusto:
Fixed by https://github.com/aiven/karapace/pull/215

Issue #67, opened 2020-06-08
Add support for ProtoBuf


nick-zh:
As CSR supports ProtoBuf, i think it would be great if karapace would do as well.

gabriel-tincu:
We're currently looking at the effort for implementing this, though it's definitely on the table. Will update this thread with new info once we have it

amrutha-shanbhag:
We at Instaclustr have commenced work to support Protobuf. Updates can be found in the fork https://github.com/instaclustr/karapace

hackaugusto:
Hello @amrutha-shanbhag , thank you for tackling this :smile: . Currently there are no big changes planned, so you should be good.

Also, here are some pointers to get you going:

- [the API needs to accept the format](https://github.com/aiven/karapace/blob/0d25b816c5fd853a6c858105f95f98e6907f0ccd/karapace/schema_registry_apis.py#L546)
- [The code needs support for parsing the data](https://github.com/aiven/karapace/blob/0d25b816c5fd853a6c858105f95f98e6907f0ccd/karapace/schema_reader.py#L75-L80). This is used for both the REST api and to read data from the topic (the [schema_str](https://github.com/aiven/karapace/blob/0d25b816c5fd853a6c858105f95f98e6907f0ccd/karapace/schema_reader.py#L56) is saved in the `_schemas` topics, [ref](https://github.com/aiven/karapace/blob/0d25b816c5fd853a6c858105f95f98e6907f0ccd/karapace/schema_registry_apis.py#L218)) 
- [and finally the compatibility checks themselves have to be enabled](https://github.com/aiven/karapace/blob/0d25b816c5fd853a6c858105f95f98e6907f0ccd/karapace/compatibility/__init__.py#L65)
   - I'd recommend adding a new module named `protobuf` [here](https://github.com/aiven/karapace/tree/0d25b816c5fd853a6c858105f95f98e6907f0ccd/karapace/compatibility)
- The code to publish the data and read it is in place, it should work out of the box.

Testing can be done with unit and integration tests. The former is good for debugging, the latter to check for compatibility w.r.t. schema registry [ref](https://github.com/aiven/karapace/pull/180)

amrutha-shanbhag:
Hi @hackaugusto , whats your preferred way to review our work? 

[...]

PR #215, opened 2021-05-10
Added support for /schemas/types endpoint

juha-aiven:
/schemas/types endpoint

Fixes issue #108

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -70,6 +70,7 @@ class KarapaceSchemaRegistry(KarapaceBase):
         self.route("/config", callback=self.config_get, method="GET", schema_request=True)
         self.route("/config", callback=self.config_set, method="PUT", schema_request=True)
         self.route("/schemas/ids/<schema_id:path>", callback=self.schemas_get, method="GET", schema_request=True)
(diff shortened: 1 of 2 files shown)


tvainika (approved):

karapace/schema_registry_apis.py
```diff
@@ -309,6 +310,9 @@ async def schemas_get(self, content_type, *, schema_id):
             response_body["schemaType"] = schema.schema_type
         self.r(response_body, content_type)
 
+    async def schemas_types(self, content_type):
+        self.r(["JSON", "AVRO"], content_type)
```
hackaugusto, line 314: It is very easy to get these out-of-sync. It would make more sense to have this as an attribute of:

https://github.com/aiven/karapace/blob/f9f0ef9987e4af4f519c11b6481bcc606cd4f511/karapace/schema_reader.py#L52

juha-aiven: The issue currently is that Protobuf is unsupported: https://github.com/aiven/karapace/issues/67, thus `TypedSchema` entries don't match the schema types supported. Once Protobuf is supported, we can make this endpoint return the `TypedSchema` entries. The other approach is to make this endpoint lie about the supported endpoints. I wouldn't do that. I added a comment a reminder in #67 to add PROTOBUF to this endpoint results once Protobuf's supported.

PR #219, opened 2021-05-11
Added parameters for SSL configurations:

germanilia:
Those parameters will allow easily to control the security protocol and then add the certificate files.
This change is needed when running karapace in docker mode on top of k8s cluster.
The next step will be to create an Helm chart for the deployment.

### DIFF
--- a/container/docker-compose.yml
+++ b/container/docker-compose.yml
@@ -79,7 +79,7 @@ services:
       KARAPACE_SSL_CA_FILE_PATH: null
       KARAPACE_SSL_CERT_FILE_PATH: null
       KARAPACE_SSL_KEY_FILE_PATH: null
-      KARAPACE_SSL_CHECK_HOSTANME: True
+      KARAPACE_SSL_CHECK_HOSTNAME: True
 
   karapace-rest:
     image: aivenoy/karapace:latest
--- a/container/docker-compose.yml
+++ b/container/docker-compose.yml
@@ -75,6 +75,11 @@ services:
       KARAPACE_REGISTRY_MASTER_ELIGIBITY: "true"
       KARAPACE_REGISTRY_TOPIC_NAME: _schemas
       KARAPACE_REGISTRY_LOG_LEVEL: WARNING
+      KARAPACE_SECURITY_PROTOCOL: PLAINTEXT
+      KARAPACE_SSL_CA_FILE_PATH: null
+      KARAPACE_SSL_CERT_FILE_PATH: null
+      KARAPACE_SSL_KEY_FILE_PATH: null
+      KARAPACE_SSL_CHECK_HOSTANME: True
 
   karapace-rest:
     image: aivenoy/karapace:latest
@@ -98,3 +103,7 @@ services:
       # new entries may need to be added to containers/start.sh
       KARAPACE_REST_ADMIN_METADATA_MAX_AGE: 0
       KARAPACE_REST_LOG_LEVEL: WARNING
+      KARAPACE_SECURITY_PROTOCOL: PLAINTEXT
+      KARAPACE_SSL_CA_FILE_PATH: null
+      KARAPACE_SSL_CERT_FILE_PATH: null
+      KARAPACE_SSL_KEY_FILE_PATH: null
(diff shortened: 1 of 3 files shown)


container/start.sh
```diff
@@ -40,10 +40,11 @@ start_karapace_registry(){
...
     "replication_factor": 1,
-    "security_protocol": "PLAINTEXT",
-    "ssl_cafile": null,
-    "ssl_certfile": null,
-    "ssl_keyfile": null
+    "security_protocol": "${KARAPACE_SECURITY_PROTOCOL:-PLAINTEXT}",
+    "ssl_cafile": "${KARAPACE_SSL_CA_FILE_PATH:-null}",
+    "ssl_certfile": "${KARAPACE_SSL_CERT_FILE_PATH:-null}",,
+    "ssl_keyfile": "${KARAPACE_SSL_KEY_FILE_PATH:-null}",
+    "ssl_check_hostname": "${KARAPACE_SSL_CHECK_HOSTANME:-True}"
```
tvainika, line 47: This produces invalid JSON, because it has quotes around `null` and booleans should be lowercased ie. `true` instead of `True` inside JSON file. Also there is double `,,` after certfile and HOSTANME spelling mistake
```
    "ssl_keyfile": "null",
    "ssl_check_hostname": "True"
```

tvainika (changes requested):
Thank you for the contributions. It looks like JSON generation inside shell script is broken in these changes, ref. inline comments. Karapace's config parser also reads environment variables, so the overall setup could be simplified by utilizing those to achieve yout objectives of this patch?

container/start.sh
germanilia: Resolved the issues, 10x

container/docker-compose.yml
```diff
@@ -75,6 +75,11 @@ services:
       KARAPACE_REGISTRY_MASTER_ELIGIBITY: "true"
       KARAPACE_REGISTRY_TOPIC_NAME: _schemas
       KARAPACE_REGISTRY_LOG_LEVEL: WARNING
+      KARAPACE_SECURITY_PROTOCOL: PLAINTEXT
+      KARAPACE_SSL_CA_FILE_PATH: null
+      KARAPACE_SSL_CERT_FILE_PATH: null
+      KARAPACE_SSL_KEY_FILE_PATH: null
+      KARAPACE_SSL_CHECK_HOSTANME: True
```
tvainika, line 82:
```suggestion
      KARAPACE_SSL_CHECK_HOSTNAME: True
```

container/start.sh
```diff
@@ -40,10 +40,11 @@ start_karapace_registry(){
...
     "replication_factor": 1,
-    "security_protocol": "PLAINTEXT",
-    "ssl_cafile": null,
-    "ssl_certfile": null,
-    "ssl_keyfile": null
+    "security_protocol": "${KARAPACE_SECURITY_PROTOCOL:-PLAINTEXT}",
+    "ssl_cafile": "${KARAPACE_SSL_CA_FILE_PATH}",
+    "ssl_certfile": "${KARAPACE_SSL_CERT_FILE_PATH}",
+    "ssl_keyfile": "${KARAPACE_SSL_KEY_FILE_PATH}",
+    "ssl_check_hostname": ${KARAPACE_SSL_CHECK_HOSTANME:-true}
```
tvainika, line 47:
```suggestion
    "ssl_check_hostname": ${KARAPACE_SSL_CHECK_HOSTNAME:-true}
```

germanilia (commented):
.

germanilia:
Thanks  for the comment, I think the changes should be integrated to start.sh also for the certificate paths

tvainika (changes requested):
I'm failing to get this running. How have you tested these changes ie. what kind of setup do you have on your mind?
```
$ docker-compose up
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.karapace-registry.environment.KARAPACE_SSL_CHECK_HOSTNAME contains true, which is an invalid type, it should be a string, number, or a null
```

tvainika:
I'm closing this now due #264 fixing the issue here. Please re-open or open new if there is something still to fix.

PR #221, opened 2021-05-12
Split test schema

juha-aiven:
Splitting `test_schema` (https://github.com/aiven/karapace/blob/master/tests/integration/test_schema.py#L1099) function in the integration tests to multiple test methods. Currently the function does so much that it's hard to understand what it actually does. The reason for splitting is to make the test more readable and maintainable. #222 is based on this refactored test code.

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -1201,12 +1201,11 @@ async def test_schema_subject_post_invalid(registry_async_client: Client) -> Non
     )
     assert res.status == 200
 
-    # Invalid schema for existing subject should return 500
     res = await registry_async_client.post(
         f"subjects/{subject_1}",
         json={"schema": '{"type": "invalid_type"}'},
     )
-    assert res.status == 500
+    assert res.status == 500, "Invalid schema for existing subject should return 500"
     assert res.json()["message"] == f"Error while looking up schema under subject {subject_1}"
 
     # Subject is not found
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -290,7 +290,7 @@ async def test_record_nested_schema_compatibility(registry_async_client: Client,
     # change string to integer in the nested record, should fail
     schema["fields"][1]["type"]["fields"][0]["type"] = "int"
     res = await registry_async_client.post(
-        "subjects/{}/versions".format(subject),
+        f"subjects/{subject}/versions",
         json={"schema": jsonlib.dumps(schema)},
     )
     assert res.status == 409
@@ -319,8 +319,9 @@ async def test_compatibility_endpoint(registry_async_client: Client, trail: str)
         json={"schema": jsonlib.dumps(schema)},
     )
     assert res.status == 200
+    schema_id = res.json()['id']
 
-    res = await registry_async_client.get("schemas/ids/{}{}".format(res.json()["id"], trail))
+    res = await registry_async_client.get(f"schemas/ids/{schema_id}{trail}")
     schema_gotten_back = jsonlib.loads(res.json()["schema"])
     assert schema_gotten_back == schema
 
@@ -1166,20 +1167,20 @@ async def test_schema_subject_invalid_id(registry_async_client: Client, trail: s
     unique_field_factory = create_field_name_factory(trail)
 
     res = await registry_async_client.post(
-        "subjects/{}/versions".format(subject),
+        f"subjects/{subject}/versions",
         json={"schema": "{\"type\": \"string\", \"foo\": \"string\", \"%s\": \"string\"}" % unique_field_factory()}
     )
     assert res.status_code == 200
 
     # Find an invalid version 0
-    res = await registry_async_client.get("subjects/{}/versions/0".format(subject))
+    res = await registry_async_client.get(f"subjects/{subject}/versions/0")
     assert res.status_code == 422
     assert res.json()["error_code"] == 42202
     assert res.json()["message"] == \
         'The specified version is not a valid version id. Allowed values are between [1, 2^31-1] and the string "latest"'
 
     # Find an invalid version (too large)
-    res = await registry_async_client.get("subjects/{}/versions/15".format(subject))
+    res = await registry_async_client.get(f"subjects/{subject}/versions/15")
     assert res.status_code == 404
     assert res.json()["error_code"] == 40402
     assert res.json()["message"] == "Version 15 not found."
@@ -1196,7 +1197,7 @@ async def test_schema_subject_post_invalid(registry_async_client: Client) -> Non
     # Create the subject
     subject_1 = subject_name_factory()
     res = await registry_async_client.post(
-        "subjects/{}/versions".format(subject_1),
+        f"subjects/{subject_1}/versions",
         json={"schema": schema_str},
     )
     assert res.status == 200
@@ -1251,7 +1252,7 @@ async def test_schema_lifecycle(registry_async_client: Client, trail: str) -> No
 
     unique_1 = unique_field_factory()
     res = await registry_async_client.post(
-        "subjects/{}/versions".format(subject),
+        f"subjects/{subject}/versions",
         json={"schema": "{\"type\": \"string\", \"foo\": \"string\", \"%s\": \"string\"}" % unique_1}
     )
     assert res.status_code == 200
@@ -1259,7 +1260,7 @@ async def test_schema_lifecycle(registry_async_client: Client, trail: str) -> No
 
     unique_2 = unique_field_factory()
     res = await registry_async_client.post(
-        "subjects/{}/versions".format(subject),
+        f"subjects/{subject}/versions",
         json={"schema": "{\"type\": \"string\", \"foo\": \"string\", \"%s\": \"string\"}" % unique_2}
     )
     schema_id_2 = res.json()["id"]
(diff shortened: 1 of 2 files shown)


tests/integration/test_schema.py
```diff
@@ -1096,30 +1094,42 @@ async def test_schema_types(registry_async_client, trail):
...
     assert res.status_code == 200
     assert res.json()["schema"] == schema_str
 
-    # repost same schema again to see that a new id is not generated but an old one is given back
+    await assert_schema_versions(registry_async_client, trail, schema_id, [(subject, 1)])
+
     res = await registry_async_client.post(
         f"subjects/{subject}/versions{trail}",
-        json={"schema": '{"type": "string"}'},
+        json={"schema": '{"type": "string", "unique" : "%s"}' % unique},
```
hackaugusto, line 1122:
```suggestion
        json={"schema": schema_str},
```

tests/integration/test_schema.py
```diff
@@ -1096,30 +1094,42 @@ async def test_schema_types(registry_async_client, trail):
...
     assert res.status == 200
     assert "id" in res.json()
     schema_id = res.json()["id"]
+
     res = await registry_async_client.get(f"schemas/ids/{schema_id}{trail}")
     assert res.status_code == 200
     assert res.json()["schema"] == schema_str
 
-    # repost same schema again to see that a new id is not generated but an old one is given back
+    await assert_schema_versions(registry_async_client, trail, schema_id, [(subject, 1)])
```
hackaugusto, line 1118: This is unrelated to this test.

tests/integration/test_schema.py
```diff
@@ -1096,30 +1094,42 @@ async def test_schema_types(registry_async_client, trail):
...
-    schema_str = '{"type": "string"}'
+async def test_schema_repost(registry_async_client, trail):
+    """"
+    Repost same schema again to see that a new id is not generated but an old one is given back
+    """
+    subject = create_subject_name_factory(f"test_schema_repost-{trail}")()
+    unique_field_factory = create_field_name_factory(trail)
+
+    unique = unique_field_factory()
+    schema_str = '{"type": "string", "unique" : "%s"}' % unique
```
hackaugusto, line 1105:
```suggestion
    schema_str = json.dumps({"type": "string", "unique" : unique})
```
Using a serializer is less error prone than writing JSON by hand. The code can also benefit of the linting, and formatting by the static tools, and gets easier to read because one doesn't have to escape double quotes.

This also applies to the other tests.

tests/integration/test_schema.py
```diff
@@ -1096,30 +1094,42 @@ async def test_schema_types(registry_async_client, trail):
 
 
 @pytest.mark.parametrize("trail", ["", "/"])
-async def test_schema(registry_async_client, trail):
-    subject = new_random_name("subject")
-    schema_str = '{"type": "string"}'
+async def test_schema_repost(registry_async_client, trail):
```
hackaugusto, line 1097: please always add typing when introducing new code (this applies to the other functions too)

tests/integration/test_schema.py
```diff
@@ -1128,41 +1138,39 @@ async def test_schema(registry_async_client, trail):
...
+async def test_schema_subject_invalid_id(registry_async_client, trail):
+    """
+    Creates a subject with a schema and trying to find the invalid versions for the subject.
+    """
+    subject = create_subject_name_factory(f"test_schema_subject_invalid_id-{trail}")()
+    unique_field_factory = create_field_name_factory(trail)
 
-    res = await registry_async_client.get("subjects/{}/versions/1".format(subject))
+    res = await registry_async_client.post(
+        "subjects/{}/versions".format(subject),
```
hackaugusto, line 1170:
```suggestion
        f"subjects/{subject}/versions"
```
nitpick: smaller expressions are usually easier to read with f-strings, it would also make the coder consistent to use mostly one style. I would use this on most of the places instead of the operator `%`

tests/integration/test_schema.py
```diff
@@ -1177,11 +1185,135 @@ async def test_schema(registry_async_client, trail):
...
     assert res.json()["message"] == "Version 15 not found."
 
+
+async def test_schema_subject_post_invalid(registry_async_client):
+    """
+    Tests posting to /subjects/{subject} with different invalid values.
+    """
+    subject_name_factory = create_subject_name_factory("test_schema_subject_post_invalid")
+
+    schema_str = '{"type": "string"}'
```
hackaugusto, line 1195: what about the random field?

tests/integration/test_schema.py
```diff
@@ -1177,11 +1185,135 @@ async def test_schema(registry_async_client, trail):
...
+
+    # Create the subject
+    subject_1 = subject_name_factory()
+    res = await registry_async_client.post(
+        "subjects/{}/versions".format(subject_1),
+        json={"schema": schema_str},
+    )
+    assert res.status == 200
+
+    # Invalid schema for existing subject should return 500
```
hackaugusto, line 1205: please use this as the `assert` message below. it makes it easier to read test failures

tests/integration/test_schema.py
```diff
@@ -1237,7 +1379,11 @@ async def test_schema(registry_async_client, trail):
                 "name": "first_name",
                 "type": "string",
             },
-        ]
+            {
+                "name": f"{unique}",
```
hackaugusto, line 1383:
```suggestion
                "name": unique,
```
`unique` should already be a string. This also applies to the usages below

tests/integration/test_schema.py
```diff
@@ -1219,16 +1351,26 @@ async def test_schema(registry_async_client, trail):
...
-    subject = new_random_name("subject")
+
+@pytest.mark.parametrize("trail", ["", "/"])
+async def test_schema_version_numbering(registry_async_client, trail):
+    """
+    Test updating the schema of a subject increases its version number.
+    Deletes the subjects and asserts that when recreated, has a greater version number.
+    """
+    subject = create_subject_name_factory(f"test_schema_version_numbering-{trail}")()
+    unique_field_factory = create_field_name_factory(trail)
```
hackaugusto, line 1368: `trail` is not really useful here.

tests/integration/test_schema.py
```diff
@@ -1250,48 +1396,77 @@ async def test_schema(registry_async_client, trail):
                 "name": "first_name",
                 "type": "string",
             },
+            {
+                "name": f"{unique}",
```
hackaugusto, line 1400: This seems odd, the unique marker shouldn't be reused, or the old schema should be used here.

tests/integration/test_schema.py
```diff
@@ -1250,48 +1396,77 @@ async def test_schema(registry_async_client, trail):
...
+                "type": "string",
+            },
             {
                 "name": "last_name",
                 "type": "string",
             },
-        ]
+        ],
+        "unique": f"{unique_field_factory()}"
```
hackaugusto, line 1408:
```suggestion
        "unique": unique_field_factory()
```
The result of `unique_field_factory()` should already be a string. This also applies to the usages below

hackaugusto (changes requested):
I did a first pass, need to do a second one to double check the tests are equivalent. I will do it later because I ran out of time today.

tests/integration/test_schema.py
juha-aiven: Removing.

tests/integration/test_schema.py
juha-aiven: I'll add annotations to all functions in `test_schema.py`. It's nicer to have them in all, not just in the new ones.

tests/integration/test_schema.py
juha-aiven: Fixing.

tests/integration/test_schema.py
juha-aiven: This `"subjects/{}/versions".format(subject)` pattern is there already in master everywhere, it's not that this PR introduces them (at least too many). I'd use f-strings too, but is it worth refactoring (now)? That can be done later too.

Yes, it'd be best to use one string formatting mechanism. Unfortunately f-strings have an issue with curly braces. In order to have a literal curly brace, you need to have two of those. This makes strs containing JSON awkward to write using f-strings,e.g.

```
'{"type": "string", "unique": "%s"}' % unique_field_factory()
```

This would need to be something like this using f-strings:

```
f'{{"type": "string", "unique": "{unique_field_factory()}"}}'
```

So much stuff in the end of the line..... And it gets even worse if use double-quotes for the string.

```
f'{{"type": "string", "unique": "{unique_field_factory()}"}}'
```

The `%` operator is `format` method looks cleaner here.

juha-aiven: So, I'm not changing these right now.

tests/integration/test_schema.py
juha-aiven: Fixing

tests/integration/test_schema.py
juha-aiven: Yes.... In #222 I've been changing these to use `jsonlib.dumps`. Especially when checking against against Schema Registry the format of the incoming JSON string is something that we don't know. Also, this PR doesn't introduce this practise of checking equality of two JSONs using strings - it was already there.

The scope of this PR is to split the code, not try to make everything perfect. There's still a lot of "non-optimal" stuff in the code.

I wouldn't do these fixes here. They'll be fixed to large extent in #222.

tests/integration/test_schema.py
juha-aiven: Fixing!

tests/integration/test_schema.py
juha-aiven: Yes, fixing.

tests/integration/test_schema.py
juha-aiven: I reworked on the test, please have a look.

Maybe all the places where uniqueness is needed, the code should use the `name` of the AVRO schema to make the schema unique?

tests/integration/test_schema.py
juha-aiven: The idea is to make the field names unique for the two parallel tests.

hackaugusto (approved):
thanks for the change.

PR #222, opened 2021-05-17
Improving Schema Registry compatibility

juha-aiven:
This PR is based on #221 that should be merged in first.

This goal of this PR is to make the tests in `schema_test.py` green against Schema Registry 6.1.1. The idea is to change Karapace tests to work with Schema Registry and change Karapace to work against the changed tests.

Currently all tests are green against Karapace and two tests are not green against Schema Registry:

* `test_schema_hard_delete_and_recreate` Karapace reuses the schema id after a permanent delete. SR assigns a new one. Fixing this was left out from this PR to limit the scope.
* `test_schema_types`, since Schema Registry supports Protobuf and Karapace currently doesn't.

Each commit fixes (max) one incompatibility and contains the changes in the test (if any) and changes in Karapace.

Most changes are minor and only about error codes and error messages.

There's a functional change in Avro enum compatibility check. Before this change adding or removing an entry in an Avro enum was never considered incompatibility. In Schema Registry, e.g. remoing an enum entry is not a backwards compatible change

https://github.com/aiven/karapace/compare/sr-compat?expand=1#diff-718d8187ad2280c5fe8cef5daea4210e4257c72ca037719fc4ba44f7884b4668R49


juha-aiven:
We'll do #225 first and then open up a new PR for changes in Karapace for SR compatibility,

PR #223, opened 2021-05-18
Added information how to run integration tests against Confluence stack

juha-aiven:
Instructions in README.rst
Docker Compose file to start the Confluence stack

(all contributors please complete this section, including maintainers)

### DIFF
--- a/tests/integration/confluent-docker-compose.yml
+++ b/tests/integration/confluent-docker-compose.yml
@@ -39,7 +39,7 @@ services:
       SCHEMA_REGISTRY_LISTENERS: http://0.0.0.0:8081
 
   rest:
-    image: confluentinc/cp-kafka-rest:latest
+    image: confluentinc/cp-kafka-rest:6.1.1
     depends_on:
       - kafka
     ports:
--- a/tests/integration/confluent-docker-compose.yml
+++ b/tests/integration/confluent-docker-compose.yml
@@ -1,0 +1,51 @@
+---
+version: '2'
+services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper:latest
+    hostname: zookeeper
+    ports:
+      - "2181:2181"
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
+
+  kafka:
+    image: confluentinc/cp-kafka:latest
+    hostname: kafka
+    depends_on:
+      - zookeeper
+    ports:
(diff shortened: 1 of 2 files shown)


tests/integration/confluent-docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+      - "9101:9101"
+    environment:
+      KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:29092,PLAINTEXT_HOST://localhost:9092
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      KAFKA_CONFLUENT_SCHEMA_REGISTRY_URL: http://schema-registry:8081
+
+  schema-registry:
+    image: confluentinc/cp-schema-registry:6.1.1
```
tvainika, line 29:
```suggestion
    image: confluentinc/cp-schema-registry:latest
```
Maybe this one should be also `latest` tag?

juha-aiven: Good catch. Maybe we should fix the version we test against and aim for compatibility against that specific version?

Here are instructions from Confluent https://docs.confluent.io/platform/current/quickstart/ce-docker-quickstart.html#ce-docker-quickstart, and here https://github.com/confluentinc/cp-all-in-one/blob/6.1.1-post/cp-all-in-one/docker-compose.yml the compose file referred to in the instructions. In that file the versions are

* `confluentinc/cp-schema-registry:6.1.1`
* `confluentinc/cp-kafka-rest:6.1.1`

So, I think 6.1.1 for both would make sense.

tvainika: Either way, having `latest` or `6.1.1` everywhere is fine as long as those are in sync. I'm a slighly favour of `latest` actually as this is not automatically ran, so we might not want to spend effort to keep these in sync with CP's versioning until there is some automated test run, which we guarantee that tests will pass word-by-word?

juha-aiven: Changed to 6.1.1 for both.

Also added mentioning that Karapace aims to be compatible with 6.1.1 and listing of known incompatibilities.

tvainika (approved):

PR #225, opened 2021-05-24
Schema Registry compatibility - round one

juha-aiven:
This PR changes tests in `test_schema.py` in order for those to work against Schema Registry. The reason why the tests didn't pass against SR are multiple, but mostly related to the tests colliding with other tests e.g. by sharing a schema name. The details are in the individual commits that fix one test at time.

Note the changes are 100% on the tests, so no change needed in Karapace itself to fix these issues.

After this PR, what is expected is that the tests that this PR touched work against SR (and naturally against Karapace). There are tests that still fail against SR. There's going to be another PR later to address those. An exception is `test_schema_versions_deleting` that doesn't work against SR, a minor change in Karapace is needed to make it compatible.


tvainika (approved):

hackaugusto:
Something is not right, if the tests were passing with Karapace when the names are being reused, but they are failing with SR, it means that Karapace is being more permissive than SR. Why did the test with SR fail?

tvainika:
> Something is not right, if the tests were passing with Karapace when the names are being reused, but they are failing with SR, it means that Karapace is being more permissive than SR. Why did the test with SR fail?

As far as I see, the tests here were making assertions about implementation specific details, not API compatibility. Thus now the tests are more aligned to test API, and its compatbility instead of asserting specific implementation details.

hackaugusto:
> As far as I see, the tests here were making assertions about implementation specific details, not API compatibility. Thus now the tests are more aligned to test API, and its compatbility instead of asserting specific implementation details.

Sounds like I misunderstood something:

> This PR changes tests in test_schema.py in order for those to work against Schema Registry. The reason why the tests didn't pass against SR are multiple, but mostly related to the tests colliding with other tests e.g. by sharing a schema name

I understood that as the tests were not working when running against SR, but worked when running against Karapace. That would mean that Karapace is accepting inputs that SR is not, which is an incompatibility.

If that is right, and `no change needed in Karapace itself to fix these issues.`, that means we did not fix the compatibility issues.

juha-aiven:
You can have a look at each commit on how a particular test was changed. The commits and the commit messages should explain what was done and why.

Some examples.

* There are tests that were failing against SR because they were testing JSONs as strings, e.g. https://github.com/aiven/karapace/pull/225/commits/309360abe41be7212401ec8af0b008b8007190c2 instead of using "JSON-native" equality check. So the test was relying on a particular way of serializing JSON as string, Bad. Of course it's just natural that another implementation of the API can serialize JSONs diferently (but still in a valid way).
* Many of the tests would have worked against KR and Schema Registry given that a) the tests are not run in parallel and b) Karapace and Schema Registry are restarted with empty state between each test, an example being https://github.com/aiven/karapace/pull/225/commits/a0d76535aa0a99fbba53e402cb266d324cf3d92c. Before making e.g. the schema names unique the tests were colliding with each other (this example actually also fixes the JSON check). Another such example https://github.com/aiven/karapace/pull/225/commits/979d238dc697d8b7ccafa3282f41adc1a3cb58c3.

hackaugusto:
> There are tests that were failing against SR because they were testing JSONs as strings, e.g. 309360a instead of using "JSON-native" equality check. So the test was relying on a particular way of serializing JSON as string, Bad. Of course it's just natural that another implementation of the API can serialize JSONs diferently (but still in a valid way).

Okay, this makes sense, the test was broken.

> Many of the tests would have worked against KR and Schema Registry given that a) the tests are not run in parallel and b) Karapace and Schema Registry are restarted with empty state between each test, an example being a0d7653. Before making e.g. the schema names unique the tests were colliding with each other (this example actually also fixes the JSON check). Another such example 979d238.

I did this:

```
git fetch origin
git checkout '979d238^'
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml down
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d
pytest tests/integration/test_schema.py::test_compatibility_endpoint --registry-url http://localhost:8081 --rest-url http://localhost:8082/ --kafka-bootstrap-servers localhost:9092
```

And the test passed. How can I make it fail?

juha-aiven:
`showcase_clashing_tests` is a branch before the fixes where I removed all but two tests in `test_schema.py`.

```
git checkout showcase_clashing_tests
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml down
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d
```

Run this repeatedly and observe the failures are not deterministic:

```
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ --kafka-bootstrap-servers 127.0.0.1:9092 tests/integration/test_schema.py:
```

Running the tests sequentially works:

``` 
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ --kafka-bootstrap-servers 127.0.0.1:9092 tests/integration/test_schema.py::test_compatibility_endpoint
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ --kafka-bootstrap-servers 127.0.0.1:9092 tests/integration/test_schema.py::test_record_schema_compatibility
```

hackaugusto:
I didn't fail for me, I ran it 100 sequentially and 10 times in parallel with:
```
git checkout showcase_clashing_tests
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml down
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d
for _ in $(seq 100); do
 pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ --kafka-bootstrap-servers 127.0.0.1:9092 tests/integration/test_schema.py::test_compatibility_endpoint;
done
```
and
```
git checkout showcase_clashing_tests
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml down
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d
for _ in $(seq 10); do
 pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ --kafka-bootstrap-servers 127.0.0.1:9092 tests/integration/test_schema.py::test_compatibility_endpoint &
done
```

hackaugusto:
For documentation purposes. The issue was *not* running tests in parallel. In particular, the [`test_compatibility_endpoint`](https://github.com/aiven/karapace/blob/7b08e56f0bcd3eeb1c91b63772f88a883318baba/tests/integration/test_schema.py#L222-L223) is parametrized and there are two tests running in parallel.

I did end up reproducing the issue with the help of @juha-aiven , but only when running `test_compatibility_endpoint` in the same test run as `test_record_schema_compatibility`. In particular:

Running these two tests in parallel is green:

```sh
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ \
    --kafka-bootstrap-servers 127.0.0.1:9092 \
    tests/integration/test_schema.py::test_record_schema_compatibility
```

Running these two tests in parallel is green:

```sh
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ \
    --kafka-bootstrap-servers 127.0.0.1:9092 \
    tests/integration/test_schema.py::test_compatibility_endpoint
```

But running these four is not:

```sh
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ \
    --kafka-bootstrap-servers 127.0.0.1:9092 \
    tests/integration/test_schema.py::test_compatibility_endpoint \
    tests/integration/test_schema.py::test_record_schema_compatibility
```

Neither these four is not:

```sh
pytest -v --registry-url http://127.0.0.1:8081 --rest-url http://127.0.0.1:8082/ \
    --kafka-bootstrap-servers 127.0.0.1:9092 \
     tests/integration/test_schema.py::test_record_schema_compatibility \
     tests/integration/test_schema.py::test_compatibility_endpoint
```

And interestingly it is always the same test with its two variants that fails.

hackaugusto:
> Perhaps the issue is a race condition in SR?

Yep. I added a `time.sleep` in the beginning of `test_record_schema_compatibility` and the issue went away. I wonder why having the extra field had the same effect

PR #226, opened 2021-05-24
Handle gracefully if no node is master eligible

tvainika:

Karapace configuration allows configuring node to not be eligible for master.  Handle gracefully ie. read-only mode if all nodes are configured non-eligible for master.

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -30,7 +30,8 @@ class SchemaCoordinator(BaseCoordinator):
     hostname = None
     port = None
     scheme = None
-    master = None
+    are_we_master = None
+    is_master_eligible = None
     master_url = None
     master_eligibility = True
     log = logging.getLogger("SchemaCoordinator")
@@ -49,16 +50,25 @@ class SchemaCoordinator(BaseCoordinator):
 
     def _perform_assignment(self, leader_id, protocol, members):
         self.log.info("Creating assignment: %r, protocol: %r, members: %r", leader_id, protocol, members)
-        self.master = None
+        self.are_we_master = None
         error = NO_ERROR
         urls = {}
+        fallback_urls = {}
         for member_id, member_data in members:
             member_identity = json.loads(member_data.decode("utf8"))
             if member_identity["master_eligibility"] is True:
                 urls[get_identity_url(member_identity["scheme"], member_identity["host"],
                                       member_identity["port"])] = (member_id, member_data)
-        self.master_url = sorted(urls, reverse=self.election_strategy.lower() == "highest")[0]
-        schema_master_id, member_data = urls[self.master_url]
+            else:
+                fallback_urls[get_identity_url(member_identity["scheme"], member_identity["host"],
+                                               member_identity["port"])] = (member_id, member_data)
+        if len(urls) > 0:
+            self.master_url = sorted(urls, reverse=self.election_strategy.lower() == "highest")[0]
+            schema_master_id, member_data = urls[self.master_url]
+        else:
+            # Protocol guarantees there is at least one member thus if urls is empty, fallback_urls cannot be
+            self.master_url = sorted(fallback_urls, reverse=self.election_strategy.lower() == "highest")[0]
+            schema_master_id, member_data = fallback_urls[self.master_url]
         member_identity = json.loads(member_data.decode("utf8"))
         identity = self.get_identity(
             host=member_identity["host"],
@@ -92,10 +102,12 @@ class SchemaCoordinator(BaseCoordinator):
         )
         if member_assignment["master"] == member_id:
             self.master_url = master_url
-            self.master = True
+            self.are_we_master = True
+            self.is_master_eligible = member_identity["master_eligibility"]
         else:
             self.master_url = master_url
-            self.master = False
+            self.are_we_master = False
+            self.is_master_eligible = member_identity["master_eligibility"]
         return super(SchemaCoordinator, self)._on_join_complete(generation, member_id, protocol, member_assignment_bytes)
 
     def _on_join_follower(self):
@@ -160,7 +172,7 @@ class MasterCoordinator(Thread):
     def get_master_info(self):
         """Return whether we're the master, and the actual master url that can be used if we're not"""
         with self.lock:
-            return self.sc.master, self.sc.master_url
+            return self.sc.are_we_master, self.sc.is_master_eligible, self.sc.master_url
 
     def close(self):
         self.log.info("Closing master_coordinator")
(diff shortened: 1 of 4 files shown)


karapace/master_coordinator.py
```diff
@@ -49,16 +50,25 @@ def group_protocols(self):
 
     def _perform_assignment(self, leader_id, protocol, members):
         self.log.info("Creating assignment: %r, protocol: %r, members: %r", leader_id, protocol, members)
-        self.master = None
+        self.are_we_master = None
         error = NO_ERROR
         urls = {}
+        fallback_urls = {}
         for member_id, member_data in members:
```
hackaugusto, line 57: Suggestion, I think this is a bit more straightforward:

```python
choose_highest = self.election_strategy.lower() == "highest"

# Make sure master eligible nodes are in front of the *sorted* list
def rank(member: dict) -> int:
    if member["master_eligibility"] is True:
        return 1 if choose_highest else -1
    return 0

members_parsed_data = [
    (member_id, json.loads(member_data.decode("utf8")))
    for member_id, member_data in members
]
members_with_rank = [
    (
        rank(member_data),
        member_identity["scheme"],
        member_identity["host"],
        member_identity["port"],
        member_id,
        member_data,
    )
    for member_id, member_data in members_parsed_data
]
master = sorted(members_with_rank, reverse=choose_highest)[0]
master_identity = self.get_identity(
    host=master[2],
    port=master[3],
    scheme=master[1],
    json_encode=False,
)
```

karapace/master_coordinator.py
```diff
@@ -160,7 +172,7 @@ def init_schema_coordinator(self):
     def get_master_info(self):
```
hackaugusto, line 172: Please add typing information.

hackaugusto (changes requested):
I think it doesn't make sense to have a node that is not master eligible set as master in the metadata.

karapace/master_coordinator.py
hackaugusto: Note: Perhaps instead of having `rank` like we have above, we should instead filter out the nodes that have `member["master_eligibility"] == False`

karapace/master_coordinator.py
```diff
@@ -49,16 +50,25 @@ def group_protocols(self):
...
+        self.are_we_master = None
         error = NO_ERROR
         urls = {}
+        fallback_urls = {}
         for member_id, member_data in members:
             member_identity = json.loads(member_data.decode("utf8"))
             if member_identity["master_eligibility"] is True:
                 urls[get_identity_url(member_identity["scheme"], member_identity["host"],
                                       member_identity["port"])] = (member_id, member_data)
-        self.master_url = sorted(urls, reverse=self.election_strategy.lower() == "highest")[0]
```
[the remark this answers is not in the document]
tvainika: I didn't get your point here. This line is old code, and it is the problematic accessing of the array out of bounds. In this PR there is a fix that the array is not accessed out-of-bounds.

The ordering has been by DNS/IP, and AFAIK it works. The values of `member_id` are originating Kafka communication.

karapace/master_coordinator.py
tvainika: To be honest, I find my code more readable, to be concrete your `rank` function has double negation for `choose_highest`(first in rank, then reverse flag for sort).

And the beef in this PR is anyway handle the case where filtering out `member["master_eligibility"] == False` lead to zero length list, and that case must be handled in one way or another.

tvainika:
> I think it doesn't make sense to have a node that is not master eligible set as master in the metadata.

I changed `are_we_master` logic and renamed `is_master_eligible` to `has_eligible_master`, so I think this clarifies the logic

karapace/master_coordinator.py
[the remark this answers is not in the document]
hackaugusto: > I didn't get your point here. This line is old code, and it is the problematic accessing of the array out of bounds. In this PR there is a fix that the array is not accessed out-of-bounds.

The point I'm raising here is not about the `IndexError`, it is about the `sorted` itself. I know the code was not introduced by this PR, but when the code was introduced really doesn't change if it works as intended or not.

Regarding the `sort` itself. I don't think there is any order in IP addresses or DNS names. This leaves two options:

1. Kafka shuffles the data of `members` around for some reason, and the original `sort` was not about finding the newest node, but about having a stable result on subsequent runs (otherwise the master would change on every run).
2. The original sort was added to order nodes by age, but unfortunately used inadequate data.

Regardless of which one happened, the second commit tried to improve the code based on the sort order, but it doesn't work as expected. So, I raised the point with `member_id` as a candidate to fix the issue. Instead of sorting by IP/DNS, maybe sorting by `member_id` would work. I just don't know the semantics of the value by heart and it needs to be checked if it will fix the issue.

karapace/master_coordinator.py
hackaugusto: > To be honest, I find my code more readable, to be concrete your rank function has double negation for choose_highest(first in rank, then reverse flag for sort).

Fair enough, I don't like that myself. Just to clarify, I think of simpler as code that uses the fewest concepts and that is a bit more declarative. And in this case I was trying to reduce the number of states in the function (the if/else statements scattered all around).

hackaugusto (changes requested):
> I changed are_we_master logic and renamed is_master_eligible to has_eligible_master, so I think this clarifies the logic

My bad, I should have been clearer on what I meant by "have a node that is not master eligible set as master in the metadata.". This change is setting the variable `master_url` even if there is not master. That variable is of type `Optional[str]`, so if its value is `None` then there should be no `master`, otherwise it should have the URL set to the current master.

The above means the variable `has_eligible_master` is unnecessary.

karapace/master_coordinator.py
```diff
@@ -49,16 +50,25 @@ def group_protocols(self):
 
     def _perform_assignment(self, leader_id, protocol, members):
         self.log.info("Creating assignment: %r, protocol: %r, members: %r", leader_id, protocol, members)
-        self.master = None
+        self.are_we_master = None
         error = NO_ERROR
         urls = {}
+        fallback_urls = {}
```
hackaugusto, line 56: I know you probably want to get this merged, just one more thing. But we don't need this, `fallback_urls` doesn't have any `master_eligibility` nodes in it, so these URLs are not really used for fallback.

tvainika: AFAIU `_perform_assignment` should return Kafka assignment. `fallback_urls` is used for returning assignment even if not setting it as master. Thus `fallback_urls` was named just based on `urls`. Do you have suggestions for better naming?

hackaugusto: My bad, I got tunnel vision here and forgot about the `assignments` below.

hackaugusto (approved):
LGTM

PR #230, opened 2021-05-26
test_schema.py test_malformed_kafka_message fixes

juha-aiven:
Works now even with external registry.

The issue is that when specifying the registry url etc on the command line `registry_async` is `None` and e.g. `topic` cannot be fetched from that. Added to fetch from the cmd line parameters instead. Not sure if this is the best/correct approach. Actually, `topic` is fixed to `_schemas`. AFAIK we don't have a way to configure that, so it should work?

```
docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d
pytest -vv tests/ --registry-url=http://localhost:8081 --rest-url=http://localhost:8082 --kafka-bootstrap-servers=localhost:9092 -k test_malformed_kafka_message
```

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -6,11 +6,13 @@ See LICENSE for details
 """
 from http import HTTPStatus
 from kafka import KafkaProducer
+from karapace import config
 from karapace.rapu import is_success
 from karapace.schema_registry_apis import KarapaceSchemaRegistry
 from karapace.utils import Client
 from tests.utils import (
-    create_field_name_factory, create_schema_name_factory, create_subject_name_factory, repeat_until_successful_request
+    create_field_name_factory, create_schema_name_factory, create_subject_name_factory, KafkaServers,
+    repeat_until_successful_request
 )
 from typing import List, Tuple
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_schema.py
```diff
@@ -2000,10 +2000,17 @@ async def test_schema_remains_constant(registry_async_client: Client) -> None:
...
-    topic = registry_async.config["topic_name"]
+async def test_malformed_kafka_message(
+    command_line: Dict[str, str], registry_async: KarapaceSchemaRegistry, registry_async_client: Client
+) -> None:
+    if registry_async:
+        topic = registry_async.config["topic_name"]
+        bootstrap_servers = registry_async.config["bootstrap_uri"]
+    else:
+        topic = "_schemas"
+        bootstrap_servers = command_line["kafka_bootstrap_servers"]
```
hackaugusto, line 2011: Please use the existing `kafka_servers` fixture instead. The fixture will properly wait for the servers to become available.

tests/integration/test_schema.py
```diff
@@ -2000,10 +2000,17 @@ async def test_schema_remains_constant(registry_async_client: Client) -> None:
...
-async def test_malformed_kafka_message(registry_async: KarapaceSchemaRegistry, registry_async_client: Client) -> None:
-    topic = registry_async.config["topic_name"]
+async def test_malformed_kafka_message(
+    command_line: Dict[str, str], registry_async: KarapaceSchemaRegistry, registry_async_client: Client
+) -> None:
+    if registry_async:
+        topic = registry_async.config["topic_name"]
+        bootstrap_servers = registry_async.config["bootstrap_uri"]
+    else:
+        topic = "_schemas"
```
hackaugusto, line 2010: Perhaps we should make this a constant, maybe create a `karapace.constants` file and name it `DEFAULT_SCHEMA_TOPIC`?

hackaugusto (dismissed):

hackaugusto (approved):
LGTM

HelenMel (approved):

PR #232, opened 2021-05-26
schema_registry: breaking changes in some HTTP error messages

juha-aiven:
Now HTTP error messages match with the ones coming from Schema Registry.
Adjusted test_http_headers in test_schema.py to correctly check the messages.

Since the Kafka REST and Schema Registry REST share `rapu.py`, I assume this will also change the error messages in Kafka REST? Is that OK? Do we test those errors?

### DIFF
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -202,7 +202,7 @@ class RestApp:
         if method in {"POST", "PUT"}:
             if not content_matcher:
                 http_error(
-                    message=HTTPStatus.UNSUPPORTED_MEDIA_TYPE.description,
+                    message="HTTP 415 Unsupported Media Type",
                     content_type=result["content_type"],
                     code=HTTPStatus.UNSUPPORTED_MEDIA_TYPE,
                 )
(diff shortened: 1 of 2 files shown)


tvainika:
> Since the Kafka REST and Schema Registry REST share `rapu.py`, I assume this will also change the error messages in Kafka REST? Is that OK? Do we test those errors?

AFAIU tests/integration/test_rest.py and tests/integration/test_rest_consumer.py are related tests for REST api, but those don't seem to have tests for these cases. Maybe you could add something relevant there?

hackaugusto:
Does it makes sense to test error messages though? What is the point?

juha-aiven:
> Does it makes sense to test error messages though? What is the point?

It might be an overkill to test the errors. But that's what the tests do already and they do that a lot. We can start removing the checks if we want to.

hackaugusto:
Just for documentation: No one has strong feelings about the error messages, @tvainika mentioned that these changes have introduced useful documentation, so we are keeping the asserts for the error messages

hackaugusto (approved):

PR #233, opened 2021-05-26
 schema_registry: breaking change in /schemas/ids/<>/versions

juha-aiven:
/schemas/ids/<schema_id:path>/versions now returns empty list in case nothing is found.
This is the behaviour of SR. Karapace used to fail in this case before this change.

The tests test_schema_lifecycle and test_schema_versions_deleting now works against Schema Registry (in addition to Karapace)

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -1070,25 +1070,28 @@ async def test_schema_versions_deleting(registry_async_client: Client, trail: st
     subject = create_subject_name_factory(f"test_schema_versions_deleting_{trail}")()
     schema_name = create_schema_name_factory(f"test_schema_versions_deleting_{trail}")()
 
+    field_name = create_field_name_factory(trail)()
     schema_1 = {
         "type": "record",
         "name": schema_name,
         "fields": [{
-            "name": "f1",
-            "type": "string",
+            "name": field_name,
+            "type": "string"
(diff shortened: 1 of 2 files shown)


tvainika (approved):

tests/integration/test_schema.py
```diff
@@ -1074,21 +1074,23 @@ async def test_schema_versions_deleting(registry_async_client: Client, trail: st
         "type": "record",
         "name": schema_name,
         "fields": [{
-            "name": "f1",
-            "type": "string",
```
hackaugusto, line 1078: This is an *extreme* nitpick. But maybe we should agree on a default here.

I always add the trailing comma for two reasons:

- YAPF formatting is better in this case
- When a new entry is added later, we don't have to modify the previous line, it makes using `git blame` easier because only the relevant line is changed. This is especially useful for function arguments

Because of the above I always add trailing commas. We should probably figure something out so that we don't have every other commit commas being added and then removed

juha-aiven: Yes, it seems to be a widely adopted practise to add the trailing comma in Python arrays and I completely understand why that practise is there. It just that many languages don't support that and at least my muscle memory doesn't write it automatically.

But yes, we can agree of having the trailing comma. I'm all for it.

PR #234, opened 2021-05-26
test_schema.py: test_schema_hard_delete_and_recreate

juha-aiven:
Using a unique schema name to support parallel execution.
Small improvement in the assert messages.

The test doesn't work against Schema Registry. The issue is that SR doesn't reuse the same schema id after hard delete. Karapace reuses.

We still can merge this PR and handle the incompatibility in another PR.

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -2369,11 +2369,13 @@ async def test_schema_hard_delete_whole_schema(registry_async_client: Client) ->
 
 async def test_schema_hard_delete_and_recreate(registry_async_client: Client) -> None:
     subject = create_subject_name_factory("test_schema_hard_delete_and_recreate")()
+    schema_name = create_schema_name_factory("test_schema_hard_delete_and_recreate")()
+
     res = await registry_async_client.put("config", json={"compatibility": "BACKWARD"})
     assert res.status == 200
     schema = {
         "type": "record",
-        "name": "myenumtest",
+        "name": schema_name,
         "fields": [{
             "type": {
                 "type": "enum",
@@ -2402,7 +2404,7 @@ async def test_schema_hard_delete_and_recreate(registry_async_client: Client) ->
     )
     assert res.status == 200
     assert "id" in res.json()
-    assert schema_id == res.json()["id"], "the same schema registered, the same identifier"
+    assert schema_id == res.json()["id"], "after soft delete the same schema registered, the same identifier"
 
     # Soft delete whole schema
     res = await registry_async_client.delete(f"subjects/{subject}")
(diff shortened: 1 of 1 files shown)


tests/integration/test_schema.py
```diff
@@ -2356,11 +2356,13 @@ async def test_schema_hard_delete_whole_schema(registry_async_client: Client) ->
...
 async def test_schema_hard_delete_and_recreate(registry_async_client: Client) -> None:
     subject = create_subject_name_factory("test_schema_hard_delete_and_recreate")()
+    schema_name = create_schema_name_factory("test_schema_hard_delete_and_recreate")()
+
     res = await registry_async_client.put("config", json={"compatibility": "BACKWARD"})
     assert res.status == 200
     schema = {
         "type": "record",
-        "name": "myenumtest",
+        "name": schema_name,
```
hackaugusto, line 2365: Why is this necessary?

hackaugusto: I'm getting a failure here with SR:

```
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d
pytest tests/integration/test_schema.py::test_schema_hard_delete_and_recreate --registry-url http://localhost:8081 --rest-url http://localhost:8082/ --kafka-bootstrap-servers localhost:9092
```

Results in:
```
tests/integration/test_schema.py:2415: in test_schema_hard_delete_and_recreate
    assert schema_id == res.json()["id"], "after permanent deleted the same schema registered, the same identifier"
E   AssertionError: after permanent deleted the same schema registered, the same identifier                                                                                                                                                   
E   assert 1 == 2 
```

juha-aiven: > Why is this necessary?

The state in SR that is persisted across test runs messes things up. It's cleaner to always have a fresh schema name (and no such schema in SR at the beginning of the test).

Try this. 

Do as you did, but change first `myenumtest` to something else, let's say `s1`.  Run it. Fails. Run again. Succeeds!

juha-aiven: It's supposed to fail against SR, I mentioned that in the PR. That's a small incompatibility between SR and Karapace and should probably be fixed. I haven't checked how to fix it.

hackaugusto: > It's supposed to fail against SR, I mentioned that in the PR. That's a small incompatibility between SR and Karapace and should probably be fixed. I haven't checked how to fix it.

My bad, I went mechanically through every PR and didn't pay attention to what you wrote. Thanks.

hackaugusto: > I haven't checked how to fix it.

On a hard delete we should delete the schema from the `schemas` dictionary. Then the schema should not be found when sent again:

https://github.com/aiven/karapace/blob/9dd407d69cfb5cf61b48df394b8d55ce0d50ac86/karapace/schema_reader.py#L196-L204

It would be interesting to see what happens when we do this:

- Start SR
- Publish schema / soft delete / hard delete
- Restart SR
- Publish the same schema

hackaugusto:
> The test doesn't work against Schema Registry. The issue is that SR doesn't reuse the same schema id after hard delete. Karapace reuses.

Leaving this to be handled another time.

hackaugusto (dismissed):

hackaugusto (approved):

PR #235, opened 2021-05-26
test_schema.py: split test_enum_schema into two tests

juha-aiven:
New ones: test_enum_schema and test_enum_schema_nested

The tests run correctly against Schema Registry. I think the tests now are correct.

The tests DON'T run against Karapace. I think Karapace enum handling is buggy.

### DIFF
--- a/karapace/compatibility/__init__.py
+++ b/karapace/compatibility/__init__.py
@@ -7,8 +7,7 @@ See LICENSE for details
 from enum import Enum, unique
 from jsonschema import Draft7Validator
 from karapace.avro_compatibility import (
-    ReaderWriterCompatibilityChecker as AvroChecker, SchemaCompatibilityResult, SchemaCompatibilityType,
-    SchemaIncompatibilityType
+    ReaderWriterCompatibilityChecker as AvroChecker, SchemaCompatibilityResult, SchemaIncompatibilityType
 )
 from karapace.compatibility.jsonschema.checks import compatibility as jsonschema_compatibility
 from karapace.schema_reader import SchemaType, TypedSchema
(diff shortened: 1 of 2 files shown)


juha-aiven:
Could you @hackaugusto or @tvainika have a look at the enum logic and if there really is a bug in Karapace.

You can also take the `master` version of the test and run that against SR and see how that fails.

tests/integration/test_schema.py
```diff
@@ -700,56 +700,98 @@ async def test_map_schema_field_add_compatibility(
...
 
-        # Remove a symbol
-        schema["symbols"].pop(1)
-        res = await registry_async_client.post(f"subjects/{subject}/versions", json={"schema": jsonlib.dumps(schema)})
-        assert res.status == 200
+    res = await registry_async_client.put(f"config/{subject}", json={"compatibility": compatibility})
+    assert res.status == 200
 
-        # Change the name
+    if operation == "ADD":
```
hackaugusto, line 722: Suggestion, perhaps this could be done with a declarative approach, e.g.:
```python
ADD = "ADD"
SUCCESS = HTTPStatus.OK
TEST_OPERATIONS = [
  (ADD, modified_schema, CompatibilityModes.BACKWARD, SUCCESS,)
  ... 
]
```

karapace/compatibility/__init__.py
```diff
@@ -48,14 +47,7 @@ def is_transitive(self) -> bool:
 
 
 def check_avro_compatibility(reader_schema, writer_schema) -> SchemaCompatibilityResult:
-    result = AvroChecker().get_compatibility(reader=reader_schema, writer=writer_schema)
-    if (
```
hackaugusto, line 52: TBH I never understood why this code existed. The correct behavior is what SR does.

hackaugusto:
This failed for me when running against SR:

```
tests/integration/test_schema.py:607: in test_enum_schema_field_add_compatibility                                                                                                                                                             
    assert res.status == status_code                                                                                   
E   assert 409 == 200                                                                                                                                                                                                                         
E    +  where 409 = Result(status=409, json_result={'error_code': 409, 'message': 'Schema being registered is incompatible with an earlier schema for subject "subject_test_enum_schema_field_add_compatibility-__1_390078e4"'}).status
```

karapace/compatibility/__init__.py
juha-aiven: The same. I tried to understand the reasoning behind it but couldn't. It's basically just eating (valid) compatibility check errors. It could be a leftover of previous functionality or something like that.

hackaugusto:
The tests are not passing against 6.1.1, closing it for now.

PR #240, opened 2021-05-28
KarapaceAll: startup fix

juha-aiven:
When started from `KarapaceAll`, the `__init__` of KarapaceSchemaRegistry is not called.
schema_lock is initialized in `__init__`. Thus it's not called when using `KarapaceAll`.

Fix is to move `schema_lock` init to `_init()` which gets called also when using `KarapaceAll`.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -49,7 +49,6 @@ class KarapaceSchemaRegistry(KarapaceBase):
         super().__init__(config_file_path=config_file_path, config=config)
         self._add_routes()
         self._init(config=config)
-        self.schema_lock = asyncio.Lock()
 
     def _init(self, config: dict) -> None:  # pylint: disable=unused-argument
         self.ksr = None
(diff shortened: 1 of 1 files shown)


hackaugusto:
@juha-aiven the `KarapaceAll.__init__` seems a bit messy, did you try using `super().__init__()` instead? Not calling the parent classes constructors is a no-no.

tvainika:
> the `KarapaceAll.__init__` seems a bit messy, did you try using `super().__init__()` instead? Not calling the parent classes constructors is a no-no.

For me this looks that multiple inheritance is reason why `super().__init__()` doesn't work here.

tvainika (approved):
The fix looks correct to me. It would be nice to have an integration test for `KarapaceAll` init to avoid issues like this in the future.

juha-aiven:
I'm having a look to refactor this to make it nicer / cleaner / simpler. But, this fix should fix things in that sense that `KarapaceAll` works.

PR #241, opened 2021-05-28
Fix super

hackaugusto:
Fixes the issue raised here: https://github.com/aiven/karapace/pull/240#issuecomment-850317904 

This was tested with:

```sh
sudo docker-compose -f ./tests/integration/confluent-docker-compose.yml up -d kafka
karapace karapace.config.json
```

And then:

```sh
pytest tests/integration -n0 --registry-url http://localhost:8081 --rest-url http://localhost:8081/ --kafka-bootstrap-servers localhost:9092
```

Two tests failed, but those are tests bugs.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -37,12 +37,12 @@ class FormatError(Exception):
 
 class KafkaRest(KarapaceBase):
     # pylint: disable=attribute-defined-outside-init
-    def __init__(self, config_file_path: str, config: dict) -> None:
-        super().__init__(config_file_path=config_file_path, config=config)
-        self._add_routes()
-        self._init(config=config)
+    def __init__(self, config: dict) -> None:
+        super().__init__(config=config)
+        self._add_kafka_rest_routes()
+        self._init_kafka_rest(config=config)
 
-    def _init(self, config: dict) -> None:
+    def _init_kafka_rest(self, config: dict) -> None:
         self.serializer = SchemaRegistrySerializer(config=config)
         self.log = logging.getLogger("KarapaceRest")
         self.loop = asyncio.get_event_loop()
@@ -58,7 +58,7 @@ class KafkaRest(KarapaceBase):
         self.producer_refs = []
         self.producer_queue = asyncio.Queue()
 
-    def _add_routes(self):
+    def _add_kafka_rest_routes(self) -> None:
         # Brokers
         self.route("/brokers", callback=self.list_brokers, method="GET", rest_request=True)
 
@@ -328,8 +328,9 @@ class KafkaRest(KarapaceBase):
         self.producer_queue = None
         return
 
-    def close(self):
-        super().close()
+    async def close(self) -> None:
+        await super().close()
+        await self.close_producers()
         if self.admin_client:
             self.admin_client.close()
             self.admin_client = None
--- a/karapace/karapace.py
+++ b/karapace/karapace.py
@@ -22,20 +22,21 @@ logging.basicConfig(level=logging.INFO, format=LOG_FORMAT_JOURNAL)
 
 
 class KarapaceBase(RestApp):
-    def __init__(self, config_file_path: str, config: dict) -> None:
+    def __init__(self, config: dict) -> None:
+        sentry_config = config.get("sentry", {"dsn": None}).copy()
+        super().__init__(app_name="karapace", sentry_config=sentry_config)
+
         self.config = {}
         self.producer = None
         self.kafka_timeout = 10
-        self.config_path = config_file_path
         self.config = config
-        self._sentry_config = self.config.get("sentry", {"dsn": None}).copy()
+        self._sentry_config = sentry_config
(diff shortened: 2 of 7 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -67,98 +67,108 @@ def _add_routes(self):
             "/consumers/<group_name:path>/instances/<instance:path>/offsets",
             callback=self.commit_consumer_offsets,
             method="POST",
-            rest_request=True
+            rest_request=True,
```
hackaugusto, line 70: these changes were done automatically by the formatting tool.

karapace/karapace.py
```diff
@@ -22,20 +22,21 @@
 
 
 class KarapaceBase(RestApp):
-    def __init__(self, config_file_path: str, config: dict) -> None:
+    def __init__(self, config: dict) -> None:
+        sentry_config = config.get("sentry", {"dsn": None}).copy()
+        super().__init__(app_name="karapace", sentry_config=sentry_config)
```
hackaugusto, line 27: Not strictly necessary, but it is always easier to reason about these things if `super().__init__` is always done at start or end.

juha-aiven:
Can you separate the formatting changes from the actual code changes? Now they're mixed together and it's hard to read what's changed.

hackaugusto:
> Can you separate the formatting changes from the actual code changes? Now they're mixed together and it's hard to read what's changed.

tvainika:
As this issue with initialization and multi-inheritance came up, so to clarify my understanding: has `_init_kafka_rest` and `_init_schema_registry` still have rationale of being there, or should those be squashed into respective `__init__` methods? For me, if I understand the original code right, those `_init` has been added as workaround multi-inheritance initialization instead of fixing it this way (not sure if earlier Python versions had something to do with this)?

hackaugusto:
> not sure if earlier Python versions had something to do with this?

I couldn't find when multiple inheritance was introduced in Python, I believe it was in early 2.x series. But I know that inheritance works the same way for almost 2 decades, [since 2.3](https://www.python.org/download/releases/2.3/mro/). 

> For me, if I understand the original code right, those _init has been added as workaround multi-inheritance initialization

Maybe, but if that was the case it was incomplete.

This is roughly how it works:

- If a method doesn't call `super`, this method shadows the others
- Otherwise `super` is used and we have cooperative inheritance, were we use `super` to call the next method in the method-resolution-order (**Note**: this may not necessarily be a parent's method, it depends on the hierarchy)

In this case, having both classes with a method named `_init` without calling `super()` means implementation shadows the other, which is not what we want:

```python
>>> class A:
...   def f(self): print('A')
... 
>>> class B:
...   def f(self): print('B')
... 
>>> class Test1(A,B): pass
... 
>>> class Test2(B,A): pass
... 
>>> Test1().f()
A
>>> Test2().f()
B
```

In the example above, the method `f` is being shadowed, and the result depends on the order of the definition of the parent classes. This is what the code used to be, and this is not correct for our use case, we want both `_init` to be called.

There are two possible fixes here:

1. Change the code to participate in the cooperative inheritance. In this case we would need to:
  - Add a method `_init` in a shared parent of both classes. In our code that would be `KarapaceBase`, the method in that class would implement the base case for the inheritance.
  - Add the missing call to `super` everywhere but `KarapaceBase._init` (because this is the base case, the parent doesn't have `_init`)
2. Do what I did in the PR, have a unique name for each `_init`, so that one does not shadows the other, and we can call both of them.

karapace/karapace_all.py
```diff
@@ -11,29 +10,17 @@
...
         # pylint: disable=unused-argument
         self.close()
 
-    def close(self):
-        for cls in [KarapaceBase, KafkaRest, KarapaceSchemaRegistry]:
-            try:
-                cls.close(self)
-            except:  # pylint: disable=bare-except
-                pass
-            KafkaRest.close_producers(self)
```
tvainika, line 35: After these changes `close_producers` is no longer called at all as far as I can see (except in integration tests)? Should it be called or eliminated altogether if no longer needed?

Of course existing code is a bit oddity as it is calling async method here without waiting for completion.

hackaugusto: Good catch. Since that is part of `KafkaRest` IMO it done on `KafkaRest.close`. And we definitely should wait on it. This code here was just creating a coroutine that was never executed, so it had no effect.

It also looks like we are not calling `close` in the main method. It seems this method was never called.

tvainika: Isn't `close`method called from `close_by_app` which is triggered by signal handler?

hackaugusto: > Isn't closemethod called from close_by_app which is triggered by signal handler?

Yep, you're right. [Here is the docs](https://docs.aiohttp.org/en/stable/web_advanced.html?highlight=on_shutdown#graceful-shutdown)

hackaugusto: So, the code example in the docs uses an async function. So I converted the `close` methods to async and now we can call `close_producers` there.

tvainika (approved):

PR #251, opened 2021-06-04
bugfix: logging must not be configured at import time

hackaugusto:
This was bugging me for a while, it made the logging output much more verbose and harder to parse.

```
the `basicConfig` at the top-level was conflicting with the test suite,
the end result is that some of the existing handlers lost their output
file and the logging failed with:

    --- Logging error ---
      File "/usr/lib64/python3.8/logging/__init__.py", line 1088, in emit
        stream.write(msg + self.terminator)

This really cluttered the output and made debugging tests harder
```


tvainika (changes requested):
This fix looks sound to me, but there was now conflicts after merging other PRs.

hackaugusto:
conflicts fixed

tvainika (approved):

PR #254, opened 2021-06-23
Add example command to register JSON Schema

### LINKED ISSUE #253: What's the correct Content-Type header for JSON schema?
# What can we help you with?

I'm trying to POST a new JSON schema, but I'm getting back a 415 error code with the message "HTTP 415 Unsupported Media Type". I'm using the headers I found at https://github.com/aiven/karapace/blob/6c7ed9c811e47923d83ab2fc649537b59ce0a16e/tests/utils.py#L89-L92 so

```
--header 'Content-Type: application/vnd.kafka.jsonschema.v2+json' \
--header 'Accept: application/vnd.kafka.jsonschema.v2+json, application/vnd.kafka.v2+json, application/json, */*' 
```
but no luck. It actually seems like `application/vnd.schemaregistry.v1+json` and `application/json` are the only accepted content type, but both assume an Avro schema.

# Where would you expect to find this information?

It would be great to have example `curl` requests in the Quickstart section of the repo's README for all supported schema types

tvainika:

Add command into readme how to register a version of a schema using JSON Schema. Fixes #253

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,13 +74,19 @@ Alternatively you can do a source install using::
 Quickstart
 ==========
 
-To Register the first version of a schema under the subject "test"::
+To register the first version of a schema under the subject "test" using Avro schema::
 
   $ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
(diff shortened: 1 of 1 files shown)


hackaugusto (approved):
LGTM

PR #255, opened 2021-06-23
[WIP] Add API documentation for Karapace

### LINKED ISSUE #181: REST API Documentation
hello!
Is there a place where I could review the endpoints documentation? 
Similar to https://api.aiven.io/doc/#operation/ServiceKafkaTopicCreate

Thanks in advance

lornajane:
# Let's start assembling formal API documentation for Karapace

Fixes #181 

There doesn't seem to be much python tooling for openapi (sorry) so this introduces some NodeJS dependencies. They are run using `npx` and are wrapped by the `make openapi` step which validates the description and then generates an HTML page for docs.

The initial OpenAPI spec (for just the schema registry component) is in `openapi/` and you can also find rendered HTML docs there.


tvainika (changes requested):
Lint seems to fail with generated `schema-registry.html`, and in any case this needs decision if we want to put generated files into the repository, or should be use some other publishing process that does not put generated content into the repository?

lornajane:
I mostly added the rendered docs so people can quickly preview how things are progressing while we work on this but we should definitely have a plan for whether this belongs in the main branch or not. I'd be entirely happy to simply publish the OpenAPI description and people can do what they want with it. We could also render some docs to include in karapace.io, what do you think?

I'm interested if this is a good way to proceed, building the OpenAPI, and if there's any other feedback before I do a lot more with it.

hackaugusto:
closing the PR for the lack of activity, it can be reopened if necessary.

Issue #271, opened 2021-08-02
[Pipeline] Tests run twice on Pull Request Creation


jlprat:
# What happened?
When a new pull request is created the test suite is run twice:
- One because it's triggered at pull request creation
- One because there is a push event

# What did you expect to happen?
The test suite is run only once when the following happens:
- Pull request creation (pr creation + push)
- Pushing more than one commit at a time
- Re opening the pull request

jlprat:
Actually, Linting also suffers from the same thing

PR #272, opened 2021-08-03
Config GitHub actions to avoid double triggering #271

jlprat:
References: #271
Lint and tests will now run when:
- A PR is created, updated, reopened...
- A push event happens on the master branch


jlprat:
I'll push some other changes to test that the config works as expected (no double triggering)

jlprat:
Ready for review

PR #275, opened 2021-08-03
Clean up obsolete sections and remove "simply" wording

jlprat:
Words like easy, painless, straightforward, trivial, simple and just
when referring to simplicity should be avoided from any docs.
Just might be OK to use when it referst to the temporal adverb, like
"the thing we did just before".


hackaugusto (approved):
LGTM

PR #282, opened 2021-09-21
schemas: better error messages for invalid schemas

hackaugusto:
This provides a bit more of context of the error to the user

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -1,6 +1,8 @@
+from avro.schema import SchemaParseException
 from contextlib import closing
 from enum import Enum, unique
 from http import HTTPStatus
+from json import JSONDecodeError
 from karapace import version as karapace_version
 from karapace.avro_compatibility import is_incompatible
 from karapace.compatibility import check_compatibility, CompatibilityModes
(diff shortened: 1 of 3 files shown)


karapace/schema_registry_apis.py
```diff
@@ -761,12 +763,16 @@ def write_new_schema_local(self, subject, body, content_type):
         schema_type = SchemaType(body.get("schemaType", SchemaType.AVRO))
         try:
             new_schema = TypedSchema.parse(schema_type=schema_type, schema_str=body["schema"])
-        except (InvalidSchema, InvalidSchemaType):
+        except (InvalidSchema, InvalidSchemaType) as e:
             self.log.warning("Invalid schema: %r", body["schema"], exc_info=True)
+            if isinstance(e.__cause__, (SchemaParseException, JSONDecodeError)):
+                human_error = f"Error: {e.__cause__.args[0]}"  # pylint: disable=no-member
+            else:
+                human_error = "Provided schema is not valid"
```
hackaugusto, line 771: using `str(e)` would show the stacktrace to the user, which may be worse than having a generic message. Another alternative would be to instead of having cases for exception types, just check if `args` is not empty and assume the first argument is the error message:

```python
if hasattr(e.__cause__, "args") and e.__cause__.args:
    human_error = f"Error: {e.__cause__.args[0]}"
````

tvainika (approved):

PR #284, opened 2021-09-28
Fix memory leak

hackaugusto:
Fix for https://github.com/aiven/karapace/issues/283


willyborankin (changes requested):
During discussion a bug was found

PR #288, opened 2021-11-03
Improved error message for the addition of a property to an open model

hackaugusto:
This tries to make the error message a bit more informative, users may be unaware that the lack of a property in a schema means that any value is acceptable.

I also sneaked a mypy config in there, just to get that rolling.


mhoffm-aiven (approved):
LGTM

PR #289, opened 2021-11-24
fix compatibility with aiohttp >= 3.8.0 and aiokafka >= 0.7.0

kmichel-aiven:
Fix compatibility with aiohttp >= 3.8.0 and aiokafka >= 0.7.0

# Why this way
## aiokafka
Removed the loop parameter and let python use its default behavior.

## aiohttp
Deferred the creation of the client to the first async function that needs it.
This might be a problematic approach if `utils.Client` is a public API and
can't have its `client` parameter removed.

SchemaRegistryClient didn't need to force the `client` value since it was
equivalent to the default.

While the `aiohttp.ClientSession` doesn't require the factory function
to be async, the `aiohttp_client` used during tests needed it.

### DIFF
--- a/karapace/utils.py
+++ b/karapace/utils.py
@@ -9,6 +9,7 @@ from http import HTTPStatus
 from kafka.client_async import BrokerConnection, KafkaClient, MetadataRequest
 from urllib.parse import urljoin
 
+import aiohttp
 import datetime
 import decimal
 import json as jsonlib
@@ -81,15 +82,21 @@ class Result:
         return 200 <= self.status_code < 300
 
 
+async def get_aiohttp_client() -> aiohttp.ClientSession:
+    return aiohttp.ClientSession()
+
+
 class Client:
-    def __init__(self, server_uri=None, client=None):
+    def __init__(self, server_uri=None, client_factory=get_aiohttp_client):
         self.server_uri = server_uri or ""
         self.path_for = partial(urljoin, self.server_uri)
         self.session = requests.Session()
-        self.client = client
-        if self.client is None:
-            import aiohttp
-            self.client = aiohttp.ClientSession()
+        # aiohttp requires to be in the same async loop when creating its client and when using it.
+        # Since karapace Client object is initialized before creating the async context, (in
+        # kafka_rest_api main, when KafkaRest is created), we can't create the aiohttp here.
+        # Instead we wait for the first query in async context and lazy-initialize aiohttp client.
+        self.client_factory = client_factory
+        self._client = None
 
     async def close(self):
         try:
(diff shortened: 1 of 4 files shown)


tests/integration/conftest.py
```diff
@@ -328,10 +333,13 @@ async def fixture_registry_async_client(
     if registry_url:
         client = Client(server_uri=registry_url)
     else:
-        client_factory = await aiohttp_client(registry_async.app)
-        client = Client(client=client_factory)
 
-    with closing(client):
```
hackaugusto, line 334: was this removed because closing client is async? Because there is an async version of closing which would be better suited: https://docs.python.org/3/library/contextlib.html#contextlib.aclosing

karapace/utils.py
```diff
@@ -81,32 +81,42 @@ def ok(self):
...
-    def __init__(self, server_uri=None, client=None):
+    def __init__(self, server_uri=None, client_factory=None):
         self.server_uri = server_uri or ""
         self.path_for = partial(urljoin, self.server_uri)
         self.session = requests.Session()
-        self.client = client
-        if self.client is None:
-            import aiohttp
-            self.client = aiohttp.ClientSession()
+        self.client_factory = get_aiohttp_client if client_factory is None else client_factory
```
hackaugusto, line 94: Why is the code no longer RAII?

hackaugusto (changes requested):
The requirements.txt needs to be updated

karapace/utils.py
kmichel-aiven: Client is created outside of an async context.
aiohttp >= 3.8.0 doesn't like that because it remembers the current async loop (or lack of loop) at init time and then complains loudly if the loop is different when the client is actually used. That's why I had to change it to lazy initialization.

Another option would be to reshape the main to have the `KafkaRest` class be initialized in an async context,
which means removing/reshaping `kc.run` (which actually calls `aiohttp.web.run_app`, which starts the async context and "should" be used directly as the main entry point of the program, according to [their doc](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.run_app)):
https://github.com/aiven/karapace/blob/560a4ad64bfbf9ecbdcd7dd9be6f804bfc8a6986/karapace/kafka_rest_apis/__main__.py#L22-L24
Given all the inheritance happening around rapu, that might not be so easy, deferring only `_init_kafka_rest` could be an option:
https://github.com/aiven/karapace/blob/560a4ad64bfbf9ecbdcd7dd9be6f804bfc8a6986/karapace/kafka_rest_apis/__init__.py#L40-L46

tests/integration/conftest.py
kmichel-aiven: Yes, it's because closing client is async.
aclosing would be better but it's python >= 3.10.

hackaugusto: @kmichel-aiven fair point, I didn't look at the supported version.

karapace/utils.py
hackaugusto: @kmichel-aiven thanks for the explanation. Maybe just add a comment in the code? I would only have figure this out when trying to make the code RAII and failing :sweat_smile:

kmichel-aiven:
> The requirements.txt needs to be updated

I'm not sure what I should do there. aiohttp is not listed (it is in `setup.py` though), but everything has completely pinned versions.
This PR adds compatibility with newer aiohttp but does not break it with older ones.
Breaking installs where people only have access to older versions wouldn't be great (or maybe both `requirements.txt` and `requirements-dev.txt` are for devs only, and everybody else only gets the unpinned dependency constraints from `setup.py` ?)

hackaugusto (approved):

Issue #290, opened 2021-12-06
Unhandled exception when Request Entity is too large 


jlprat:
# What is currently missing?

Currently when a request's entity is too large, Karapace spams any exception log system with the stacktrace and more.

# How could this be improved?

This exception shouldn't be left unhandled and Karapace should log this situation instead of letting it bubble up.

# Is this a feature you would work on yourself?

- [ ] I plan to open a pull request for this feature

PR #291, opened 2021-12-07
karapace/rapu: handle HTTPRequestEntityTooLarge in rapu

mhoffm-aiven:
This error is expected and should be a client error instead of 500.

If the client POSTs a request body that is too large, we catch the internal aiohttp and wrap it into a client facing user error response instead of returning a internal server error response.


jlprat:
Resolves #290

hackaugusto (approved):
LGTM

PR #294, opened 2021-12-16
Revert avro validation

hackaugusto:
If there are invalid schemas already registered in the topic, the schemas won't be loaded. This needs to be handled only during registration.


rikonen:
Linter failure

ivanyu:
I'll fix

tvainika (approved):
LGTM.

PR #297, opened 2021-12-20
Fix REST and schema registry shutdown

jjaakola-aiven:
Shutdown handler for aiohttp Application was set only for
KarapaceAll. On some environments where REST application and
schema registry application is run separately, e.g. containers,
the shutdown times out and container manager sends kill signal.

Fix is to move the shutdown handler registration from KarapaceAll
to KarapaceBase.


tvainika (approved):
Good catch

tvainika (approved):

PR #298, opened 2021-12-28
Build container image in GitHub Action

tvainika:

Build container image automatically from the main branch commits with `develop` tag and versioned container images from version tags, including `latest` tag pointing to most recent release.

This used GitHub Container Registry instead of Docker Hub.

# Why this way

Usage of GitHub Container Registry integrates better with tooling, and fetching container images from GHCR is as easy as Docker Hub.

Usage of `latest` tag from most recent release tag requires linear release strategy. If a need to maintain multiple branches arises, this needs some adjustments.

I've also tested building container images in my forked repository by using both main branch and arbitrary release tags there, see https://github.com/tvainika/karapace/pkgs/container/karapace

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,57 @@
+name: Create and publish a container image
+
+on:
+  push:
+    # Publish main branch as `develop` image
+    branches:
+      - main
+      - master
+
+    # Publish tags as versioned release and `latest` image
+    tags:
+      - '*'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Log in to registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Build and push container image
+        run: |
+          # Image name may contain only lowercase letters
+          IMAGE_ID=$(echo ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }} | tr '[A-Z]' '[a-z]')
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+          # If not tag, then use `develop` as image tag
+          [ "$VERSION" == main -o "$VERSION" == master ] && VERSION=develop
+
+          echo IMAGE_ID=$IMAGE_ID
+          echo VERSION=$VERSION
+
+          docker build . --file container/Dockerfile \
+              --build-arg "CREATED=$(date --rfc-3339=seconds)" \
+              --build-arg "VERSION=$(git describe --always)" \
+              --build-arg "COMMIT=$(git rev-parse -q --verify HEAD^{commit})" \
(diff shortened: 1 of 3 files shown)


mdedetrich (dismissed):
This looks good from my end so I will approve

.github/workflows/container.yml
```diff
@@ -0,0 +1,57 @@
...
+    branches:
+      - main
+      - master
+
+    # Publish tags as versioned release and `latest` image
+    tags:
+      - '*'
+
+env:
+  REGISTRY: ghcr.io
```
hackaugusto, line 15: I think we should either:

- update the documentation, since ATM it assumes we are publishing the image on the Docker Hub registry
- also publish to Docker Hub, because AFAIK there are people already using that on their builds

jjaakola-aiven (dismissed):

.github/workflows/container.yml
```diff
@@ -0,0 +1,57 @@
...
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Log in to registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
```
hackaugusto, line 30: suggestion:

Use the existing [action for logging in](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages)

hackaugusto (dismissed):
We need to clarify the details about the registry

.github/workflows/container.yml
tvainika: Which documentation you are referring? I didn't see any. I updated reference docker-compose files to use new GHCR.

hackaugusto: > Which documentation you are referring? I didn't see any.

That is the point, we need to document where the image is published. The commit with the pudated docker-compose.yml seems to be missing.

tvainika: True, I had forgot to push docker-compose updates. However I cannot see any references to Docker Hub here, so I don't understand what you expect to be changed now? Naturally at some point obsolete image from Docker Hub can be cleaned, but also we need to make new releases of Karapace. Now Docker Hub contains only one very old image, but which is also AFAIK latest versioned release of Karapace...

.github/workflows/container.yml
tvainika: "A little copying is better than a little dependency."

That action is for unknown reasons pinned to specific version in the documentation you linked, and using third party action instead of one-liner is IMHO harder to read than this, which is widely used.

.github/workflows/container.yml
hackaugusto: > so I don't understand what you expect to be changed now? 

I guess you're saying that because GHCR won't start with an image of a release it should not be documented?

jjaakola-aiven (commented):
The`container/publish_image.sh` could be removed.

README.rst
```diff
@@ -50,7 +50,15 @@ Setup
...
 ------------
 
-To get you up and running with the latest release of Karapace, a docker setup is available::
+To get you up and running with the latest build of Karapace, a docker image is available::
+
+  # Fetch the latest build from main branch
+  docker pull ghcr.io/aiven/karapace:develop
+
+  # Fetch the latest release
+  docker pull ghcr.io/aivven/karapace:latest
```
jjaakola-aiven, line 59:
```suggestion
  docker pull ghcr.io/aiven/karapace:latest
```

hackaugusto (approved):
LGTM

PR #302, opened 2022-01-28
Add support for Python 3.10

jongiddy:
Updates to support testing and running Karapace using Python 3.10

Mostly this requires updating some lint packages to fix problems they had with the latest version of Python. And then the updated versions of the linters found some different problems to fix or ignore.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -400,7 +400,7 @@ class KafkaRest(KarapaceBase):
         await self.publish(topic, None, content_type, request.content_type, request.json)
 
     @staticmethod
-    def validate_partition_id(partition_id: str, content_type: str) -> int:
+    def validate_partition_id(partition_id: str, content_type: str) -> int:  # pylint: disable=inconsistent-return-statements
         try:
             return int(partition_id)
         except ValueError:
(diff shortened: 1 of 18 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -400,7 +400,7 @@ async def topic_publish(self, topic: str, content_type: str, *, request):
         await self.publish(topic, None, content_type, request.content_type, request.json)
 
     @staticmethod
-    def validate_partition_id(partition_id: str, content_type: str) -> int:
+    def validate_partition_id(partition_id: str, content_type: str) -> int:  # pylint: disable=inconsistent-return-statements
```
tvainika, line 403: Note for later action: refactor `KafkaRest.not_found` and `.r` methods such that the code has `raise` statement here instead of there. Should improve readability, and avoid this pylint ignore, right?

tvainika (dismissed):

tvainika:
Looked good in my eyes, but it seems that the tests catched something with Python 3.10?

jongiddy:
Updated `requirements.txt` to use newer `aiokafka` that supports 3.10, and to pin indirect dependencies.

tvainika (approved):

PR #303, opened 2022-01-28
Enable black

hackaugusto:
# What it does

Replaces YAPF with Black

# Why this way

Black is faster and more predictable


hackaugusto:
converted to draft:

- current version of pylint is [2.4.4](https://github.com/aiven/karapace/blob/master/requirements-dev.txt#L4)
- pylint [requires isort>=4.2.5,<5](https://github.com/PyCQA/pylint/blob/pylint-2.4.4/pylint/__pkginfo__.py#L32)
- isort>5.0.0 is being here because it has a black compatible profile

hackaugusto:
note: #302 fixed most of the issues I mentioned above. When this PR is merged I will upgrade isort and fix the configuration.

jongiddy (dismissed):
Nice work. I wish `make lint` would complain but not change code, and `make reformat` and `pre-commit` can change the code, but that is an unchanged state of affairs.

jongiddy:
Not merging because there is a conflict.

jongiddy:
Black and Pylint are both checking line length. Should we disable `line-too-long` for PyLint?

hackaugusto:
> Nice work. I wish make lint would complain but not change code, and make reformat and pre-commit can change the code, but that is an unchanged state of affairs.

I was wondering if it makes sense to have the Makefile rules. Don't you think pre-commit is enough?

hackaugusto:
> Black and Pylint are both checking line length. Should we disable line-too-long for PyLint?

black does formatting on a best effort basis. pylint does nagging. IMO the late is good because we don't have to do it in PR reviews.

jongiddy (approved):
Thanks. Let's get this in.

PR #309, opened 2022-02-01
Add TLS termination to Karapace server

jongiddy:

Adds TLS termination to the Karapace server, so that Schema Registry and REST Proxy can be accessed using HTTPS.

References: HH-828

# Why this way


tvainika (dismissed):
LGTM, but these new configuration variables need documentation in  [README.md chapter configuration-keys](https://github.com/aiven/karapace#configuration-keys)

tvainika (approved):

PR #310, opened 2022-02-01
protobuf: using standard library dataclasses

hackaugusto:
This is the only class not using the dataclasses module, for consistency it should also be based on the std library.


RyanSkraba (approved):
LGTM for consistency

PR #311, opened 2022-02-01
tests,protobuf: simplified test_protoc

hackaugusto:
Simplified the test code:

- Exceptions are allowed to propagate, that makes the tests fail and allows `pytest --pdb` to work
- Relying on the context manager to close the file handlers
- Removing print statements
- Using a logger name based on the module's name


RyanSkraba (approved):
Looks good, very readable.

PR #312, opened 2022-02-01
git: added .git-blame-ignore-revs

hackaugusto:
Add `.git-blame-ignore-revs` with the commit that formatted the code using black.


RyanSkraba (approved):
LGTM!  Tested using `git blame karapace/avro_compatibility.py | grep 2022 | wc`.  

Before this change, :point_up: returned 39 lines, after this change none (consistent with the history).

Maybe consider adding the `git config ...` line to CONTRIBUTING.md ?

PR #314, opened 2022-02-01
protobuf: removed unused method

hackaugusto:
This method is not called


hackaugusto:
it seems this is part of the enum interface: https://docs.python.org/3/library/enum.html#supported-sunder-names

PR #316, opened 2022-02-01
karapace: using Event instead of Lock

hackaugusto:
Exclusive lock is not the right abstraction, instead the code should just make the readers wait until the state set by the writer is available, which happens only once.


hackaugusto:
this also fixes the issue warned by pylint:

```
karapace/master_coordinator.py:145:8: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
```

tvainika (approved):

PR #317, opened 2022-02-01
pylint fixes

hackaugusto:
This is fixing a few minor issues caught by pylint, the goal is to have them fixed and upgrade pre-commit to version 2.12.2


hackaugusto:
Humm, annoying, I had to revert to fixes:

```
tests/integration/conftest.py:503:0: E0012: Bad option value 'consider-using-with' (bad-option-value)
tests/integration/conftest.py:567:0: E0012: Bad option value 'consider-using-with' (bad-option-value)
```

I will add them back on the PR that actually does the upgrade

tvainika (approved):

PR #318, opened 2022-02-01
Enable mypy in precommit

hackaugusto:
- Adds mypy to the precommit
- Fixes some typing issues with code that should have been type checked
- Made the configuration for type checking opt-out instead of opt-in (the idea is that code in new modules should be typed by default)


hackaugusto:
Notes:

- Mypy treats repeated module names as an error. Because of theses, the `conftest.py` files used to configure pytest results in errors. The fix is to run mypy on each module independently https://github.com/python/mypy/issues/4008, because of this I'm not yet enabling mypy for the test code
- We support python >=3.7, [user defined type guards](https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards) were only introduced on python >=3.10, so for now the code has quite a few asserts for types

jlprat:
Do you think it would be worth to add this to CONTRIBUTING.md so people know are aware of this?

hackaugusto:
> Do you think it would be worth to add this to CONTRIBUTING.md so people know are aware of this?

We do [mention it](https://github.com/aiven/karapace/blob/master/CONTRIBUTING.md#static-checking-and-linting), the problem was that mypy was not configured to run with pre-commit.

hackaugusto:
Note: to enable mypy for the tests, something like this should be used:

```yaml
  # `mypy tests` can't be run because mypy complains about the repeated conftest.py
  # https://github.com/python/mypy/issues/4008: current solution is to run mypy separately
  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: v0.931
    hooks:
      - id: mypy
        name: Mypy integration tests
        pass_filenames: false
        args: ["--ignore-missing-imports", "tests/integration"]

  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: v0.931
    hooks:
      - id: mypy
        name: Mypy unit tests
        pass_filenames: false
        args: ["--ignore-missing-imports", "tests/unit"]
```

hackaugusto:
I did a quick look on how to model the JSON Schema using recursive types, but that is just not really feasible at the type for a few reasons:

- To represent dictionaries with non-homegenous value types we need `TypedDict` which is only >=3.8 (otherwise we would need to add the typing extensions as a dependency and that has other issues)
  - Even if we had TypedDict, it seems that it is a closed model (unknown keys are not allowed), whereas the JSON Schema object is an open model (additional keys are ignored)
- Recursion is not really supported by Mypy with type aliases, only with protocols, and as far as I can see those only works with classes, not dicts

karapace/compatibility/__init__.py
```diff
@@ -161,10 +163,6 @@ def check_compatibility(
             )
 
     else:
-        result = SchemaCompatibilityResult.incompatible(
-            incompat_type=SchemaIncompatibilityType.type_mismatch,
-            message=f"Unknow schema_type {old_schema.schema_type}",
-            location=[],
-        )
+        assert_never(f"Unknow schema_type {old_schema.schema_type}")
```
jongiddy, line 166:
```suggestion
        assert_never(f"Unknown schema_type {old_schema.schema_type}")
```

jongiddy (commented):
Looks good, apart for one spelling error.

karapace/compatibility/__init__.py
hackaugusto: fixed

jongiddy (approved):

PR #320, opened 2022-02-03
Fix spelling of eligibility, clarify docs

jongiddy:
A quick fix to the spelling of "eligibility". I also changed the docs for these settings a little to (hopefully) make them clearer.


tvainika (approved):

ivanyu:
This will require a major version bump as it's a breaking change

jongiddy:
@ivanyu Should we support the old spelling as well, so we don't need the bump?

PR #322, opened 2022-02-03
Support misspelt environment variable for compatibility

jongiddy:
The misspelling of `KARAPACE_REGISTRY_MASTER_ELIGIBILITY` was fixed but we need to support the misspelling for backwards compatibility.


hackaugusto (approved):
LGTM

PR #324, opened 2022-02-03
rest proxy: handle union fields

hackaugusto:
Fix for https://github.com/aiven/karapace/issues/319

### Avro encoding

When encoding Avro messages to JSON, per the spec, enums have to be tagged unions [ref](https://avro.apache.org/docs/current/spec.html#json_encoding).

So for this schema:

```json
{
  "namespace": "io.aiven.data",
  "name": "Test",
  "type": "record",
  "fields": [
    {
      "name": "attr",
      "type": [
        "null",
        "string"
      ]
    }
  ]
}
```
this is the correct encoding:
```json
{"attr":{"string":"sample data"}}
```

### Avro's python library

The library does not have a JSON encoder or decoder [AVRO-91](https://issues.apache.org/jira/browse/AVRO-91).

### Karapace bug

To support JSON messages, Karapace is using a JSON library, without handling the exception above. This works for everything but unions. The JSON representation needs to be transformed to the Python representation by removing the tag.

From:
```python
{"attr":{"string":"sample data"}}
```
To:
```python
{"attr":"sample data"}
```

### Cross language compatibility

Other languages have a proper JSON encoder/decoder in their Avro library (e.g. Java), applications in those languages properly encode the data with the tag. Requests from those client applications are incorrectly rejected by Karapace.

### Backwards compatibility

Karapace did not accept _properly_ encoded data because of the above. But Karapace _also_ accepted **invalid** data because of the above.

This means that there are python applications out there sending data without the correct encoding with tagged unions today.

### This PR

This PR adds compatibility for properly encoded data, i.e. JSON with the tagged unions, and it maintains compatibility for the clients sending invalid data. This is only done for the APIs writing data.

## Reader API

This will need another PR and probably a feature flag. The same way there are clients writing invalid data, there may be clients expecting invalid data. But this doesn't work with languages like Java.

### DIFF
--- a/karapace/serialization.py
+++ b/karapace/serialization.py
@@ -5,11 +5,12 @@ from karapace.protobuf.exception import ProtobufTypeException
 from karapace.protobuf.io import ProtobufDatumReader, ProtobufDatumWriter
 from karapace.schema_reader import InvalidSchema, SchemaType, TypedSchema
 from karapace.utils import Client, json_encode
-from typing import Dict, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple
 from urllib.parse import quote
 
 import asyncio
 import avro
+import avro.schema
 import io
 import logging
 import struct
@@ -183,6 +184,60 @@ class SchemaRegistrySerializerDeserializer:
         return schema_typed
 
 
+def flatten_unions(schema: avro.schema.Schema, value: Any) -> Any:
+    """Recursively flattens unions to convert Avro JSON payloads to internal dictionaries
+
+    Data encoded to Avro JSON has a special case for union types, values of these type are encoded
+    as tagged union. The additional tag is not expected to be in the internal data format and has to
+    be removed before further processing. This means the JSON document must be further processed to
+    remove the tag, this function does just that, recursing over the JSON document and handling the
+    tagged unions.
+
+    Given this schema:
+
+        {"name": "Test", "type": "record", "fields": [{"name": "attr", "type": ["null", "string"]}]}
+
+    The record JSON encoded as:
+
+        {"attr":{"string":"sample data"}}
+
+    The python representation is:
+
+        {"attr":"sample data"}
+
+    This function:
+
+    - Translates the first to the second when necessary, this adds compatibility for libraries that
+      perform the _correct_ encoding.
+    - Does nothing if the provided data is already in the second format. The data is improperly
+      encoded, but this maintains backwards compatibility.
+
+    See also https://avro.apache.org/docs/current/spec.html#json_encoding
+    """
+
+    if isinstance(schema, avro.schema.RecordSchema) and isinstance(value, dict):
+        result = dict(value)
+        for field in schema.fields:
+            if field.name in value:
+                result[field.name] = flatten_unions(field.type, value[field.name])
+        return result
+
+    if isinstance(schema, avro.schema.UnionSchema) and isinstance(value, dict):
+        f = next((s for s in schema.schemas if s.name in value), None)
+        if f is not None:
+            # Note: This is intentionally skipping the dictionary, here the JSON representation
+            # is flattened to the Python representation
+            return flatten_unions(f, value[f.name])
+
+    if isinstance(schema, avro.schema.ArraySchema) and isinstance(value, list):
+        return [flatten_unions(schema.items, v) for v in value]
+
+    if isinstance(schema, avro.schema.MapSchema) and isinstance(value, dict):
+        return {k: flatten_unions(schema.values, v) for (k, v) in value.items()}
+
+    return value
+
+
 def read_value(schema: TypedSchema, bio: io.BytesIO):
     if schema.schema_type is SchemaType.AVRO:
         reader = DatumReader(schema.schema)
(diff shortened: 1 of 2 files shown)


tests/unit/test_serialization.py
```diff
@@ -45,6 +48,121 @@ async def test_happy_flow(default_config_path, mock_registry_client):
...
+                            "type": ["null", "string"],
+                        }
+                    ],
+                },
+            }
+        ),
+    )
+    record = {"foo": {"attr1": {"string": "sample data"}}}
+    flatten_record = {"foo": {"attr1": "sample data"}}
+    assert flatten_unions(typed_schema.schema, record) == flatten_record
```
juha-aiven, line 126: The `None` case could be tested too.

tests/unit/test_serialization.py
```diff
@@ -45,6 +48,121 @@ async def test_happy_flow(default_config_path, mock_registry_client):
...
+                            "type": ["null", "string"],
+                        }
+                    ],
+                },
+            }
+        ),
+    )
+    record = [{"attr": {"string": "sample data"}}]
+    flatten_record = [{"attr": "sample data"}]
+    assert flatten_unions(typed_schema.schema, record) == flatten_record
```
juha-aiven, line 100: The None case could be tested too.

tests/unit/test_serialization.py
```diff
@@ -45,6 +48,121 @@ async def test_happy_flow(default_config_path, mock_registry_client):
         assert 1 in o.ids_to_schemas
 
 
+def test_avro_json_flatten_unions():
```
juha-aiven, line 51: Maybe split into multiple tests - one for `record`, one for `list` etc. Now there are two tests for `array` and a test for `map` between those.

tests/unit/test_serialization.py
```diff
@@ -45,6 +48,121 @@ async def test_happy_flow(default_config_path, mock_registry_client):
...
+    typed_schema = TypedSchema.parse(
+        SchemaType.AVRO,
+        json.dumps({"type": "array", "items": ["null", "string", "int"]}),
+    )
+    record = [{"string": "foo"}, None, {"int": 1}]
+    flatten_record = ["foo", None, 1]
+    assert flatten_unions(typed_schema.schema, record) == flatten_record
+
+
+def assert_avro_json_write_invalid(schema, records):
```
juha-aiven, line 137: This is only called in `def test_avro_json_write_invalid()`? Embed inside that?

juha-aiven (changes requested):
Left some comments. 

I do understand what this PR fixes, but it's a little difficult to say whether this PR causes compatibility issues. In practise, can there be a client sending malformed data that now is accepted, but won't be accepted after this PR?

Did I get it right, that the opposite is fine and fixes an issue - that currently correct data is not accepted now but will be accepted after this PR? How do things break exactly now? Could we add a test illustrating that and show that this PR fixes the issue? Or does the new test implicitly already demonstrate that?

karapace/serialization.py
```diff
@@ -183,6 +184,63 @@ async def get_schema_for_id(self, schema_id: int) -> TypedSchema:
...
+
+    - Translates the first to the second when necessary, this adds compatibility for libraries that
+      perform the _correct_ encoding.
+    - Does nothing if the provided data is already in the second format. The data is improperly
+      encoded, but this maintains backwards compatibility.
+
+    See also https://avro.apache.org/docs/current/spec.html#json_encoding
+    """
+
+    def _flatten_unions(ss: avro.schema.Schema, vv: Any) -> Any:
```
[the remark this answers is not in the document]
hackaugusto: It is an artifact of a refactor. The inner function used to be a closure. Will change it.

tests/unit/test_serialization.py
hackaugusto: what do you mean?

hackaugusto: ping

hackaugusto: ping

jjaakola-aiven: Probably the case where `attr=None`.

hackaugusto: That is tested two lines down

HelenMel (approved):
LGTM. Intention is clear, Documentation is in place. Covered with unittests

HelenMel (approved):

hackaugusto:
The comment above was indeed an issue. Updated the PR with a unit test for it and fixed the problem. Marking the PR as ready for review again.

tests/unit/test_serialization.py
hackaugusto: This hasn't receive a reply, marking as done.

karapace/serialization.py
```diff
@@ -207,8 +262,15 @@ def read_value(schema: TypedSchema, bio: io.BytesIO):
 
 def write_value(schema: TypedSchema, bio: io.BytesIO, value: dict) -> None:
     if schema.schema_type is SchemaType.AVRO:
+
+        # Backwards compatibility: Support JSON encoded data without the tags for unions.
+        if avro.io.Validate(schema.schema, value):
+            data = value
+        else:
+            data = flatten_unions(schema.schema, value)
```
hackaugusto, line 270: This is now more forgiving than it needs to be, because the old format is handled in the branch above. A valid approach here would be to reuse the old version of flatten that has strict validation

https://github.com/aiven/karapace/pull/285

tvainika (approved):
LGTM

PR #325, opened 2022-02-04
bugfixes: limit number of instantiated producers

hackaugusto:
- The previous code would reuse producers, but would instantiate an
  unbound number of producers. Under heavy load that caused the process
  to use excessive amount of memory.
- close_producers and produce_message were not synchronized, so it was
  possible for a producer to be closed while a message was being sent,
  which _could_ have unexpected behavior (e.g. the message is produced to
  kafka but an exception is raised in the request handler, giving a
  false error to the caller)

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -36,13 +36,9 @@ class FormatError(Exception):
 
 
 class KafkaRest(KarapaceBase):
-    # pylint: disable=attribute-defined-outside-init
     def __init__(self, config: dict) -> None:
         super().__init__(config=config)
         self._add_kafka_rest_routes()
-        self._init_kafka_rest(config=config)
-
-    def _init_kafka_rest(self, config: dict) -> None:
         self.serializer = SchemaRegistrySerializer(config=config)
         self.log = logging.getLogger("KarapaceRest")
         self._cluster_metadata = None
@@ -54,8 +50,15 @@ class KafkaRest(KarapaceBase):
         self.schemas_cache = {}
         self.consumer_manager = ConsumerManager(config=config)
         self.init_admin_client()
-        self.producer_refs = []
-        self.producer_queue = asyncio.Queue()
+
+        # Lazilly initialized producer queue. The lock is not used to protected the internals of
+        # Queue, but to prevent races on attribute self._producer_queue, so every use of that
+        # attribute must be inside a critical zone of the lock.
+        self._producer_queue_lock = asyncio.Lock()
+        self._producer_queue: Optional[asyncio.Queue] = None
+
+        # The test fixtures may close the server multiple times, should be idempotent
+        self._closed = False
 
     def _add_kafka_rest_routes(self) -> None:
         # Brokers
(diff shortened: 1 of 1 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -163,16 +167,52 @@ def _add_kafka_rest_routes(self) -> None:
...
-        return await self.producer_queue.get()
+    async def maybe_initilize_producer_queue(self) -> asyncio.Queue:
+        """Lazily initalize Producers."""
+        async with self._producer_queue_lock:
+            if self._producer_queue is None:
+                producer_queue = asyncio.Queue()
+                for _ in range(self.config["producer_count"]):
+                    try:
+                        producer = await self._create_async_producer()
+                    except:
```
hackaugusto, line 178: - even though `_create_async_producer` technically never throws, it is possible for a `KeyboardException` to be raised here, so exceptions should be caught for cleanup.

hackaugusto:
Notes:
- This could be implemented with eager initialization instead, the code would be simpler, I opted to do lazy initialization because that is how the code behaved previously
- Lazy initialization could be done per item, instead of bulk. I did a first prototype using semaphore. The downside was just the additional code.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -163,16 +167,52 @@ def _add_kafka_rest_routes(self) -> None:
...
+
+        producer_queue = await self.maybe_initilize_producer_queue()
+
+        self.log.info("Disposing of async producers")
+        for _ in range(self.config["producer_count"]):
+            producer = await producer_queue.get()
+            await producer.stop()
 
     async def _create_async_producer(self) -> AIOKafkaProducer:
+        self.log.info("Creating async producers")
```
jjaakola-aiven, line 215:
```suggestion
        self.log.info("Creating an async producer")
```

jjaakola-aiven (approved):

karapace/kafka_rest_apis/__init__.py
ivanyu: Usually it's recommended to avoid articles (and some other proper grammar) in log lines...

hackaugusto: So what should be done here? Just remove the plurtal and have `Create async producer`?

jjaakola-aiven: Remove the plural as this creates single producer.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -163,16 +167,52 @@ def _add_kafka_rest_routes(self) -> None:
...
+        """Lazily initialize Producers."""
+        async with self._producer_queue_lock:
+            if self._producer_queue is None:
+                producer_queue = asyncio.Queue()
+                for _ in range(self.config["producer_count"]):
+                    try:
+                        producer = await self._create_async_producer()
+                    except:
+                        # Try to behave nicely on error, this is on a best-effort basis
+                        while not producer_queue.empty():
```
HelenMel, line 180: [Not Merge Blocker] I would log such a behavior and error if possible

karapace/kafka_rest_apis/__init__.py
```diff
@@ -163,16 +167,52 @@ def _add_kafka_rest_routes(self) -> None:
...
+        Note:
+            This method blocks until every producer can be closed
+        """
+        # Avoid initializing the producers on shutdown
+        async with self._producer_queue_lock:
+            if self._producer_queue is None:
+                self._producer_queue = asyncio.Queue()
+            return
+
+        producer_queue = await self.maybe_initilize_producer_queue()
```
HelenMel, line 207: Sorry, I don't get it, why do we have to call "maybe_initialize" when we close producers?

HelenMel:
Please, do not treat my comments as merge blockers

karapace/kafka_rest_apis/__init__.py
hackaugusto: Instead of calling the method `get_producer_queue` I named it to `maybe_initialize_producer_queue` to be explicit about the lazy behavior. The point was exactly to note that the getter has additional side effects.

With that said, I did add code to handle the special case were the queue is uninitialized. If you want I can replace this with a `producer_queue = self._producer_queue` within the critical section above.

karapace/kafka_rest_apis/__init__.py
hackaugusto: done

jjaakola-aiven (dismissed):

tvainika (commented):
In my opinion replacing `get_producer` returning a producer with `maybe_initialize_producer_queue` returning a queue changes the abstractions and responsibilities to worse. It would be better to keep such initialization process and implementation details inside `get_producer`. Also creating producer one by one can be simply done (with the lock already added here) by just comparing length of `producer_refs` to `config["producer_count"]`. That way initialization of queue can be kept inside constructor, which simplifies things overall, and not all changes here are needed anymore.

However I see this was already approved before minor logging related changes, so feel free to go ahead if think the currently proposed changes are best way forward.

hackaugusto:
closing this in favor of https://github.com/aiven/karapace/pull/328

PR #326, opened 2022-02-04
Add namespace patching if given Avro schema has invalid name for record type

jjaakola-aiven:
Some generated Avro schemas have name `record` which is not allowed in
Python Avro parser. Patch this by adding a namespace to the root if it does not exist. This
creates enclosing namespace also for nested elements.

# Why this way
Java Avro parser accepts the `record` as a name. The specification only mentions that primitive types are not allowed to be used as name. Python Avro parser has stricter validation and rejects if `record` or any other complex type name is used.

As I know only of Flink generating such schemas and the issue is only on Avro type name `record` this PR does not try to mitigate any other complex types as name.

I think the patching by adding namespace if namespace does not exist is least intrusive option.

Other options are to:
* Add the patching to the name itself, `patched.name.record`.
* Change the name by uppercasing/adding lodash the first character, `_record`, `Record`.

One issue of this change I can think of is if Avro validation is made stricter to not allow type names at all. This renders the patching with namespaces invalid.

### DIFF
--- a/karapace/avro_compatibility.py
+++ b/karapace/avro_compatibility.py
@@ -33,6 +33,33 @@ import json
 E = TypeVar("E", bound=Enum)
 
 
+def _patch_invalid_record_name_in_avro_schema(schema_json: Dict[str, str]) -> None:
+    """Patch by adding namespace to root level if Avro complex type name `record` is used as name.
+
+    Certain generated Avro schemas can contain complex Avro type names on root level as name.
+    In this case it is mandatory to patch a namespace to remove the conflict. Python Avro
+    parser does not allow to have the type name as a name.
+
+    For example Flink, at least versions from (1.13.x) with Confluent Avro value format generates
+    following schema from table SQL of "`a1` LONG NOT NULL".
+    {
+      "type":"record",
+      "name":"record",
+      "fields":
+        [{"name":"a1","type":"long"}]}
+    }
+    """
+    name: Optional[str] = schema_json.get("name", None)
+    if name == RECORD and not schema_json.get("namespace", None):
+        patch_namespace = "patched.name"
+        schema_json["namespace"] = patch_namespace
+        docstring_with_name_patch_note = f"Namespace was added to patch `{name}` to `{patch_namespace}.{name}`."
+        docstring = schema_json.get("doc")
+        if docstring:  # Keep original docstring if present, add the note.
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):
This LGTM -- nice fix!

You can add arbitrary JSON properties to an Avro schema, something like `schema_json["karapace.avro-3370-patched"] = true`.  This could help in the future if the behaviour changes... but I'm not really sure how it might be useful yet.

hackaugusto:
question: is this backwards compatible?

RyanSkraba:
> question: is this backwards compatible?

That's a good question.  This change modifies the schema being stored and returned, so it depends on how the schema is being used.  It looks promising:

Avro schema resolution [ignores namespaces](https://avro.apache.org/docs/current/spec.html#Schema+Resolution) when matching records.  If the user was using a common "reader" schema to read mismatched files, it should work transparently.  This is true if the reader schema is either the patched or unpatched schema.

Names and namespaces aren't ever present in serialized binary data, so the binary data serialized with the unpatched schema is always deserializable with the patched schema and vice versa.

However, if the user is comparing schema metadata, there could pose a problem.  For example writing Avro files from a Kafka topic using the unpatched schema, then switch to Karapace with the patched schema.  The two files will have different schema metadata, which might be unexpected.

It's tricky to know exactly what the issues might be without actually trying.  That being said, if this is a completely new job and installation, Karapace will have never accepted and served the unpatched schema.

hackaugusto (changes requested):
Blocking merge for now. Considering doing the fix in the upstream library.

RyanSkraba:
I've asked for a 1.11.1 release on Avro, we should probably be seeing release candidates with the upstream fix in the next few weeks.

kiliangrashoff:
Avro 1.11.1 has been released, it would be nice if Karapace can be upgraded to it so the issue that this PR tried to fix is resolved (commenting here since I cannot find an issue for this).

RyanSkraba:
I've raised this issue #439 -- a PR will be incoming and we can close this.

The next version of Avro will accept the `record` name (https://issues.apache.org/jira/browse/AVRO-3370) but it will also forbid invalid enum defaults (https://issues.apache.org/jira/browse/AVRO-3229)

jjaakola-aiven:
Closing this PR, the Avro 1.11.x is used in Karapace.

PR #330, opened 2022-02-07
Optimize deepcopy

hackaugusto:
While investigating performance issues with the REST Proxy API I took a profile using `py-spy`, the profiler shows the cost of deepcopy is significant for the produce endpoint. This PR replaces the use of `copy.deepcopy` with a faster version based on `ujson`, at the same time replaces all usages of `json` with `ujson`.

Using stdlib's deepcopy:

profile

Using the ujson deepcopy:

profile2

The commit also contains a small timeit script to compare the performance of the two implementations.

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -8,10 +8,10 @@ from enum import Enum, unique
 from pathlib import Path
 from typing import Dict, IO, List, Optional, Union
 
-import json
 import os
 import socket
 import ssl
+import ujson
 
 Config = Dict[str, Union[None, str, int, bool, List[str]]]
 
@@ -112,13 +112,13 @@ def set_config_defaults(config: Config) -> Config:
 
 
 def write_config(config_path: Path, custom_values: Config) -> None:
-    config_path.write_text(json.dumps(custom_values))
+    config_path.write_text(ujson.dumps(custom_values))
 
 
 def read_config(config_handler: IO) -> Config:
     try:
-        config = json.load(config_handler)
-    except json.JSONDecodeError as ex:
+        config = ujson.load(config_handler)
+    except ValueError as ex:
         raise InvalidConfiguration("Configuration is not a valid JSON") from ex
 
     return set_config_defaults(config)
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -12,12 +12,11 @@ from karapace.karapace import KarapaceBase
 from karapace.rapu import HTTPRequest
 from karapace.schema_reader import SchemaType
 from karapace.serialization import InvalidMessageSchema, InvalidPayload, SchemaRegistrySerializer, SchemaRetrievalError
-from karapace.utils import convert_to_int, KarapaceKafkaClient
+from karapace.utils import convert_to_int, deepcopy, KarapaceKafkaClient
 from typing import List, Optional, Tuple
 
 import asyncio
 import base64
-import copy
 import logging
 import time
 import ujson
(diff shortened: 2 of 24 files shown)


karapace/utils.py
```diff
@@ -20,11 +20,24 @@
...
 
 log = logging.getLogger("KarapaceUtils")
 NS_BLACKOUT_DURATION_SECONDS = 120
 
 
+def deepcopy(obj: Any) -> Any:
+    """Replacement for the standard library deepcopy.
+
+    The stdlib's copy.deepcopy is notoriously slow. After benchmarking the REST Proxy produce API
+    the deepcopy was identified as a bootleneck. Currenlty it is faster to encode the objects as
```
mble, line 33: Small typo
```suggestion
    the deepcopy was identified as a bottleneck. Currently it is faster to encode the objects as
```

karapace/kafka_rest_apis/__init__.py
```diff
@@ -292,7 +291,7 @@ def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
                         content_type="application/json",
                         status=HTTPStatus.INTERNAL_SERVER_ERROR,
                     )
-        return copy.deepcopy(self._cluster_metadata)
+        return deepcopy(self._cluster_metadata)
```
tvainika, line 294: Does this need deepcopy at all? For me looks like `self._cluster_metadata` is assigned in this method with new value, and then returned.

tvainika (commented):
Good catch & explanation.

I'd rather see `import ujson as json` to avoid gazillion of lines changed just for `s/json/ujson/`.

Also what comes to usage of deepcopy, could it be optimized even more, see inline comment? Other deepcopy changes are in the test code, so not relevant to benchmarking.

hackaugusto:
> I'd rather see import ujson as json to avoid gazillion of lines changed just for s/json/ujson/.

Their APIs are similar but not a drop in replacement, for instance `ujson.dump` doesn't accept `fp` as a keyword argument, but only a position argument. Using `ujson` is more explicit this.

> Other deepcopy changes are in the test code, so not relevant to benchmarking.

True, I changed for consistency.

karapace/config.py
```diff
@@ -112,13 +112,13 @@ def set_config_defaults(config: Config) -> Config:
...
-    config_path.write_text(json.dumps(custom_values))
+    config_path.write_text(ujson.dumps(custom_values))
 
 
 def read_config(config_handler: IO) -> Config:
     try:
-        config = json.load(config_handler)
-    except json.JSONDecodeError as ex:
+        config = ujson.load(config_handler)
+    except ValueError as ex:
```
jjaakola-aiven, line 121: Error class will get corrected in the next release of ujson.
https://github.com/ultrajson/ultrajson/issues/497

karapace/kafka_rest_apis/__init__.py
jjaakola-aiven: Cluster metadata gets mutated, so copy is needed. For example here: https://github.com/aiven/karapace/blob/2578805875d115146675d79c50df9501c3cc5edd/karapace/kafka_rest_apis/__init__.py#L676

I think this the copy is ok here. Alternative would be to create the copies where mutation occurs.

jjaakola-aiven (approved):

hackaugusto:
@jjaakola-aiven , the PR had conflicts and needed a rebase. A new review is needed.

jjaakola-aiven (changes requested):
The `karapace/avro_compatibility` uses `json` and not `ujson`.

hackaugusto:
> The karapace/avro_compatibility uses json and not ujson.

Yes, because ujson can not be used there, at least I didn't find in the API the position of the parser error to implement the same behavior.

jjaakola-aiven (approved):

karapace/kafka_rest_apis/__init__.py
tvainika: Seems we're back here with #366

hackaugusto: I didn't review the code to check how the dict is used, is a shallow copy sufficient?

PR #333, opened 2022-02-10
Bump Avro library to 1.11.0

jjaakola-aiven:
Bump Python Avro dependency to upstream 1.11.0 from forked and patched version.
Changes in this PR mainly consist of using 1.11.0 Avro API's.
 
Backwards compatibility study:
* Enum type defaults are not validated in 1.11.0, same behavior as with 1.10.x. This will be validated in version 1.11.1 and resulting exception will be `avro.errors.InvalidDefault: Enum default 'B' is not a valid member of symbols '['A']`. A test is added for this in. This test will fail with 1.11.1.
* With default values on fields accepting null the validation is similar with 1.11.0 as with 1.10.x. Following schema is parsed fine: `{"type": "record", "name": "test_record", "fields": [{"type": ["null", "int"], "name": "test_default", "default": "null"}]}` See that value `"null"` does not conform with the field type `null`.
* This version bump PR still uses the schema compatibility check module from Karapace. Move to Avro provided schema compatiblity checking is in #334.

My current knowledge is that version 1.11.0 does not cause backwards compatibility issues. The next Avro release 1.11.1 does cause.

### DIFF
--- a/karapace/avro_compatibility.py
+++ b/karapace/avro_compatibility.py
@@ -1,28 +1,14 @@
+from avro.compatibility import SchemaType
 from avro.schema import (
-    ARRAY,
     ArraySchema,
-    BOOLEAN,
-    BYTES,
-    DOUBLE,
-    ENUM,
     EnumSchema,
     Field,
-    FIXED,
     FixedSchema,
-    FLOAT,
-    INT,
-    LONG,
-    MAP,
     MapSchema,
     NamedSchema,
-    Names,
-    NULL,
-    RECORD,
+    parse,
     RecordSchema,
     Schema,
-    SchemaFromJSONData,
-    STRING,
-    UNION,
     UnionSchema,
 )
 from enum import Enum, unique
@@ -49,8 +35,7 @@ def parse_avro_schema_definition(s: str) -> Schema:
 
         json_data = json.loads(s[: e.pos])
 
-    names = Names()
-    return SchemaFromJSONData(json_data, names)
+    return parse(json.dumps(json_data))
 
 
 def is_incompatible(result: "SchemaCompatibilityResult") -> bool:
@@ -212,27 +197,36 @@ class ReaderWriterCompatibilityChecker:
         assert writer is not None
         result = SchemaCompatibilityResult.compatible()
         if reader.type == writer.type:
-            if reader.type in {NULL, BOOLEAN, INT, LONG, FLOAT, DOUBLE, BYTES, STRING}:
+            if reader.type in {
+                SchemaType.NULL,
+                SchemaType.BOOLEAN,
+                SchemaType.INT,
+                SchemaType.LONG,
+                SchemaType.FLOAT,
+                SchemaType.DOUBLE,
+                SchemaType.BYTES,
+                SchemaType.STRING,
+            }:
                 return result
-            if reader.type == ARRAY:
+            if reader.type == SchemaType.ARRAY:
                 reader, writer = cast(ArraySchema, reader), cast(ArraySchema, writer)
                 return result.merged_with(self.get_compatibility(reader.items, writer.items, "items", location))
-            if reader.type == MAP:
+            if reader.type == SchemaType.MAP:
                 reader, writer = cast(MapSchema, reader), cast(MapSchema, writer)
                 return result.merged_with(self.get_compatibility(reader.values, writer.values, "values", location))
-            if reader.type == FIXED:
+            if reader.type == SchemaType.FIXED:
                 reader, writer = cast(FixedSchema, reader), cast(FixedSchema, writer)
                 result = result.merged_with(self.check_schema_names(reader, writer, location))
                 return result.merged_with(self.check_fixed_size(reader, writer, location))
-            if reader.type == ENUM:
+            if reader.type == SchemaType.ENUM:
                 reader, writer = cast(EnumSchema, reader), cast(EnumSchema, writer)
                 result = result.merged_with(self.check_schema_names(reader, writer, location))
                 return result.merged_with(self.check_reader_enum_contains_writer_enum(reader, writer, location))
-            if reader.type == RECORD:
+            if reader.type == SchemaType.RECORD:
                 reader, writer = cast(RecordSchema, reader), cast(RecordSchema, writer)
                 result = result.merged_with(self.check_schema_names(reader, writer, location))
                 return result.merged_with(self.check_reader_writer_record_fields(reader, writer, location))
-            if reader.type == UNION:
+            if reader.type == SchemaType.UNION:
                 reader, writer = cast(UnionSchema, reader), cast(UnionSchema, writer)
                 for i, writer_branch in enumerate(writer.schemas):
                     location.append(f"{i}")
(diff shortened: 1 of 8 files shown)


tests/integration/test_schema.py
```diff
@@ -2316,21 +2322,26 @@ async def test_invalid_schema_should_provide_good_error_messages(registry_async_
...
-    # Unfortunately the AVRO library doesn't provide a good error message, it just raises an TypeError
     schema_str = jsonlib.dumps({"type": "enum", "name": "error"})
     res = await registry_async_client.post(
         f"subjects/{test_subject}/versions",
         json={"schema": schema_str},
     )
-    assert res.json()["message"] == "Invalid AVRO schema. Error: Provided schema is not valid"
+    assert (
+        res.json()["message"]
+        == "Invalid AVRO schema. Error: Enum symbols must be a sequence of strings, but it is <class 'NoneType'>"
```
tvainika, line 2332: This is not understandable error message if received by Schema Registry users

tests/integration/test_schema.py
```diff
@@ -2316,21 +2322,26 @@ async def test_invalid_schema_should_provide_good_error_messages(registry_async_
...
     # This is an upstream bug in the python AVRO library, until the bug is fixed we should at least have a nice error message
     schema_str = jsonlib.dumps({"type": "enum", "name": "error", "symbols": {}})
     res = await registry_async_client.post(
         f"subjects/{test_subject}/versions",
         json={"schema": schema_str},
     )
-    assert res.json()["message"] == "Invalid AVRO schema. Error: error is a reserved type name."
+    assert (
+        res.json()["message"]
+        == "Invalid AVRO schema. Error: Enum symbols must be a sequence of strings, but it is <class 'dict'>"
```
tvainika, line 2343: Same here. This is not understandable error message if received by Schema Registry users

tvainika (changes requested):
Those error messages need some more polishing to be understandable.

Also regarding backward compatibility, didn't we have any specific Avro schemas we suspected might be breaking now? I think tests for handling those should be added. It is fine to reject registering new erroneous schemas, but if such schemas have been already stored in `_schemas`  topic, then upgrade Karapace must be able to handle those.

tests/integration/test_schema.py
RyanSkraba: This looks like this is something that should be improved upstream in the Avro Python SDK.  I can raise a JIRA, but I would suggest either (1) rewriting all Avro error messages systematically in karapace if we want them to be consistent, or (2) passing them on as-is if we want them to be "searchable" in a search engine.

There's currently a proposal to improve and unify some error messages across the different language SDKs, but it wouldn't be ready in the next minor release.

RyanSkraba (dismissed):
Thanks for this work -- this **LGTM** depending on the decision you want to make about sending back error messages directly to the user.

tests/integration/test_schema.py
hackaugusto: +1 for improving this upstream. The message is not drastically worse and it is probably fine to wait for the next release.

RyanSkraba: Just for info -- [this is the proposition](https://github.com/apache/avro/pull/1526#issuecomment-1039668331) for some error messages for the Java SDK.  I'm hoping that the syntax is adopted across language SDKs.

jjaakola-aiven: I am for the upstream change just because of the option 2 has the searchability.

karapace/avro_compatibility.py
```diff
@@ -359,13 +353,15 @@ def check_reader_writer_record_fields(
     ) -> SchemaCompatibilityResult:
         result = SchemaCompatibilityResult.compatible()
         location.append("fields")
+        field_index = 0
         for reader_field in reader.fields:
             reader_field = cast(Field, reader_field)
-            location.append(f"{reader_field.index}")
+            location.append(f"{field_index}")
+            field_index += 1
```
hackaugusto, line 360: nit: use `enumerate` instead.

karapace/serialization.py
```diff
@@ -207,7 +207,8 @@ def read_value(schema: TypedSchema, bio: io.BytesIO):
 
 def write_value(schema: TypedSchema, bio: io.BytesIO, value: dict) -> None:
     if schema.schema_type is SchemaType.AVRO:
-        writer = DatumWriter(schema.schema)
+        # Expect that schema is the same as writer used, this is the default.
```
hackaugusto, line 210: I don't understand this comment. This is the writer, no? So this is the schema the writer is using.

tests/integration/test_schema.py
```diff
@@ -1863,7 +1863,13 @@ async def test_invalid_namespace(registry_async_client: Client) -> None:
...
     schema = {"type": "record", "name": "foo", "namespace": "foo-bar-baz", "fields": []}
     res = await registry_async_client.post(f"subjects/{subject}/versions", json={"schema": jsonlib.dumps(schema)})
-    assert res.ok, res.json()
+    assert res.status == 422, res.json()
+    json_res = res.json()
+    assert json_res["error_code"] == 44201, json_res
+    assert (
+        json_res["message"] == "Invalid AVRO schema. "
+        "Error: foo-bar-baz is not a valid Avro name because it does not match the pattern (?:^|\\.)[A-Za-z_][A-Za-z0-9_]*$"
+    ), json_res
```
hackaugusto, line 1872:
```suggestion
    expected_message = (
        "Invalid AVRO schema. Error: foo-bar-baz is not a valid Avro name because it does not match the pattern (?:^|\\.)[A-Za-z_][A-Za-z0-9_]*$"
    )
    assert json_res["message"] == expected_message, json_res
```

tests/unit/test_avro_compatibility.py
```diff
@@ -442,20 +454,22 @@ def test_union_to_simple_comparison(field):
...
-STRING_UNION_SCHEMA = UnionSchema([STRING_SCHEMA])
-BYTES_UNION_SCHEMA = UnionSchema([BYTES_SCHEMA])
-INT_STRING_UNION_SCHEMA = UnionSchema([INT_SCHEMA, STRING_SCHEMA])
-STRING_INT_UNION_SCHEMA = UnionSchema([STRING_SCHEMA, INT_SCHEMA])
-INT_FLOAT_UNION_SCHEMA = UnionSchema([INT_SCHEMA, FLOAT_SCHEMA])
-INT_LONG_UNION_SCHEMA = UnionSchema([INT_SCHEMA, LONG_SCHEMA])
-INT_LONG_FLOAT_DOUBLE_UNION_SCHEMA = UnionSchema([INT_SCHEMA, LONG_SCHEMA, FLOAT_SCHEMA, DOUBLE_SCHEMA])
-NULL_INT_ARRAY_UNION_SCHEMA = UnionSchema([NULL_SCHEMA, INT_ARRAY_SCHEMA])
-NULL_INT_MAP_UNION_SCHEMA = UnionSchema([NULL_SCHEMA, INT_MAP_SCHEMA])
+NULL_UNION_SCHEMA = UnionSchema([NULL_SCHEMA.to_json()], Names())
```
hackaugusto, line 457: Question: why is the `to_json` needed here?

tests/unit/test_serialization.py
```diff
@@ -59,6 +59,7 @@ async def test_deserialization_fails(default_config_path, mock_registry_client):
         await deserializer.deserialize(invalid_header_payload)
 
     # for now we ignore the packed in schema id
+    # This raises UnicodeDecodeError in avro.io.read_utf8
```
hackaugusto, line 62: Is this a left over comment? the test is not caughting this exception

tests/unit/test_serialization.py
```diff
@@ -69,11 +70,21 @@ async def test_deserialization_fails(default_config_path, mock_registry_client):
     schema["name"] = "BadUser"
     schema["fields"][0]["type"] = "int"
     obj = {"name": 100, "favorite_number": 2, "favorite_color": "bar"}
-    writer = avro.io.DatumWriter(avro.io.schema.parse(json.dumps(schema)))
+    writer = avro.io.DatumWriter(avro.schema.parse(json.dumps(schema)))
```
hackaugusto, line 73: I wonder if there is a way of creating the schema from a dict, this seems silly at a glance.

tests/unit/test_serialization.py
```diff
@@ -69,11 +70,21 @@ async def test_deserialization_fails(default_config_path, mock_registry_client):
...
     with io.BytesIO() as bio:
         enc = avro.io.BinaryEncoder(bio)
         bio.write(struct.pack(HEADER_FORMAT, START_BYTE, 1))
         writer.write(obj, enc)
         enc_bytes = bio.getvalue()
+    # Avro 1.11.0 does not assert anymore if the bytes io read function
```
hackaugusto, line 79: is this a left over comment? there is no `raises(AssertionError)`

hackaugusto (commented):
@jjaakola-aiven could you update the PR with the investigation on the backwards compatibility? As far as I understand this release is not breaking backwards compatibility, but it may be an issue with `1.11.1`

tests/unit/test_avro_compatibility.py
jjaakola-aiven: For example the UnionSchema, the Python 3 compatibility in Avro 1.10.x has this constructor: https://github.com/apache/avro/blob/release-1.10.2/lang/py3/avro/schema.py#L820
And on Python 2.x it uses this: https://github.com/apache/avro/blob/release-1.10.2/lang/py/avro/schema.py#L769

On the 1.11.0 the Python 3 layer is removed and code follows the previous 2.x logic.

tests/unit/test_serialization.py
jjaakola-aiven: Left over, clarifications for self when testing the differences with Aiven fork and Avro versions.
Will remove.

tests/unit/test_serialization.py
jjaakola-aiven: Sure, `avro.schema.make_avsc_object`.

tests/unit/test_serialization.py
jjaakola-aiven: No, there is no more assertion errors. This comment server a purpose of explaining what is the behavior.
The test is for this line:
https://github.com/aiven/karapace/pull/333/files#diff-af2237d5ef00ea62612450b70e195a61f09b19b9aa0e224795942b2d3418a646R258

karapace/avro_compatibility.py
jjaakola-aiven: Done.

tests/integration/test_schema.py
jjaakola-aiven: Done.

tests/unit/test_avro_compatibility.py
hackaugusto: ugh, `to_json` doesn't return json, it returns the corresponding python object.

tests/integration/test_schema.py
tvainika: Ok, we can go with this, and wait for the upstream improving the error messages. At least here the error message is also improvement over old one unlike with reserved keyword (`"error"` being reserved word) case later.

tvainika (changes requested):
Agreed about error messages to go with new upstream version later, not stuck with that.

Now has conflicts to be resolved, though.

tvainika (approved):

PR #340, opened 2022-02-11
ci: saving invidual log files

hackaugusto:
- Kafka logs no longer printed to stdout
- ci: The log files from pytest and kafka are saved in artifacts for inspection
- fixed a bug in rapu, 500 internal errors did not set the response content type
- removed duplicated code from the Makefile (relying exclusively on `pre-commit`)
- conftest automatically downloading Kafka to the `runtime` directory
- ground work to run test suite with different kafka versions


.github/workflows/tests.yml
```diff
@@ -39,3 +39,15 @@ jobs:
...
+      uses: actions/upload-artifact@v2
+      with:
+          name: logs
+          # This directory is created by the fixture session_tmppath. The path is composed of a
+          # prefix defined by --basepath, followed by a folder name, in this case it is "karapace".
+          #
+          # See:
+          # - https://docs.pytest.org/en/6.2.x/tmpdir.html#base-temporary-directory
+          # - fixture session_tmppath in tests/integration/conftest.py
+          path: /tmp/pytest*
```
hackaugusto, line 53: I have tried:

- `/tmp`: didn't work because not all files have read permission
- `/tmp/pytest-of-*/karapace`: not sure why it did not work
- `/tmp/pytest*`

hackaugusto: The files are inside `/tmp/pytest-of-runner/pytest-<number>/popen-<worker>/karapacecurrent`

hackaugusto: These directories contained the Kafka data files, which are pre allocated files. So the artifact was around 10GB of data. This PR made the directory for log files configurable and separate from the data directory

.github/workflows/tests.yml
```diff
@@ -34,7 +31,13 @@ jobs:
          version: '3.13.0'
 
     - name: Execute unit-tests
-      run: make unittest
+      run: python3 -m pytest -s -vvv tests/unit/
 
     - name: Execute integration-tests
-      run: make integrationtest
+      run: python3 -m pytest -s -vvv tests/integration/ --log-dir=/tmp/ci-logs
```
hackaugusto, line 37: it is more straight forward to configure the tests with flags for pytest instead of env variables for the Makefile

.gitignore
```diff
@@ -12,8 +12,5 @@ __pycache__/
 /dist/
 /karapace.egg-info/
 /karapace-rpm-src.tar
-/rpm/
-/kafka_*.tgz
```
hackaugusto, line 16: using `runtime` since that is also used by the protobuf tests

.gitignore
```diff
@@ -12,8 +12,5 @@ __pycache__/
 /dist/
 /karapace.egg-info/
 /karapace-rpm-src.tar
-/rpm/
```
hackaugusto, line 15: this was a left over, should have been removed a while ago

Makefile
```diff
@@ -1,99 +1,19 @@
...
-KAFKA_TAR = $(KAFKA_PATH).tgz
-PYTHON_SOURCE_DIRS = karapace/
-PYTHON_TEST_DIRS = tests/
-ALL_PYTHON_DIRS = $(PYTHON_SOURCE_DIRS) $(PYTHON_TEST_DIRS)
-GENERATED = karapace/version.py
 PYTHON = python3
-DNF_INSTALL = sudo dnf install -y
 
-# Keep these is sync with tests/integration/conftest.py
-KAFKA_VERSION=2.7.0
```
hackaugusto, line 13: now this is only in the pytest file, easier to maintain

Makefile
```diff
@@ -1,99 +1,19 @@
...
-.PHONY: isort
-isort:
-	pre-commit run isort --all-files
-
-.PHONY: reformat
-reformat: isort black
-
-.PHONY: pre-commit
-pre-commit: $(GENERATED)
-	pre-commit run --all-files
```
hackaugusto, line 96: the contributing recommends installing the hooks. deleting from here to remove duplicated code

tests/conftest.py
```diff
@@ -50,6 +53,7 @@ def split_by_comma(arg: str) -> List[str]:
 
 def pytest_addoption(parser, pluginmanager) -> None:  # pylint: disable=unused-argument
     parser.addoption("--kafka-bootstrap-servers", type=split_by_comma)
+    parser.addoption("--kafka-version", default=KAFKA_VERSION)
```
hackaugusto, line 56: this is just preparing the way to run the tests against multiple Kafka versions

tests/conftest.py
```diff
@@ -69,18 +75,41 @@ def fixture_validate_options(request) -> None:
...
+@pytest.fixture(scope="session", name="session_logdir")
+def fixture_session_logdir(request, tmp_path_factory) -> Path:
+    """All useful log data for debugging should be stored here.
 
-@pytest.fixture(scope="session", name="session_tmppath")
-def fixture_session_tmppath(tmp_path_factory) -> Path:
-    return tmp_path_factory.mktemp("karapace")
+    These files are persisted by the CI for debugging purposes.
+    """
+    log_dir = request.config.getoption("log_dir")
```
hackaugusto, line 102: locally we can just have the data and log files in the same directories, but for the CI the logs and Kafka's data must be in separate folders, otherwise the artifacts will be a couple of GB in size (Kafka pre allocate the files)

tests/integration/config/log4j.properties
```diff
@@ -0,0 +1,71 @@
+# Unspecified loggers and loggers with additivity=true output to server.log
```
hackaugusto, line 1: This is the default logging configuration, the only change is that stdout is no longer used.

karapace/karapace.py
```diff
@@ -7,28 +7,19 @@
...
+        super().__init__(app_name="karapace", config=config)
 
         self.kafka_timeout = 10
-        self.config = config
-        self._sentry_config = sentry_config
-        if os.environ.get("SENTRY_DSN"):
-            self._sentry_config["dsn"] = os.environ["SENTRY_DSN"]
-        if "tags" not in self._sentry_config:
-            self._sentry_config["tags"] = {}
-        self._sentry_config["tags"]["app"] = "Karapace"
```
hackaugusto, line 30: This was super weird, the variable `self._sentry_config` is never used, but it refers to the same object as `sentry_config`, which is forwarded to `RestApp` then `StatsClient`, where these side effects are used.

For the sake of simplicity I changed this to be set with the other configs.

karapace/rapu.py
```diff
@@ -333,9 +335,9 @@ async def _handle_request(
                 headers = ex.headers
             except:  # pylint: disable=bare-except
                 self.log.exception("Internal server error")
+                headers = {"Content-Type": "application/json"}
                 data = {"error_code": HTTPStatus.INTERNAL_SERVER_ERROR.value, "message": "Internal server error"}
                 status = HTTPStatus.INTERNAL_SERVER_ERROR
-                headers = {}
```
hackaugusto, line 338: The lack of headers IMO is a bug, and that was cluttering the failures in the tests because `requests.get().json()` throws an exception because the response was lacking the correct content type.

tests/integration/conftest.py
```diff
@@ -4,136 +4,73 @@
...
-
-CLASSPATH = os.path.join(BASEDIR, "libs", "*")
+REPOSITORY_DIR = pathlib.Path(__file__).parent.parent.parent.absolute()
+RUNTIME_DIR = REPOSITORY_DIR / "runtime"
+TEST_INTEGRATION_DIR = REPOSITORY_DIR / "tests" / "integration"
 KAFKA_WAIT_TIMEOUT = 60
-
-
-@dataclass
-class ZKConfig:
```
hackaugusto, line 47: I moved almost everything that is not a fixture into the `tests/integration/utils` module.

tests/integration/conftest.py
```diff
@@ -142,40 +79,90 @@ def fixture_kafka_server(request, session_tmppath: Path) -> Iterator[KafkaServer
...
+    zk_dir = session_logdir / "zk"
 
-    with ExitStack() as stack:
-        # Synchronize xdist workers, data generated by the winner is shared through
-        # transfer_file (primarily the server's port number)
+    # File used to share data among test runners, including the dynamic
+    # configuration for this session (mainly port numbers), and and
+    # synchronization data.
+    transfer_file = session_logdir / "transfer"
+    lock_file = lock_path_for(transfer_file)
```
hackaugusto, line 88: This is why the `session_logdir` needs to be a shared dir, so that each work uses the same file as a lock.

tests/integration/conftest.py
```diff
@@ -142,40 +79,90 @@ def fixture_kafka_server(request, session_tmppath: Path) -> Iterator[KafkaServer
...
+        with FileLock(str(lock_file)):  # pylint: disable=abstract-class-instantiated
+            assert transfer_file.exists(), "transfer_file disappeared"
+            config_data = ujson.loads(transfer_file.read_text())
+            config_data[WORKER_COUNTER_KEY] -= 1
+            transfer_file.write_text(ujson.dumps(config_data))
+
+        # Wait until every worker finished before stopping the servers
+        worker_counter = float("inf")
+        while worker_counter > 0:
+            with FileLock(str(lock_file)):  # pylint: disable=abstract-class-instantiated
```
hackaugusto, line 159: This was a bug. The "leader" node could finish running the tests before the other workers, and it would stop the server before the tests finished running.

I verified this in a very rudimentary way:

- Made sure gw1 was the leader by adding a sleep of 0.5s for gw0 before acquiring the lock
- Picked the test_union_to_union and made gw0 wait for 30s before running the test, this makes sure gw1 finish running the tests and stops kafka before gw0 starts
- The above above gives a connection error when gw0 runs the test

tests/integration/conftest.py
```diff
@@ -333,7 +323,7 @@ async def fixture_registry_async(
             # Because of this every test must be written in such a way that it can
             # be executed twice with the same servers.
             # "topic_name": new_random_name("topic"),
-            # "group_id": new_random_name("schema_registry")
+            "group_id": new_random_name("registry_async"),
```
hackaugusto, line 326: This is the bug that I started hunting down. `registry_async` and `registry_async_tls` used the same `group_id` but different configs, which caused flakiness on tests when trying to use TLS with the non TLS node.

tests/integration/utils/kafka_server.py
```diff
@@ -0,0 +1,164 @@
...
+        "delete.topic.enable": "true",
+        "inter.broker.listener.name": "PLAINTEXT",
+        "inter.broker.protocol.version": kafka_description.protocol_version,
+        "listeners": listeners,
+        "log.cleaner.enable": "true",
+        "log.dirs": kafka_config.datadir,
+        "log.message.format.version": kafka_description.protocol_version,
+        "log.retention.check.interval.ms": 300000,
+        "log.segment.bytes": 200 * 1024 * 1024,  # 200 MiB
+        "log.preallocate": False,
```
hackaugusto, line 134: I'm leaving this behind, even though it may not be necessary now that logs and data are on different directories, but I added this because on the first try everything was included and the generated artifact was a couple of GiB because the files had been preallocated.

hackaugusto:
I had this PR split into multiple ones:

https://github.com/aiven/karapace/pull/357
https://github.com/aiven/karapace/pull/359
https://github.com/aiven/karapace/pull/360
https://github.com/aiven/karapace/pull/356
https://github.com/aiven/karapace/pull/358

Just for the sake of clarify, since some of the comments I have added here are no longer relevant. I'm closing this and reopening to get a clean slate.

PR #341, opened 2022-02-11
Update Protobuf status

### LINKED ISSUE #67: Add support for ProtoBuf

tvainika:

Karapace now supports Protobuf schemas
Fixes #67


hackaugusto (approved):
LGTM

PR #342, opened 2022-02-11
deps: updated pinning based on security vulnerabilities

hackaugusto:
- add the missing protobuf dependency
- add minimum required version for dependencies with known CVEs


HelenMel (dismissed):

hackaugusto:
> Something is wrong for 3.10 @hackaugusto Is it a blocker?

It was just flakiness.

> It looks like very important change

it should be okay, by default pip installs the largest version for a dependency, so the correct version should be installed by default. This only fixes issues for virtualenvs with the dependency already installed (because pip doesn't automatically update installed dependencies unless there is a version mismatch)

tvainika (approved):

PR #344, opened 2022-02-16
Fix cluster_metadata caching and locking

tvainika:

Fix `cluster_metadata` caching

# Why this way

In case `cluster_metadata` was fetched first for a limited set of
topics, the method internally cached partial metadata, which was then
returned regardless if full metadata was requested.

Also use time.monotonic() for cache expiry so that difference is
always valid.

Adjust test configuration for ability to test caching issues.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -279,24 +279,31 @@ class KafkaRest(KarapaceBase):
             return self.admin_client.get_topic_config(topic)
 
     def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
-        if self._metadata_birth is None or time.time() - self._metadata_birth > self.metadata_max_age:
-            self._cluster_metadata = None
-        if not self._cluster_metadata:
-            self._metadata_birth = time.time()
-            with self.admin_lock:
-                try:
-                    self._cluster_metadata = self.admin_client.cluster_metadata(topics)
-                except NodeNotReadyError:
-                    self.log.exception("Could not refresh cluster metadata")
-                    self.r(
-                        body={
-                            "message": "Kafka node not ready",
-                            "code": RESTErrorCodes.HTTP_INTERNAL_SERVER_ERROR.value,
-                        },
-                        content_type="application/json",
-                        status=HTTPStatus.INTERNAL_SERVER_ERROR,
-                    )
-        return copy.deepcopy(self._cluster_metadata)
+        with self.admin_lock:
+            if self._metadata_birth is None or time.monotonic() - self._metadata_birth > self.metadata_max_age:
+                self._cluster_metadata = None
+
+            if self._cluster_metadata and topics is None:
+                return copy.deepcopy(self._cluster_metadata)
+
+            try:
+                metadata_birth = time.monotonic()
+                metadata = self.admin_client.cluster_metadata(topics)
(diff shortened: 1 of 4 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -279,24 +279,31 @@ def get_topic_config(self, topic: str) -> dict:
             return self.admin_client.get_topic_config(topic)
 
     def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
```
hackaugusto, line 281: IMO we should avoid writing code that rely on flags as much as possible (in this case the flag is `None` vs a `list`). This is a good example why, the code would be less error prone and easier to understand if written with two functions, one `cluster_metadata` and the other `topics_metadata`. This would also make it easy to use a caching decorator for this.

With that said I don't think the PR should be blocked because of this, but it would be nice to gradually introduce a safer coding style as we do new changes.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -279,24 +279,31 @@ def get_topic_config(self, topic: str) -> dict:
...
-                        },
-                        content_type="application/json",
-                        status=HTTPStatus.INTERNAL_SERVER_ERROR,
-                    )
-        return copy.deepcopy(self._cluster_metadata)
+        with self.admin_lock:
+            if self._metadata_birth is None or time.monotonic() - self._metadata_birth > self.metadata_max_age:
+                self._cluster_metadata = None
+
+            if self._cluster_metadata and topics is None:
```
hackaugusto, line 286: question: why is the cache being skipped if the argument topics is provided? Is that intentional?

karapace/kafka_rest_apis/__init__.py
```diff
@@ -279,24 +279,31 @@ def get_topic_config(self, topic: str) -> dict:
...
-                    self.r(
-                        body={
-                            "message": "Kafka node not ready",
-                            "code": RESTErrorCodes.HTTP_INTERNAL_SERVER_ERROR.value,
-                        },
-                        content_type="application/json",
-                        status=HTTPStatus.INTERNAL_SERVER_ERROR,
-                    )
-        return copy.deepcopy(self._cluster_metadata)
+        with self.admin_lock:
```
jjaakola-aiven, line 282: Aiohttp is providing the thread. in a case where full metadata is requested and processing takes some time and a subsequent request is done it is possible to have deadlock as `threading.Lock` would be already held by the same thread. I would propose to use `asyncio.Lock`, this lock must be created in the eventloop thread.

jjaakola-aiven (changes requested):

karapace/kafka_rest_apis/__init__.py
tvainika: I did it intentionally to always return only given topics. This code could be place for more refactoring, I agree, but for now I was trying to make it correct first.

karapace/kafka_rest_apis/__init__.py
tvainika: There was already `_async_provider_lock = asyncio.Lock()` so it was easy to verify it is created in the eventloop thread.

jjaakola-aiven (approved):
Lock usage is now ok.

hackaugusto:
Merging because the PR is approved. I did not review the lock changes

PR #345, opened 2022-02-23
logging: add optional aiohttp.access debug logger

mble:
`aiohttp.access` logs can be numerous should clients be constantly
talking to `karapace`, and often these access logs don't provide
significant value for operators.

In order to not throw the baby out with the bath water, this set of
changes allows the use of a logger that outputs `aiohttp.access` logs as
`DEBUG` instead of `INFO` statements. Log filtering can then be applied
at log collection to drop the `DEBUG` logging.

# Why this way
This is really the only way to modify the log level that `aiohttp.access` logs are emitted at, as `aiohttp` only lets you totally disable them otherwise. This lets us optionally keep the logs.

### DIFF
--- a/karapace.config.json
+++ b/karapace.config.json
@@ -1,4 +1,5 @@
 {
+    "access_logs_debug": false,
     "advertised_hostname": "localhost",
     "bootstrap_uri": "127.0.0.1:9092",
     "client_id": "sr-1",
--- a/karapace/kafka_rest_apis/__main__.py
+++ b/karapace/kafka_rest_apis/__main__.py
@@ -1,7 +1,9 @@
+from aiohttp.web_log import AccessLogger
 from contextlib import closing
 from karapace import version as karapace_version
 from karapace.config import DEFAULT_LOG_FORMAT_JOURNAL, read_config
 from karapace.kafka_rest_apis import KafkaRest
+from karapace.utils import DebugAccessLogger
 
 import argparse
 import logging
@@ -10,7 +12,12 @@ import sys
 
 def main() -> int:
     parser = argparse.ArgumentParser(prog="karapace rest", description="Karapace: Your Kafka essentials in one tool")
-    parser.add_argument("--version", action="version", help="show program version", version=karapace_version.__version__)
+    parser.add_argument(
+        "--version",
+        action="version",
+        help="show program version",
+        version=karapace_version.__version__,
+    )
     parser.add_argument("config_file", help="configuration file path", type=argparse.FileType())
     arg = parser.parse_args()
 
@@ -19,9 +26,14 @@ def main() -> int:
 
     logging.basicConfig(level=logging.INFO, format=DEFAULT_LOG_FORMAT_JOURNAL)
     logging.getLogger().setLevel(config["log_level"])
+    if config.get("access_logs_debug"):
+        access_logger = DebugAccessLogger
+        logging.getLogger("aiohttp.access").setLevel(logging.DEBUG)
+    else:
+        access_logger = AccessLogger
     kc = KafkaRest(config=config)
     try:
-        kc.run(host=kc.config["host"], port=kc.config["port"])
+        kc.run(host=kc.config["host"], port=kc.config["port"], access_logger=access_logger)
     except Exception:  # pylint: disable-broad-except
         if kc.raven_client:
             kc.raven_client.captureException(tags={"where": "karapace_rest"})
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -7,11 +7,12 @@ Copyright (c) 2019 Aiven Ltd
 See LICENSE for details
 """
 from accept_types import get_best_match
+from aiohttp.abc import AbstractAccessLogger
 from http import HTTPStatus
 from karapace.statsd import StatsClient
 from karapace.utils import json_encode
 from karapace.version import __version__
-from typing import Dict, NoReturn, Optional, overload, Union
+from typing import Dict, NoReturn, Optional, overload, Type, Union
 
 import aiohttp
 import aiohttp.web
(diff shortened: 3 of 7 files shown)


karapace/rapu.py
```diff
@@ -464,7 +494,7 @@ async def http_request(self, url, *, method="GET", json=None, timeout=10.0, veri
                 rdns=False,
                 verify_ssl=verify,
                 ssl_context=sslcontext,
-            )
+            )  #
```
ryanisroot, line 497: probably don't need this

ryanisroot (approved):
This looks good to me. Better to solve the issue here vs further upstream.

rikonen:
Would be nice to have aggregate information about the API access which does grouping by (method, path) and prints the counts every minute or something like that so that there's no total loss of visibility to the API requests being performed when the individual requests aren't logged. (Don't know how many unique combinations that would typically generate and if that too would end up being spammy but I'd presume most of the time there's a lot of requests for just a couple of different combinations like that.)

karapace/rapu.py
mble: Huh, how did this sneak in?

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 125
+include = '\.pyi?$'
```
mble, line 3: This is added to allow editors that support `black` for auto-formatting to align with the `pre-commit` configuration for `black`.

tvainika: I see many lines of what looks like only formatting changes, and some of those don't even hit this line length of 125. Are those all intentional, or just happened before adding this?

mble: @tvainika These are all intentional from running `make lint`.

tvainika: Just running `make lint` on current main branch does not trigger those reformattings (like with `parser.add_argument`), but those are caused by adding trailing commas. Those cause much more than necessary formatting changes here, not only running `make lint`. Not objecting having those all now, but generally I feel that formatting changes are better to separate.

karapace/utils.py
```diff
@@ -332,3 +344,33 @@ def ns_blackout(self):
 
     def blacked_out(self):
         return self.ns_blackout() or super().blacked_out()
+
+
+class DebugAccessLogger(AccessLogger):
+    """Logs access logs as DEBUG instead of INFO."""
+
+    def log(
```
hackaugusto, line 352: review note: This code is copied form the aiohttp library with the log level changed. So the coding style follows the one of aiohttp.

- ignoring broad except
- ignoring unidiomatic instance check

ref: https://github.com/aio-libs/aiohttp/blob/d01e257da9b37c35c68b3931026a2d918c271446/aiohttp/web_log.py#L191-L210

karapace.config.json
```diff
@@ -1,4 +1,5 @@
 {
+    "access_logs_debug": false,
```
hackaugusto, line 2: I'm not sure if this is the best approach. If we follow this style we would end up with a flag for each logger. It makes more sense to add a logging config dictionary:

https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig

And disable the logging for that particular logger, partial example:

```json
{
  "logging_config": {
    "version": 1,
    "loggers": {
      "aiohttp.access": {
        "level": "CRITICAL"
      }
    }
  }
}
```

That needs additional code though, to setup the handlers.

rikonen:
> @rikonen wouldn't it make more sense to add metrics for that instead of logging?

That is one option but I suggested the aggregate logging because:

1. Metrics cardinality could end up being bad for cases where there are a lot of different paths being accessed, possibly catastrophically so for some pathological use cases
2. This change will remove the logging from people who currently have it and they might not have metrics integration configured. E.g. in case of Aiven you can see logs as part of basic product feature set but metrics are not exposed directly in the UI like that

karapace.config.json
hackaugusto: Humm, I forgot about some of the rationale here. The idea was exactly to patch aiohttp to change the log level, because it is easier to configure https://github.com/aiven/journalpump to filter messages by that. I suppose the alternative would be to have a search with a negative REGEX, but that may not be the best idea. So never mind my suggestion above.

hackaugusto:
> This change will remove the logging from people who currently have it and they might not have metrics integration configured. E.g. in case of Aiven you can see logs as part of basic product feature set but metrics are not exposed directly in the UI like that

@mble correct me if I'm wrong, but that shouldn't be the case. It is only changing the log level from info to debug, so the logs are still generated.

mble:
@hackaugusto The logs would still be generated, but depends on the config of the log forwarder (if one is present) to get to remote storage.

rikonen:
I presume the whole point of this is that we would by default be suppressing the logs going forward. Which is why the aggregate stuff could be good to have to avoid backlash and us having to manually change the log level for a lot of services.

mble:
One of the challenges with aggregate stats in a log is what to aggregate on - path, user agent, status code etc. It also loses some fidelity with response sizes. I can put together an aggregate logger and see what that would look like.

rikonen:
I'd go with (method, path). That should give pretty good idea of what sort of requests are happening. There will of course be loss of information with that and I'm not setting requirements anyway so maybe we don't want it, just afraid that if we don't do anything but drop the calls from logs then we'll end up with lot of requests to get them back while the aggregates might be good enough for most who might otherwise want the old logging re-enabled.

karapace/kafka_rest_apis/__main__.py
```diff
@@ -19,9 +26,14 @@ def main() -> int:
 
     logging.basicConfig(level=logging.INFO, format=DEFAULT_LOG_FORMAT_JOURNAL)
     logging.getLogger().setLevel(config["log_level"])
+    if config.get("access_logs_debug"):
```
hackaugusto, line 29: This branch will be taken if the value of the config is truthy, including the string `"false"`, should have validation for the type been `bool` to prevent unexpected behavior

hackaugusto (changes requested):
changing the status of the PR because of @rikonen request

hackaugusto:
@mble ping

rikonen:
I'm fine merging as is (provided this is sensible otherwise, didn't really review properly) but we then need to have discussion of what makes sense before actually changing our config to make use of this.

mble:
This has diverged significantly from `master` so I'm closing this and re-doing it with the current architecture.

PR #348, opened 2022-03-09
Use configured protobuf runtime directory

tvainika:

Use configured protobuf runtime directory

# Why this way

Fix `protobuf_runtime_directory` configuration handling to pass configured value around instead of falling back to defaults.

### DIFF
--- a/karapace/protobuf/io.py
+++ b/karapace/protobuf/io.py
@@ -1,5 +1,4 @@
 from io import BytesIO
-from karapace import config
 from karapace.protobuf.encoding_variants import read_indexes, write_indexes
 from karapace.protobuf.exception import IllegalArgumentException, ProtobufSchemaResolutionException, ProtobufTypeException
 from karapace.protobuf.message_element import MessageElement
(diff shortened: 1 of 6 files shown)


karapace/protobuf/io.py
```diff
@@ -92,24 +91,26 @@ def read_data(writer_schema: ProtobufSchema, reader_schema: ProtobufSchema, bio:
...
             self._reader_schema = self._writer_schema
-        return protobuf_to_dict(read_data(self._writer_schema, self._reader_schema, bio), True)
+        return protobuf_to_dict(read_data(self.config, self._writer_schema, self._reader_schema, bio), True)
 
 
 class ProtobufDatumWriter:
     """ProtobufDatumWriter for generic python objects."""
 
-    def __init__(self, writer_schema: ProtobufSchema = None):
+    def __init__(self, config: dict, writer_schema: ProtobufSchema = None):
```
hackaugusto, line 112: nit pick: Please fix typing when touching code:

- config type should be `Config` from `karapace.config`
- `writer_schema` should have an explicit `Optional` (configuration dependent, module is probably does not have the setting set)
- initializer should have explicit return type `-> None` (good practice, not required by mypy)

```suggestion
    def __init__(
        self,
        config: Config,
        writer_schema: Optional[ProtobufSchema] = None,
    ) -> None:
```

hackaugusto (approved):
the type of `config` is `karapace.config.Config`. It also makes sense to only give the necessary config to the protobuf serializes (in this case only the runtime directory and not the whole config).

These could be done in a follow up.

PR #349, opened 2022-03-09
Fix internal server error with JSON non-object body

tvainika:

Fix internal server error with JSON non-object body

# Why this way

If HTTP request body was valid JSON, but not JSON object, Karapace
threw 500 Internal server error due assuming JSON body is always
object (dict in Python).

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -9,7 +9,7 @@ from karapace.kafka_rest_apis.admin import KafkaRestAdminClient
 from karapace.kafka_rest_apis.consumer_manager import ConsumerManager
 from karapace.kafka_rest_apis.error_codes import RESTErrorCodes
 from karapace.karapace import KarapaceBase
-from karapace.rapu import HTTPRequest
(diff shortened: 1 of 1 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -398,7 +398,11 @@ async def partition_publish(self, topic, partition_id, content_type, *, request)
 
     async def topic_publish(self, topic: str, content_type: str, *, request):
         self.log.debug("Executing topic publish on topic %s", topic)
-        await self.publish(topic, None, content_type, request.content_type, request.json)
+        body = request.json
+        if not isinstance(body, dict):
+            raise HTTPResponse(body="Request body is not JSON object", status=HTTPStatus.BAD_REQUEST)
```
hackaugusto, line 403: I think it makes sense to add validation of the request body with a schema using something like pydantic, which gives more precise errors.

hackaugusto: @tvainika could you please add an update here.

tvainika: Makes sense to use pydantic instead of this.

PR #354, opened 2022-03-15
Use only non-versioned dependencies

tvainika:

Use only non-versioned dependencies in `setup.py` in unified fashion for all dependencies, and leave version dependencies to locking file (`requirements.txt`)

References: #342

# Why this way

Use `requirements.txt` for pinning recent good versions, and remove partial CVE listings and minimum requirements lagging behind.

This also allows usage of system dependencies where `requirements.txt` dependencies is not in use, for example Fedora 34, which has `python3-protobuf v.3.14` with a patch for referred CVE.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,8 +30,8 @@ setup(
         "jsonschema",
         "kafka-python",
         "networkx",
-        "protobuf>=3.16",  # 3.15 vunerability CVE-2021-22570
-        "requests>=2.21",  # 2.20 vunerability CVE-2018-18074
+        "protobuf",
+        "requests",
     ],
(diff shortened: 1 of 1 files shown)


jlprat:
Just my opinion on this:
> for example Fedora 34, which has python3-protobuf v.3.14 with a patch for referred CVE.

This makes it more complicated to assess if our software is vulnerable or not to a given CVE.

hackaugusto:
I agree with @jlprat , IMO this modification should be done in a separate branch and fedora packages should be released from there.

hackaugusto (approved):
LGTM

PR #355, opened 2022-03-15
Fix protobuf_runtime_directory absolute directory

tvainika:

Fix protobuf_runtime_directory absolute directory

Related to #348

# Why this way

Invocation of `protoc` failed if `protobuf_runtime_directory` was configured as absolute directory instead of relative subdir.

Change current working directory for `protoc` invocation to have all relative to configured directory to get inputs and outputs there.

### DIFF
--- a/karapace/protobuf/io.py
+++ b/karapace/protobuf/io.py
@@ -51,7 +51,7 @@ def find_message_name(schema: ProtobufSchema, indexes: List[int]) -> str:
 def get_protobuf_class_instance(schema: ProtobufSchema, class_name: str, cfg: Dict) -> Any:
     directory = cfg["protobuf_runtime_directory"]
     proto_name = calculate_class_name(str(schema))
-    proto_path = f"{directory}/{proto_name}.proto"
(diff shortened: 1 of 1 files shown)


karapace/protobuf/io.py
```diff
@@ -51,7 +51,7 @@ def find_message_name(schema: ProtobufSchema, indexes: List[int]) -> str:
 def get_protobuf_class_instance(schema: ProtobufSchema, class_name: str, cfg: Dict) -> Any:
     directory = cfg["protobuf_runtime_directory"]
     proto_name = calculate_class_name(str(schema))
-    proto_path = f"{directory}/{proto_name}.proto"
+    proto_path = f"{proto_name}.proto"
     class_path = f"{directory}/{proto_name}_pb2.py"
```
hackaugusto, line 55: what about this directory? it can be absolute?

tvainika: Yes

hackaugusto (approved):

PR #359, opened 2022-03-16
Removed extra entrypoints

hackaugusto:
`karapace/karapace_all.py` is the only exposed entrypoint, and that covers the usages from here:

https://github.com/aiven/karapace/blob/f7e88cef6180391b51296ca45d6a0956d8a98561/setup.py#L45-L50


karapace/schema_registry_apis.py
```diff
@@ -907,29 +901,3 @@ def no_master_error(self, content_type):
...
-    logging.getLogger().setLevel(config["log_level"])
-    kc = KarapaceSchemaRegistry(config=config)
-    try:
-        kc.run(host=kc.config["host"], port=kc.config["port"])
-    except Exception:  # pylint: disable-broad-except
-        if kc.raven_client:
-            kc.raven_client.captureException(tags={"where": "karapace"})
-        raise
-
-    return 0
```
jlprat, line 931: If I understand the PR right, this was a secondary way to call Karapace, but we shouldn't have it as we want people to use the main one. Right?

hackaugusto: Pretty much

jlprat (approved):
LGTM as long as this is big letters in the release notes.

tvainika (changes requested):
This breaks container images as https://github.com/aiven/karapace/blob/master/container/start.sh uses those endpoints

hackaugusto:
> This breaks container images as https://github.com/aiven/karapace/blob/master/container/start.sh uses those endpoints

good caught, I forgot about that. fixing it

jlprat (approved):

juha-aiven (changes requested):
`karapace/schema_registry_apis.py` has a conflict.

tvainika (approved):
LGTM now

hackaugusto:
@jlprat here is the release: https://github.com/aiven/karapace/releases/tag/3.0.0

PR #360, opened 2022-03-16
Simplified sentry config

hackaugusto:
This unifies how the sentry configuration is set

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -95,15 +95,15 @@ def set_config_defaults(config: Config) -> Config:
             env_name = k.upper()
         else:
             env_name = f"karapace_{k}".upper()
-        if env_name in os.environ:
-            val = os.environ[env_name]
+        env_val = os.environ.get(env_name)
+        if env_val is not None:
             LOG.debug(
                 "Populating config value %r from env var %r with %r instead of config file",
                 k,
                 env_name,
-                val,
+                env_val,
             )
-            config[k] = parse_env_value(os.environ[env_name])
+            config[k] = parse_env_value(env_val)
         config.setdefault(k, v)
 
     master_election_strategy = config["master_election_strategy"]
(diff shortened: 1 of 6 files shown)


karapace/karapace.py
```diff
@@ -7,28 +7,19 @@
...
+        super().__init__(app_name="karapace", config=config)
 
         self.kafka_timeout = 10
-        self.config = config
-        self._sentry_config = sentry_config
-        if os.environ.get("SENTRY_DSN"):
-            self._sentry_config["dsn"] = os.environ["SENTRY_DSN"]
-        if "tags" not in self._sentry_config:
-            self._sentry_config["tags"] = {}
-        self._sentry_config["tags"]["app"] = "Karapace"
```
hackaugusto, line 30: This is what was hard to understand, `self._sentry_config` is never used, only modified. Turns out that `self._sentry_config` is the same object as `config["sentry"]` and that makes its way to the stats client, so that object over there was relying on this spooky action at a distance

juha-aiven:
You have a conflict in `karapace/schema_registry_apis.py`, please fix.

tvainika (changes requested):
conflicts

karapace/config.py
```diff
@@ -88,17 +90,45 @@ def parse_env_value(value: str) -> Union[str, int, bool]:
...
+    new_config = DEFAULTS.copy()
+    new_config.update(config)
 
+    set_settings_from_envionment(new_config)
+    set_sentry_dsn_from_environment(new_config)
+    validate_config(new_config)
+    return new_config
+
+
+def set_settings_from_envionment(config: Config):
```
tvainika, line 102:
```suggestion
def set_settings_from_environment(config: Config):
```

typo in function name

tvainika (dismissed):
LGTM overall, but found one typo to fix before merging

jjaakola-aiven (approved):

PR #361, opened 2022-03-16
Save logs as artifacts

hackaugusto:
- Kafka logs no longer printed to stdout
- ci: The log files from pytest and kafka are saved in artifacts for inspection
- conftest automatically downloading Kafka to the `runtime` directory.
   - Edit: This was an old request from @tvainika , I endedup doing on this PR because I didn't want to keep both places in sync, and it made it easier to run the test suit with different Kafka versions. This does have a downside of the Kafka download time being included as part of the test run, and that can cause failures with `pytest-timeout`, because of that I changed the timing to be for the test run alone, which is not perfect because now the fixtures can be arbitrarily long. Not sure what is the best trade off here.
- ground work to run test suite with different kafka versions


hackaugusto:
I marked this as ready for review because I'm trying to trigger the build.

tvainika (approved):

tvainika:
Looks like merging this caused even linter to fail. Was it semantic conflict, do you have time to check it out soon @hackaugusto or should I revert?

hackaugusto:
just to not leave the comment unanswered: this was fixed by https://github.com/aiven/karapace/pull/386/

PR #364, opened 2022-03-22
bugfix: hotloop on get_offset_from_queue

hackaugusto:
Under some circunstances it may happen that an unexpected offset is
added to the Queue. When that happens, there is no thread to consume the
offset, so the offset is `get`ed and immediatelly `put`ed back. The
result is that `get_offset_from_queue` becomes a hot loop and spams the
logging system and slows Karapace as a whole.

Additionally, when dozens of offsets are added to the Queue, each
iteration of the loop will get the element from the top of the queue,
check it, and if it is not the target offset put it back at the end of
the queue. The result is that the method degrades to a very slow
linear search.

This commit does not fix the bug that introduces an unexpected offset to
the Queue, but it does introduced a few fixes:

- Fixes the hotloop, with this change a single Event object is used for
  each offset.
- Adds a timeout for the waiting. Not seeing the expected offset may be
  another issue (yet to be confirmed).
- Removes the linear search by an Event object.

---

This is how the issue manifests itself:

```
2022-03-20T09:38:21.574419+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574865+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574921+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175273 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.574699+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175269 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.574750+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574306+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574027+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175257 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.574250+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175261 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.574561+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574976+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574083+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574644+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574195+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
2022-03-20T09:38:21.574808+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175271 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.574362+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175263 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.574603+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175267 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.575032+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175275 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.575364+0000 karapace: KarapaceAll MainThread WARNING Put the offset: 175281 back to queue, someone else is waiting for this?
2022-03-20T09:38:21.575194+0000 karapace: KarapaceAll MainThread INFO    Starting to wait for offset: 786599 from ksr queue
```


karapace/schema_registry_apis.py
```diff
@@ -241,8 +227,27 @@ def send_kafka_message(self, key, value):
...
+
+        if self.ksr.offset_watcher.wait_for_offset(sent_offset, timeout=1 * MINUTE) is True:
+            self.log.info(
+                "Schema reader has found key. key: %r, value: %r, offset: %r",
+                key,
+                value,
+                sent_offset,
+            )
+        else:
+            raise RuntimeError(
```
hackaugusto, line 247: This will be converted to an [internal server error](https://github.com/aiven/karapace/blob/63f6b50140c3ad30183a477fc75a63d3623a33f6/karapace/rapu.py#L334-L338).

To introduce an error message the content type needs to be passed along from all the way down from the handler

karapace/schema_registry_apis.py
```diff
@@ -241,8 +227,27 @@ def send_kafka_message(self, key, value):
         future = self.producer.send(self.config["topic_name"], key=key, value=value)
         self.producer.flush(timeout=self.kafka_timeout)
         msg = future.get(self.kafka_timeout)
-        self.log.debug("Sent kafka msg key: %r, value: %r, offset: %r", key, value, msg.offset)
-        self.get_offset_from_queue(msg.offset)
+        sent_offset = msg.offset
+
+        self.log.info(
+            "Waiting for schema reader to caugth up. key: %r, value: %r, offset: %r",
```
jjaakola-aiven, line 233:
```suggestion
            "Waiting for schema reader to caught up. key: %r, value: %r, offset: %r",
```

karapace/schema_registry_apis.py
```diff
@@ -241,8 +227,27 @@ def send_kafka_message(self, key, value):
...
+            key,
+            value,
+            sent_offset,
+        )
+
+        if self.ksr.offset_watcher.wait_for_offset(sent_offset, timeout=1 * MINUTE) is True:
```
jjaakola-aiven, line 239: One minute timeout is plenty, could this be lower? Some HTTP clients have shorter default timeouts.

karapace/schema_registry_apis.py
jjaakola-aiven, line 247: Should the error be domain specific here? And the response to client to be `504 Gateway timeout`?

jjaakola-aiven (dismissed):
Could you add a test?

hackaugusto:
> Could you add a test?


I thought about this, but I'm not sure if there is a good way to test the endpoint. The old implementation was working AFAIK, the problem was just that it spammed the logs and was slow. So I don't have a good way to implement a regression test.

I can add a unit test for the `OffsetsWatcher` interface. Or if you really want I can mock `log` to assert on how many times it is called.

karapace/schema_reader.py
```diff
@@ -88,6 +88,43 @@ class SchemaType(str, Enum):
...
+
+    def offset_seen(self, new_offset: int) -> None:
+        with self._events_lock:
+            # Note: The reader thread could be already waiting on the offset,
+            # in that case the existing object must be used, otherwise a new
+            # one is created.
+            event = self._events.setdefault(new_offset, Event())
+            event.set()
+
+    def wait_for_offset(self, expected_offset: int, timeout: int) -> bool:
```
tvainika, line 112: `event.wait` takes timeout as a floating point number specifying a timeout for the operation in seconds (or fractions thereof). Therefore suggest convering type to float, and somehow documenting the unit of timeout (seconds).

karapace/schema_registry_apis.py
tvainika: Also this is wrong as it is not one minute but actually thousand minutes, see karapace constants.py and my comment about `wait_for_offset` type signature.

tvainika (changes requested):
Looks much improvement over original code, but see inline comments about timeout units.

karapace/schema_registry_apis.py
hackaugusto: What would be the benefit of having a shorter timeout here? I'm trying to avoid causing further problems, so I was conservative.

Thanks @tvainika , you're right, MINUTE is in ms and this API is in secs

karapace/schema_registry_apis.py
hackaugusto: These are the current errors:

https://github.com/aiven/karapace/blob/60430a8eee3bbe8949dac3a43af9f623884333b9/karapace/schema_registry_apis.py#L18-L35

I tried to not change this ATM because those are compatible with CSR

tests/unit/test_schema_reader.py
```diff
@@ -0,0 +1,44 @@
+from concurrent.futures import ThreadPoolExecutor
```
hackaugusto, line 1: I'm using a `ThreadPoolExecutor` instead of plain threads because the `Future` object does the work of capturing the exception and re-raising it on the main thread.

hackaugusto:
@mhoffm-aiven suggested to instead of using Events, to just do a hot loop on the Schema Registry API. Something like this:

```python
while msg.offset > self.ksr.offset:
   time.sleep(0.1)
```

I prefer the blocking code with synchronization primitives instead, but that is undoubtly more complicated, comments?

mhoffm-aiven:
> @mhoffm-aiven suggested to instead of using Events, to just do a hot loop on the Schema Registry API. Something like this:
> 
> ```python
> while msg.offset > self.ksr.offset:
>    time.sleep(0.1)
> ```
> 
> I prefer the blocking code with synchronization primitives instead, but that is undoubtly more complicated, comments?

i would also suggest using configurable amount of retries and retry backoff and returning 202 ACCEPTED if we dont see the offset during the retry loop

karapace/schema_registry_apis.py
jjaakola-aiven: Internal error looks to be the behavior in CSR. Although there is an error code for operation timeout `50002`, e.g. https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string-%20subject)-versions

karapace/schema_reader.py
```diff
@@ -11,15 +11,61 @@
...
 import logging
 import time
 import ujson
 
 log = logging.getLogger(__name__)
+Offset = int
+
+# The value `0` is a valid offset and it repreesents the first message produced
+# to a topic, therefore it can not be used.
+OFFSET_EMPTY = -1
```
hackaugusto, line 26: I hit this bug while implementing the alternative https://github.com/aiven/karapace/pull/373. Basically the code thought the first message was consumed already because the start value was incorrect, so the follow up request didn't get the produced schema and the test failed.

karapace/schema_reader.py
```diff
@@ -11,15 +11,61 @@
...
 from typing import Dict
 
 import logging
 import time
 import ujson
 
 log = logging.getLogger(__name__)
+Offset = int
+
+# The value `0` is a valid offset and it repreesents the first message produced
```
jjaakola-aiven, line 24:
```suggestion
# The value `0` is a valid offset and it represents the first message produced
```

karapace/schema_reader.py
```diff
@@ -32,20 +78,32 @@ def __init__(self, config, master_coordinator=None):
...
         self.running = True
         self.id_lock = Lock()
         sentry_config = config.get("sentry", {"dsn": None}).copy()
         if "tags" not in sentry_config:
             sentry_config["tags"] = {}
         self.stats = StatsClient(sentry_config=sentry_config)
 
+        # Thread synchronization objects
+        # - offset is used by the REST API to wait until this thread has
+        # consume the produced messages. This makes the REST APIs more
```
jjaakola-aiven, line 95:
```suggestion
        # consumed the produced messages. This makes the REST APIs more
```

hackaugusto:
FYI @mhoffm-aiven we are keeping the 500, ref: https://github.com/aiven/karapace/pull/364/files#r838239615

PR #366, opened 2022-03-24
perf: Skip deepcopy

fingon:
Deepcopy uses most of the CPU currently when publishing to a topic. 

Deepcopy is not necessary. So not doing it. Do one shallow copy where it matters (reading from dicts is async thread safe in Python anyway, so reusing same structure without making copy of it should be fine).


mhoffm-aiven (approved):
lgtm, good find

jjaakola-aiven (approved):

PR #367, opened 2022-03-24
Fix Karapace REST address in readme

dswiecki:
Karapace REST adress is wrongly configured in curl snippets


tvainika (approved):
Thank you!

PR #368, opened 2022-03-25
Handle incompatible schemas with produce endpoint

tvainika:

Handle incompatible schemas with produce endpoint

# Why this way

Produce message endpoint takes full schemas with the message in some cases, and registers the schema during producing the message.  Produce endpoint needs to handle also case if the schema is not compatible with a valid error message.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -464,6 +464,15 @@ class KafkaRest(KarapaceBase):
                 content_type=content_type,
                 status=HTTPStatus.BAD_REQUEST,
             )
+        except SchemaRetrievalError:
+            self.r(
+                body={
+                    "error_code": RESTErrorCodes.SCHEMA_RETRIEVAL_ERROR.value,
+                    "message": f"Error when registering schema. format = {schema_type}, subject = {topic}-{prefix}",
+                },
+                content_type=content_type,
+                status=HTTPStatus.REQUEST_TIMEOUT,
+            )
 
     async def _prepare_records(
         self,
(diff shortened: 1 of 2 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -464,6 +464,15 @@ async def validate_schema_info(self, data: dict, prefix: str, content_type: str,
...
                 status=HTTPStatus.BAD_REQUEST,
             )
+        except SchemaRetrievalError:
+            self.r(
+                body={
+                    "error_code": RESTErrorCodes.SCHEMA_RETRIEVAL_ERROR.value,
+                    "message": f"Error when registering schema. format = {schema_type}, subject = {topic}-{prefix}",
+                },
+                content_type=content_type,
+                status=HTTPStatus.REQUEST_TIMEOUT,
```
jjaakola-aiven, line 474: Why 408 timeout? Based on the PR description better alternatives would be 406 not acceptable, 409 conflict,  422 unprocessable entity.

jjaakola-aiven (changes requested):

karapace/kafka_rest_apis/__init__.py
tvainika: 408 is same what CSR implementation returns, so I reused it. I admit it is not the best suitable though.

jjaakola-aiven: Then it is the correct choice.

jjaakola-aiven (approved):

jjaakola-aiven (approved):

PR #371, opened 2022-03-29
Update precommit tools

hackaugusto:
Updates black to circumvent an issue with click, ref: https://github.com/psf/black/issues/2969
Build logs of another PR with the same error: https://github.com/aiven/karapace/runs/5735453639?check_suite_focus=true

I would also have updated pylint, but that needs quite a few a changes

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,8 +1,7 @@
 # linters and formatters
 isort==4.3.10
-pycodestyle==2.6.0
 pylint==2.7.4
-flake8==3.8.4
+flake8==4.0.1
 mypy==0.812
 black==22.3.0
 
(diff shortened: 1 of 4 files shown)


hackaugusto:
It's a bit annoying having to update the versions on two places, I did add the `requirements-dev.txt` a while back because it is easier to create a venv and integrate these tools with an editor. And I suppose changing everything to use `local` is not a great fix either, because then creating the precommits is harder. I guess it is just something that we have to live with

requirements-dev.txt
```diff
@@ -1,9 +1,9 @@
 # linters and formatters
-isort==4.3.10
-pycodestyle==2.6.0
```
hackaugusto, line 3: I ended up removing pycodestyle, it should be run by flake8.

jjaakola-aiven (approved):

PR #373, opened 2022-03-30
bugfix: hotloop on get_offset_from_queue

hackaugusto:
This PR is an alternative for https://github.com/aiven/karapace/pull/364

Under some circumstances it may happen that an unexpected offset is
added to the Queue. When that happens, there is no thread to consume the
offset, so the offset is `get`ed and immediately `put`ed back. The
result is that `get_offset_from_queue` becomes a hot loop and spams the
logging system and slows Karapace as a whole.

Additionally, when dozens of offsets are added to the Queue, each
iteration of the loop will get the element from the top of the queue,
check it, and if it is not the target offset put it back at the end of
the queue. The result is that the method degrades to a very slow linear
search.

This commit fixes the following issues:

- Fixes the hotloop, by using a sleep timeout.
- Removes the linear search by an Event object, by comparing against a
  plain offset.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -214,21 +214,6 @@ class KarapaceSchemaRegistry(KarapaceBase):
             )
         return compatibility_mode
 
-    def get_offset_from_queue(self, sent_offset):
-        start_time = time.monotonic()
-        while True:
-            self.log.info("Starting to wait for offset: %r from ksr queue", sent_offset)
-            offset = self.ksr.queue.get()
-            if offset == sent_offset:
-                self.log.info(
-                    "We've consumed back produced offset: %r message back, everything is in sync, took: %.4f",
-                    offset,
-                    time.monotonic() - start_time,
-                )
-                break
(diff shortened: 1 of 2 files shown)


karapace/schema_registry_apis.py
```diff
@@ -238,8 +223,20 @@ def send_kafka_message(self, key, value):
...
         msg = future.get(self.kafka_timeout)
-        self.log.debug("Sent kafka msg key: %r, value: %r, offset: %r", key, value, msg.offset)
-        self.get_offset_from_queue(msg.offset)
+        sent_offset = msg.offset
+
+        self.log.debug("[Reader Synchronization] Waiting for key: %r, value: %r, offset: %r", key, value, sent_offset)
+        start = time.monotonic()
+
+        while sent_offset > self.ksr.offset:
+            time.sleep(0.1)
```
MichaHoffmann, line 232: We should use a wait timeout here i think

hackaugusto: I intentionally left it out, would do in a follow up

MichaHoffmann: ah! ok nevermind then! looks good otherwise i think

PR #374, opened 2022-03-30
Typing and simplification of the utils module

hackaugusto:
Notes:
- Replaced stdlib's json with ujson in the `json_encode`, should be faster
- I simplified some function signatures:
  - removed keyword argument that have a default and were never used by any caller
  - remove keyword arguments that have a default, and were always given the same value as the default
- Moved the Client code to a separate module and add typing. Also simplified it by:
  - Remove the fallback code that used requests, turns out that `_get_client` always returned a client, so that was dead code

### DIFF
--- a/karapace/client.py
+++ b/karapace/client.py
@@ -29,9 +29,7 @@ class Result:
         json_result: JsonData,
         headers: Optional[Mapping] = None,
     ) -> None:
-        # We create both status and status_code so people can be agnostic on whichever to use
         self.status_code = status
-        self.status = status
         self.json_result = json_result
         self.headers = headers if headers else {}
 
--- a/karapace/client.py
+++ b/karapace/client.py
@@ -16,7 +16,7 @@ import ssl
 log = logging.getLogger(__name__)
 
 
-async def get_aiohttp_client() -> aiohttp.ClientSession:
+async def _get_aiohttp_client() -> aiohttp.ClientSession:
     return aiohttp.ClientSession()
 
 
@@ -40,7 +40,7 @@ class Result:
 
 
 class Client:
-    def __init__(self, server_uri=None, client_factory=get_aiohttp_client, server_ca: Optional[str] = None):
+    def __init__(self, server_uri=None, client_factory=_get_aiohttp_client, server_ca: Optional[str] = None):
         self.server_uri = server_uri or ""
         self.path_for = partial(urljoin, self.server_uri)
         self.session = requests.Session()
--- a/karapace/client.py
+++ b/karapace/client.py
@@ -4,163 +4,176 @@ karapace - utils
 Copyright (c) 2022 Aiven Ltd
 See LICENSE for details
 """
-from functools import partial
-from typing import Optional
+from aiohttp import ClientSession
+from collections.abc import Mapping
+from typing import Awaitable, Callable, Optional, Union
 from urllib.parse import urljoin
 
-import aiohttp
 import logging
-import requests
 import ssl
 
 log = logging.getLogger(__name__)
+Path = str
+Headers = dict
+JsonData = object  # Type of the result after parsing JSON
 
 
-async def _get_aiohttp_client() -> aiohttp.ClientSession:
-    return aiohttp.ClientSession()
+async def _get_aiohttp_client() -> ClientSession:
+    return ClientSession()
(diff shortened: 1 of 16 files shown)


karapace/client.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self.ssl_mode: Union[None, bool, ssl.SSLContext]
+        if server_ca is None:
+            self.ssl_mode = False
+        else:
+            self.ssl_mode = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+            self.ssl_mode.load_verify_locations(cafile=server_ca)
+        self._client: Optional[ClientSession] = None
+
+    def path_for(self, path: Path) -> str:
```
hackaugusto, line 71: mypy was inferring that `path_for` was a `Callable[[str], AnyStr]`, so I moved it to a method to get typing working.

karapace/client.py
```diff
@@ -0,0 +1,179 @@
...
+import logging
+import ssl
+
+log = logging.getLogger(__name__)
+Path = str
+Headers = dict
+JsonData = object  # Type of the result after parsing JSON
+
+
+async def _get_aiohttp_client() -> ClientSession:
```
hackaugusto, line 21: This only exists because the function should be async, but it is not used outside of this module, so I made it private. IMO it would make sense to remove this callback and just inject the `ClientSession` in the `Client` constructor.

hackaugusto:
Note: This is a breaking change, both the `Client` class is in a different module, and this removes the `Response.code` attribute. I'm not sure if people are using Karapace as a library though.

tvainika:
> I'm not sure if people are using Karapace as a library though.

I think we do not support such use case, so I don't think this counts as a breaking change.

jjaakola-aiven (commented):
The `karapace.utils` configuration in `mypy.ini` ignores errors.

hackaugusto:
> The karapace.utils configuration in mypy.ini ignores errors.

Yep, I didn't fix all the types in this PR, it was getting to big, that is part of the reason why I moved the client to a new module, now that is type checked.

Sample output from mypy if the errors are _not_ ignored and would still need to be fixed before the flag is removed:

```
> mypy karapace/utils.py
karapace/utils.py:11: error: Skipping analyzing "kafka.client_async": module is installed, but missing library stubs or py.typed marker
karapace/utils.py:11: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
karapace/utils.py:15: error: Skipping analyzing "kafka": module is installed, but missing library stubs or py.typed marker
karapace/utils.py:70: error: Function is missing a return type annotation
karapace/utils.py:70: error: Function is missing a type annotation for one or more arguments
karapace/utils.py:141: error: Function is missing a type annotation
karapace/utils.py:145: error: Function is missing a return type annotation
karapace/utils.py:145: note: Use "-> None" if function does not return a value
karapace/utils.py:159: error: Function is missing a type annotation
karapace/utils.py:166: error: Function is missing a type annotation
karapace/utils.py:172: error: Cannot determine type of "_metadata_refresh_in_progress"
karapace/utils.py:194: error: Incompatible types in assignment (expression has type "Optional[List[Any]]", variable has type "List[Any]")
karapace/utils.py:205: error: Function is missing a type annotation
karapace/utils.py:223: error: Function is missing a type annotation
karapace/utils.py:228: error: Function is missing a type annotation
karapace/utils.py:233: error: Function is missing a return type annotation
karapace/utils.py:236: error: Function is missing a return type annotation
Found 15 errors in 1 file (checked 1 source file)
```

Note: The type annotations that I did add were type checked, I did disable the flag while adding them, and just didn't commit the change because of the above.

karapace/client.py
```diff
@@ -0,0 +1,179 @@
...
+
+class Result:
+    def __init__(
+        self,
+        status: int,
+        json_result: JsonData,
+        headers: Optional[Mapping] = None,
+    ) -> None:
+        # We create both status and status_code so people can be agnostic on whichever to use
+        self.status_code = status
```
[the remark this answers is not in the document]
jjaakola-aiven: I reviewed the points where `.status` is used. Looks correct.

jjaakola-aiven (approved):

PR #375, opened 2022-03-30
Simplified schema reader

hackaugusto:
# What is this

This is reintroducing _some_ of the commits reverted by https://github.com/aiven/karapace/pull/295. It is not reintroducing the ones for freeing memory, because that particular implementation has caused issues in the past when a schema topic contains schemas with duplicated IDs.

This cherry picks the other commits that split the loop and isolate the code, it should make the code easier to follow and prepare the ground for some fixes, like enabling full typing information, and isolating the schemas dictionaries (which are currently read by the APIs and may result in exceptions for modifying a dictionary while being read)

This also includes some bugfixes:
- for soft deleted schemas which were not properly tracked.
- for config messages of new subjects.
- for updates of global id when processing soft deleted schemas.


karapace/schema_reader.py
```diff
@@ -8,18 +8,25 @@
...
 import time
 import ujson
 
 log = logging.getLogger(__name__)
 Offset = int
+Subject = str
+Version = int
+Schema = Dict[str, Any]
+# Container type for a subject, with configuration settings and all the schemas
+SubjectData = Dict[str, Any]
```
hackaugusto, line 29: I tried making this into `dataclass`, but this data is read by the API and it resulted in lots of changes.

I also considered using a `TypedDict` here, but that is not available for 2.7 (which is a supported version). It could be done in a follow up if we decide to either remove 2.7 support or to include the `typing-extensions` dependency.

hackaugusto:
I'm marking this as a draft, should write a regression test for the bug first.

tests/integration/test_schema_reader.py
```diff
@@ -0,0 +1,83 @@
...
+from karapace.config import set_config_defaults
+from karapace.constants import DEFAULT_SCHEMA_TOPIC
+from karapace.master_coordinator import MasterCoordinator
+from karapace.schema_reader import KafkaSchemaReader
+from karapace.utils import json_encode
+from tests.schemas.json_schemas import TRUE_SCHEMA
+from tests.utils import create_group_name_factory, create_subject_name_factory, KafkaServers
+
+
+def test_regression_soft_delete_schemas_should_be_registered(
```
hackaugusto, line 12: If ran on master this test will fail, here it passes because the bug is fixed.

karapace/schema_reader.py
```diff
@@ -254,105 +266,122 @@ def handle_messages(self):
...
-                    }
-                }
-                self.log.info("Setting schema_id: %r with schema: %r", value["id"], typed_schema)
-                self.schemas[value["id"]] = typed_schema
-                if value["id"] > self.global_schema_id:  # Not an existing schema
-                    self.global_schema_id = value["id"]
-            elif value.get("deleted", False) is True:
-                self.log.info("Deleting subject: %r, version: %r", subject, value["version"])
-                if not value["version"] in self.subjects[subject]["schemas"]:
-                    self.schemas[value["id"]] = typed_schema
```
hackaugusto, line 320: This is the bug, it should also be added to `self.subjects[schema_subject]["schemas"]`

karapace/schema_reader.py
```diff
@@ -254,105 +266,122 @@ def handle_messages(self):
                 if self.ready and add_offsets:
                     self.offset_watcher.offset_seen(self.offset)
 
-    def handle_msg(self, key: dict, value: dict):
```
hackaugusto, line 257: `value: dict` was not correct. For instance, a hard deleted schema has an empty value

- [Here it is set to None](https://github.com/aiven/karapace/blob/master/karapace/schema_registry_apis.py#L603-L605)
- [Here it translated to ""](https://github.com/aiven/karapace/blob/4959ad42d281859ac0efe03566141595c6fb28a2/karapace/schema_registry_apis.py#L272)

Then while [decoding the empty string](https://github.com/aiven/karapace/blob/4959ad42d281859ac0efe03566141595c6fb28a2/karapace/schema_reader.py#L242-L248), it should fail, because an empty string is not valid json:

```python
>>> import json
>>> json.loads('')
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

So the `value` can be `None`.

karapace/schema_reader.py
```diff
@@ -254,105 +266,122 @@ def handle_messages(self):
...
-        if key["keytype"] == "CONFIG":
-            if "subject" in key and key["subject"] is not None:
-                if not value:
-                    self.log.info("Deleting compatibility config completely for subject: %r", key["subject"])
-                    self.subjects[key["subject"]].pop("compatibility", None)
```
hackaugusto, line 262: This was another bug, this would throw if the first message in the topic was delete the config compatibility. For Karapace only topics this cant happen, because the delete endpoint is missing https://github.com/aiven/karapace/issues/380. This would only be an issue for a CSR -> Karapace migration, where the topic config was delete and it happened to be the first message in the topic.

karapace/schema_reader.py
```diff
@@ -254,105 +266,122 @@ def handle_messages(self):
...
             if not value:
-                subject, version = key["subject"], key["version"]
-                self.log.info("Deleting subject: %r version: %r completely", subject, version)
-                if subject not in self.subjects:
-                    self.log.error("Subject %s did not exist, should have", subject)
-                elif version not in self.subjects[subject]["schemas"]:
-                    self.log.error("Version %d for subject %s did not exist, should have", version, subject)
-                else:
-                    self.subjects[subject]["schemas"].pop(version, None)
-                return
```
hackaugusto, line 285: Here we have a memory leak. This could be the last use of the schema and `self.schemas` should be cleared too. This PR is _not_ fixing this, that was done on https://github.com/aiven/karapace/pull/284 and had to be reverted https://github.com/aiven/karapace/pull/295

hackaugusto: I believe there is another bug here, but I'm not sure how to fix this.

- If the very last produced message is for a hard delete of a schema
- The topic is compacted
- The server is restarted

Then the information of the global ID is lost, because the `key` contains only the `version` of the schema in the subject, but not the `id` that is global to all subjects. That means the `id` will be reused.

karapace/schema_reader.py
```diff
@@ -254,105 +266,122 @@ def handle_messages(self):
...
-                            "id": value["id"],
-                            "deleted": value.get("deleted", False),
-                        }
-                    }
-                }
-                self.log.info("Setting schema_id: %r with schema: %r", value["id"], typed_schema)
-                self.schemas[value["id"]] = typed_schema
-                if value["id"] > self.global_schema_id:  # Not an existing schema
-                    self.global_schema_id = value["id"]
-            elif value.get("deleted", False) is True:
```
hackaugusto, line 317: There is another bug here wrt to the `global_schema_id`. It is the same as the comment above:

https://github.com/aiven/karapace/pull/375/files#r839827344

But here we do have the data to update the global id. The bug is that it was not updated.

karapace/schema_reader.py
```diff
@@ -250,109 +262,131 @@ def handle_messages(self):
...
+
+        subjects_schemas[schema_version] = {
+            "schema": typed_schema,
+            "version": schema_version,
+            "id": schema_id,
+            "deleted": schema_deleted,
+        }
+        with self.id_lock:
+            self.schemas[schema_id] = typed_schema
+            self.global_schema_id = max(self.global_schema_id, schema_id)
```
hackaugusto, line 366: This is the bugfix for the `global_schema_id` mentioned above: https://github.com/aiven/karapace/pull/375/files#r839829337

The fix is simple, always update the `global_schema_id`.

karapace/schema_reader.py
```diff
@@ -250,109 +262,131 @@ def handle_messages(self):
...
+    def _handle_msg_schema_hard_delete(self, key: dict) -> None:
+        subject, version = key["subject"], key["version"]
+
+        if subject not in self.subjects:
+            self.log.error("Hard delete: Subject %s did not exist, should have", subject)
+        elif version not in self.subjects[subject]["schemas"]:
+            self.log.error("Hard delete: Version %d for subject %s did not exist, should have", version, subject)
+        else:
+            self.log.info("Hard delete: subject: %r version: %r", subject, version)
+            self.subjects[subject]["schemas"].pop(version, None)
```
hackaugusto, line 316: Here is where the memory leak mentioned above is still a problem:

https://github.com/aiven/karapace/pull/375/files#r839821293

The previous fix was to use a WeakDict: 

https://github.com/aiven/karapace/pull/284/files#diff-d0d95f66bd9730341ddb7c386650f61768378934401a70d05ec9dac690634ea8R126

karapace/schema_reader.py
```diff
@@ -250,109 +262,131 @@ def handle_messages(self):
...
-                subject_data = self.subjects.get(key["subject"])
-                subject_data["compatibility"] = value["compatibilityLevel"]
-            else:
-                self.log.info("Setting global config to: %r, value: %r", value["compatibilityLevel"], value)
-                self.config["compatibility"] = value["compatibilityLevel"]
-        elif key["keytype"] == "SCHEMA":
+    def _handle_msg_config(self, key: dict, value: Optional[dict]) -> None:
+        subject = key.get("subject")
+        if subject is not None:
+            if subject not in self.subjects:
```
hackaugusto, line 277: This is the fix for the bug mentioned on the side:

https://github.com/aiven/karapace/pull/375/files#r839817686

I just moved the initialization up a bit, so that it is always done.

karapace/schema_reader.py
vtainio: Could you elaborate on this a bit? Say our `global_schema_id` counter is at `5` and then we register another subject that will reuse an older schema that we have with `schema_id: 1`. Wouldn't updating to that eventually lead to duplicate id's or did you have some other update in mind?

tests/integration/test_schema_reader.py
```diff
@@ -0,0 +1,162 @@
...
+        value = ""  # Delete the config
+
+        future = producer.send(
+            DEFAULT_SCHEMA_TOPIC,
+            key=json_encode(key, binary=True),
+            value=json_encode(value, binary=True),
+        )
+        msg = future.get()
+
+        assert schema_reader.offset_watcher.wait_for_offset(msg.offset, timeout=5) is True
```
hackaugusto, line 161: This is a very fragile test, but for the current design of the schema reader it is the easiest I could do. Basically all exceptions are caught and the thread keeps running, so I couldn't do a `assert thread.is_alive()` because it never dies, so I used the offset_watcher instead, because when an exceptions happens the Event is not set.

IMO it would make sense to isolate the actual message handling code from the thread, so that can be tested directly, than this kinda of problem goes away (and the test setup is simpler)

jjaakola-aiven (approved):

tests/integration/test_schema_reader.py
```diff
@@ -0,0 +1,162 @@
...
+    kafka_servers: KafkaServers,
+    producer: KafkaProducer,
+) -> None:
+    """Soft deleted schemas should be registered in the subject.
+
+    The following would happen when:
+    - Schema 1 is registered
+    - Schema 2 is registered
+    - Schema 2 is soft deleted
+    - Topic is compatected
```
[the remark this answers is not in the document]
hackaugusto: hahahaha, that was a hell of a spelling, will fix.

hackaugusto:
@tvainika could you also have a look at this?

PR #384, opened 2022-04-04
Reduced logging

hackaugusto:
Merge after https://github.com/aiven/karapace/pull/391 (It fix port allocation for integration tests)

## Why?

Logging schemas has two problems (1) It is cpu intensive, and this is a problem while starting Karapace against a topic with a couple of thousands of schemas (2) Schemas are private information, and this changes the logging with them to `TRACE` level instead of `DEBUG`.

This also makes some logger instances private.

### DIFF
--- a/karapace/client.py
+++ b/karapace/client.py
@@ -12,10 +12,11 @@ from urllib.parse import urljoin
 import logging
 import ssl
 
-log = logging.getLogger(__name__)
 Path = str
 Headers = dict
 JsonData = object  # Type of the result after parsing JSON
+
+LOG = logging.getLogger(__name__)
 
 
 async def _get_aiohttp_client() -> ClientSession:
@@ -74,7 +75,7 @@ class Client:
             if self._client is not None:
                 await self._client.close()
         except:  # pylint: disable=bare-except
-            log.info("Could not close client")
+            LOG.error("Could not close client")
 
     async def get_client(self) -> ClientSession:
         if self._client is None:
--- a/karapace/compatibility/jsonschema/checks.py
+++ b/karapace/compatibility/jsonschema/checks.py
@@ -30,10 +30,7 @@ from karapace.compatibility.jsonschema.utils import (
 )
 from typing import Any, List, Optional
 
-import logging
 import networkx as nx
-
-LOG = logging.getLogger(__name__)
 
 INTRODUCED_INCOMPATIBILITY_MSG_FMT = "Introduced incompatible assertion {assert_name} with value {introduced_value}"
 RESTRICTED_INCOMPATIBILITY_MSG_FMT = "More restrictive assertion {assert_name} from {writer_value} to {reader_value}"
@@ -248,7 +245,6 @@ def compatibility_rec(
     # reader has type `array` to represent a list, and the writer is either a
     # different type or it is also an `array` but now it representes a tuple.
     if reader_schema is None and writer_schema is not None:
-        LOG.debug("Schema removed reader_schema.type='%r'", get_type_of(reader_schema))
         return incompatible_schema(
             incompat_type=Incompatibility.schema_removed,
             message="schema removed",
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -19,7 +19,7 @@ import ujson
 Config = Dict[str, Union[None, str, int, bool, List[str], AccessLogger]]
 LOG = logging.getLogger(__name__)
 HOSTNAME = socket.gethostname()
-
+SASL_PLAIN_PASSWORD = "sasl_plain_password"
 DEFAULTS = {
     "access_logs_debug": False,
     "access_log_class": None,
@@ -53,7 +53,7 @@ DEFAULTS = {
     "ssl_password": None,
     "sasl_mechanism": None,
     "sasl_plain_username": None,
-    "sasl_plain_password": None,
+    SASL_PLAIN_PASSWORD: None,
     "topic_name": DEFAULT_SCHEMA_TOPIC,
     "metadata_max_age_ms": 60000,
     "admin_metadata_max_age": 5,
@@ -67,6 +67,7 @@ DEFAULTS = {
     "master_election_strategy": "lowest",
     "protobuf_runtime_directory": "runtime",
 }
+SECRET_CONFIG_OPTIONS = [SASL_PLAIN_PASSWORD]
 
 
 class InvalidConfiguration(Exception):
@@ -109,12 +110,20 @@ def set_settings_from_environment(config: Config) -> None:
         env_name = config_name_with_prefix.upper()
         env_val = os.environ.get(env_name)
         if env_val is not None:
-            LOG.debug(
-                "Populating config value %r from env var %r with %r instead of config file",
-                config_name,
-                env_name,
-                env_val,
-            )
+            if config_name not in SECRET_CONFIG_OPTIONS:
+                LOG.info(
+                    "Populating config value %r from env var %r with %r instead of config file",
+                    config_name,
+                    env_name,
+                    env_val,
+                )
+            else:
+                LOG.info(
+                    "Populating config value %r from env var %r instead of config file",
+                    config_name,
+                    env_name,
+                )
+
             config[config_name] = parse_env_value(env_val)
 
 
--- a/karapace/kafka_rest_apis/admin.py
+++ b/karapace/kafka_rest_apis/admin.py
@@ -9,12 +9,10 @@ from typing import List
 
 import logging
 
+LOG = logging.getLogger(__name__)
+
 
 class KafkaRestAdminClient(KafkaAdminClient):
-    def __init__(self, **configs):
-        super().__init__(**configs)
-        self.log = logging.getLogger("AdminClient")
-
     def get_topic_config(self, topic: str) -> dict:
         config_version = self._matching_api_version(DescribeConfigsRequest)
         req_cfgs = [ConfigResource(ConfigResourceType.TOPIC, topic)]
@@ -50,7 +48,7 @@ class KafkaRestAdminClient(KafkaAdminClient):
         except Cancelled:
             if retries > 3:
                 raise
-            self.log.debug("Retrying metadata with %d retires", retries)
+            LOG.debug("Retrying metadata with %d retires", retries)
             return self.cluster_metadata(topics, retries + 1)
         return self._make_metadata_response(future.value)
 
--- a/karapace/kafka_rest_apis/consumer_manager.py
+++ b/karapace/kafka_rest_apis/consumer_manager.py
@@ -24,21 +24,20 @@ KNOWN_FORMATS = {"json", "avro", "binary", "jsonschema", "protobuf"}
 OFFSET_RESET_STRATEGIES = {"latest", "earliest"}
 
 TypedConsumer = namedtuple("TypedConsumer", ["consumer", "serialization_format", "config"])
+LOG = logging.getLogger(__name__)
+
+
+def new_name() -> str:
+    return str(uuid.uuid4())
 
 
 class ConsumerManager:
     def __init__(self, config: dict) -> None:
         self.config = config
         self.hostname = f"http://{self.config['advertised_hostname']}:{self.config['port']}"
-        self.log = logging.getLogger("RestConsumerManager")
         self.deserializer = SchemaRegistryDeserializer(config=config)
         self.consumers = {}
         self.consumer_locks = defaultdict(Lock)
-
-    def new_name(self) -> str:
-        name = str(uuid.uuid4())
-        self.log.debug("Generated new consumer name: %s", name)
-        return name
 
     @staticmethod
     def _assert(cond: bool, code: HTTPStatus, sub_code: int, message: str, content_type: str) -> None:
@@ -155,12 +154,15 @@ class ConsumerManager:
     # CONSUMER
     async def create_consumer(self, group_name: str, request_data: dict, content_type: str):
         group_name = group_name.strip("/")
-        self.log.info("Create consumer request for group  %s", group_name)
-        consumer_name = request_data.get("name") or self.new_name()
+        consumer_name = request_data.get("name") or new_name()
         internal_name = self.create_internal_name(group_name, consumer_name)
         async with self.consumer_locks[internal_name]:
             if internal_name in self.consumers:
-                self.log.error("Error creating duplicate consumer in group %s with id %s", group_name, consumer_name)
+                LOG.warning(
+                    "Error creating duplicate consumer in group %s with id %s",
+                    group_name,
+                    consumer_name,
+                )
                 KarapaceBase.r(
                     status=HTTPStatus.CONFLICT,
                     content_type=content_type,
@@ -170,11 +172,14 @@ class ConsumerManager:
                     },
                 )
             self._validate_create_consumer(request_data, content_type)
-            self.log.info(
-                "Creating new consumer in group %s with id %s and request_info %r", group_name, consumer_name, request_data
-            )
             for k in ["consumer.request.timeout.ms", "fetch_min_bytes"]:
                 convert_to_int(request_data, k, content_type)
+            LOG.info(
+                "Creating new consumer in group. group name: %s consumer name: %s request_data %r",
+                group_name,
+                consumer_name,
+                request_data,
+            )
             try:
                 enable_commit = request_data.get("auto.commit.enable", self.config["consumer_enable_auto_commit"])
                 if isinstance(enable_commit, str):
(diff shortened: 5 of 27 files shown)


karapace/schema_reader.py
```diff
@@ -248,56 +249,59 @@ def handle_messages(self) -> None:
...
+                self.offset = msg.offset
                 if self.ready and add_offsets:
                     self.offset_watcher.offset_seen(self.offset)
 
     def _handle_msg_config(self, key: dict, value: Optional[dict]) -> None:
         subject = key.get("subject")
         if subject is not None:
             if subject not in self.subjects:
-                self.log.info("Adding first version of subject: %r with no schemas", subject)
+                self._log.info("Adding first version of subject: %r with no schemas", subject)
```
hackaugusto, line 282: I thought about not logging `subject`s too, but I think in that case we would have too little logging available.

karapace/schema_reader.py
```diff
@@ -248,56 +249,59 @@ def handle_messages(self) -> None:
                 try:
                     key = ujson.loads(msg.key.decode("utf8"))
                 except ValueError:
-                    self.log.exception("Invalid JSON in msg.key: %r, value: %r", msg.key, msg.value)
+                    self._log.exception("Invalid JSON in msg.key")
+                    self._log.log(TRACE, "Invalid JSON in msg.key: %r, value: %r", msg.key, msg.value)
```
tvainika, line 253: I'm not really fond of this non-standard `TRACE` level logging. I think we should just go with debug level logging for these, which is kind of standard, but not for production usage.

karapace/utils.py
```diff
@@ -144,14 +143,15 @@ class KarapaceKafkaClient(KafkaClient):
     def __init__(self, **configs):
         kafka.client_async.BrokerConnection = KarapaceBrokerConnection
         super().__init__(**configs)
+        self._log = logging.getLogger(__name__)
```
tvainika, line 146: Here also IMHO going backwards.

karapace/schema_reader.py
```diff
@@ -247,26 +249,29 @@ def handle_messages(self) -> None:
                 try:
                     key = ujson.loads(msg.key.decode("utf8"))
                 except ValueError:
-                    self._log.exception("Invalid JSON in msg.key: %r, value: %r", msg.key, msg.value)
+                    self._log.exception("Invalid JSON in msg.key")
```
juha-aiven, line 252: Is the idea to hide the values or to have less logging?

If the latter, would it be too much to log `msg.key`. Now this message has no context of what's wrong. We could always truncate the key to e.g. 16 characters.

tvainika (changes requested):
Seems I questioned some of your decision here. Maybe some discussion to have as we might just have differing opinions here?

karapace/schema_reader.py
```diff
@@ -247,26 +249,29 @@ def handle_messages(self) -> None:
...
+                    self._log.log(TRACE, "Invalid JSON in msg.key: %r, value: %r", msg.key, msg.value)
                     continue
 
                 value = None
                 if msg.value:
                     try:
                         value = ujson.loads(msg.value.decode("utf8"))
                     except ValueError:
-                        self._log.exception("Invalid JSON in msg.value: %r, key: %r", msg.value, msg.key)
+                        self._log.exception("Invalid JSON in msg.value")
```
juha-aiven, line 261: The same as above, maybe log the key or parts of it.

karapace/schema_reader.py
```diff
@@ -247,26 +249,29 @@ def handle_messages(self) -> None:
...
+                        self._log.log(TRACE, "Invalid JSON in msg.key: %r, value: %r", msg.key, msg.value)
                         continue
 
-                self._log.info("Read new record: key: %r, value: %r, offset: %r", key, value, msg.offset)
-                self.handle_msg(key, value)
-                self.offset = msg.offset
-                self._log.info(
-                    "Handled message, current offset: %r, ready: %r, add_offsets: %r",
```
juha-aiven, line 266: Has this message been intentionally removed? Maybe combine these logged values the the above logging statement.

karapace/schema_reader.py
```diff
@@ -337,29 +342,35 @@ def _handle_msg_schema(self, key: dict, value: Optional[dict]) -> None:
...
+            schema_str=schema_str,
+        )
+        schema = {
             "schema": typed_schema,
             "version": schema_version,
             "id": schema_id,
             "deleted": schema_deleted,
         }
+
+        if schema_version in subjects_schemas:
```
juha-aiven, line 366: Particular reason why the location of this logging has been changed?

karapace/schema_reader.py
```diff
@@ -337,29 +342,35 @@ def _handle_msg_schema(self, key: dict, value: Optional[dict]) -> None:
...
+            self._log.info("Updating entry subject: %r version: %r id: %r", schema_subject, schema_version, schema_id)
+        else:
+            self._log.info("Adding entry subject: %r version: %r id: %r", schema_subject, schema_version, schema_id)
+
+        self._log.log(TRACE, "Got schema %r", schema)
```
juha-aiven, line 371: Any reason why before `typed_schema` was logged and now `schema`? Maybe because `schema` has more info?

karapace/statsd.py
```diff
@@ -125,8 +125,8 @@ def _send(self, metric: str, metric_type: bytes, value: Any, tags: Optional[Dict
                 parts.insert(1, ",{}={}".format(tag, tag_value).encode("utf-8"))
 
             self._socket.sendto(b"".join(parts), self._dest_addr)
-        except Exception as ex:  # pylint: disable=broad-except
-            self.log.error("Unexpected exception in statsd send: %s: %s", ex.__class__.__name__, ex)
```
juha-aiven, line 129: Why not log the `Exception`?

karapace/compatibility/jsonschema/checks.py
```diff
@@ -248,7 +245,6 @@ def compatibility_rec(
     # reader has type `array` to represent a list, and the writer is either a
     # different type or it is also an `array` but now it representes a tuple.
     if reader_schema is None and writer_schema is not None:
-        LOG.debug("Schema removed reader_schema.type='%r'", get_type_of(reader_schema))
```
juha-aiven, line 251: Is this really unnecessary? Maybe leave, but without parameters. It doesn't look like something that would fill up logs.

juha-aiven:
The PR is so big that it'd be worth mentioning in the description that it does other things too than just those mentioned, e.g. it makes logger instances private.

karapace/config.py
```diff
@@ -51,7 +51,7 @@
     "ssl_password": None,
     "sasl_mechanism": None,
     "sasl_plain_username": None,
-    "sasl_plain_password": None,
+    SASL_PLAIN_PASSWORD: None,
```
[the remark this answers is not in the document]
hackaugusto: what do you mean? `if 'password' in key'` or something like that?

I don't see the point of using a constant that looks out-of-place being bad at all. Sensitive data is usually the exception, and having it look weird can be a good thing to caught people's attention.

karapace/karapace.py
```diff
@@ -11,20 +11,18 @@
...
-import logging
-
 
 class KarapaceBase(RestApp):
     def __init__(self, config: Config) -> None:
         super().__init__(app_name="karapace", config=config)
 
         self.kafka_timeout = 10
         self.route("/", callback=self.root_get, method="GET")
-        self.log = logging.getLogger("Karapace")
```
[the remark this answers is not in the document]
hackaugusto: Yep, that was intentional. This is not a stable api, the change doesn't make it harder to consume logs, and the existing code was super convoluted, where this log instance was overwriting the one from RestApp, and then both `KarapaceAll` and `KafkaRest` overwrote it too, but for some reason `KarapaceSchemaRegistry` did not. So in the end the name of the logger would change based on the flags provided in the configuration file, which doesn't seem to be useful for anything.

karapace/client.py
```diff
@@ -74,7 +74,7 @@ async def close(self) -> None:
             if self._client is not None:
                 await self._client.close()
         except:  # pylint: disable=bare-except
-            log.info("Could not close client")
+            self._log.error("Could not close client")
```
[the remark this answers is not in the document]
hackaugusto: would mentioning it make the commit any better? I'm adding it there anyways but the change is changing `log.info` to `self._log.error` and the text will be `changed log from info to error`

karapace/karapace_all.py
```diff
@@ -13,9 +13,7 @@
 
 
 class KarapaceAll(KafkaRest, KarapaceSchemaRegistry):
-    def __init__(self, config: Config) -> None:
-        super().__init__(config=config)
-        self.log = logging.getLogger("KarapaceAll")
```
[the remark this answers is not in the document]
hackaugusto: same as https://github.com/aiven/karapace/pull/384#discussion_r844307880

karapace/master_coordinator.py
```diff
@@ -119,13 +119,11 @@ def _on_join_complete(self, generation, member_id, protocol, member_assignment_b
         else:
             self.master_url = master_url
             self.are_we_master = False
-        # pylint: disable=super-with-arguments
-        return super(SchemaCoordinator, self)._on_join_complete(generation, member_id, protocol, member_assignment_bytes)
+        return super()._on_join_complete(generation, member_id, protocol, member_assignment_bytes)
```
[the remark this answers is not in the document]
hackaugusto: honestly I don't even remember anymore, gut feeling says the code was right.

karapace/schema_reader.py
hackaugusto: > Is the idea to hide the values or to have less logging?

I removed anything that is private data from the default log level, also the data you asked for is in the next line when verbose logging is used.

> Now this message has no context of what's wrong

That is what the exception message is for:

```
>>> import ujson
>>> import logging
>>> try:
...   ujson.loads('this is a test')
... except:
...   logging.exception('invalid json')
... 
ERROR:root:invalid json
  File "<stdin>", line 2, in <module>
ValueError: Unexpected character found when decoding 'true'
```

karapace/schema_reader.py
hackaugusto: same answer, private data is removed. debugging data is available in the next line. context is available in the exception message.

karapace/schema_reader.py
hackaugusto: yes, this was intentionally removed, the previous line was the dispatch to the handlers, they extensive logging with additional context based on the message type

karapace/schema_reader.py
hackaugusto: because of what you said https://github.com/aiven/karapace/pull/384#discussion_r842390928

karapace/schema_reader.py
hackaugusto: more info

karapace/statsd.py
hackaugusto: the change is from `error` to `exception`. That is called `exception` because the exception is included.

karapace/schema_reader.py
hackaugusto: Any other arguments? That was intentional as trace is more verbose than debug.

karapace/schema_registry_apis.py
```diff
@@ -643,7 +644,7 @@ async def subject_versions_list(self, content_type, *, subject):
         self.r(list(subject_data["schemas"]), content_type, status=HTTPStatus.OK)
 
     async def get_master(self) -> Tuple[bool, Optional[str]]:
-        async with self.master_lock:
+        async with self._master_lock:
```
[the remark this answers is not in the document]
hackaugusto: I thought that was easy enough to review because of the size. Here is the new PR: https://github.com/aiven/karapace/pull/385

karapace/compatibility/jsonschema/checks.py
hackaugusto: - This is a recursive call, number of log lines is undetermined
- The same information is encoded in the object instantiate in the next line with additional context (namely where in the recursion is the incompatibility), and that is logged by the caller of this function

hackaugusto:
> The PR is so big that it'd be worth mentioning in the description that it does other things too than just those mentioned, e.g. it makes logger instances private.

Added `makes logger instances private` to the PR description

karapace/kafka_rest_apis/admin.py
```diff
@@ -13,7 +13,7 @@
 class KafkaRestAdminClient(KafkaAdminClient):
     def __init__(self, **configs):
         super().__init__(**configs)
-        self.log = logging.getLogger("AdminClient")
+        self._log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
jjaakola-aiven: Maybe the logging could be removed in this file, it is used only to debug log retries.

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -30,14 +30,14 @@ class ConsumerManager:
...
-        self.log = logging.getLogger("RestConsumerManager")
+        self._log = logging.getLogger(__name__)
         self.deserializer = SchemaRegistryDeserializer(config=config)
         self.consumers = {}
         self.consumer_locks = defaultdict(Lock)
 
     def new_name(self) -> str:
         name = str(uuid.uuid4())
-        self.log.debug("Generated new consumer name: %s", name)
+        self._log.debug("Generated new consumer name: %s", name)
```
jjaakola-aiven, line 40: Is this debug log line necessary?

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -155,12 +155,12 @@ def _update_partition_assignments(consumer: KafkaConsumer):
     # CONSUMER
     async def create_consumer(self, group_name: str, request_data: dict, content_type: str):
         group_name = group_name.strip("/")
-        self.log.info("Create consumer request for group  %s", group_name)
+        self._log.info("Create consumer request for group  %s", group_name)
```
jjaakola-aiven, line 158: I would combine the return value of `new_name` to the log. Also consider if this is debug log instead of info.
The group and name is logged as info also on the line 173.

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -155,12 +155,12 @@ def _update_partition_assignments(consumer: KafkaConsumer):
...
         consumer_name = request_data.get("name") or self.new_name()
         internal_name = self.create_internal_name(group_name, consumer_name)
         async with self.consumer_locks[internal_name]:
             if internal_name in self.consumers:
-                self.log.error("Error creating duplicate consumer in group %s with id %s", group_name, consumer_name)
+                self._log.error("Error creating duplicate consumer in group %s with id %s", group_name, consumer_name)
```
jjaakola-aiven, line 163: This would be a `warn` instead of error.

jjaakola-aiven (commented):
IMO the debug log calls could be removed totally.
Info logs also do much of tracing of the normal operation flow and I propose that the default log level should be set to `WARNING`.

karapace/kafka_rest_apis/admin.py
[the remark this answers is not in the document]
hackaugusto: moved all the loggers to the module.

karapace/schema_reader.py
hackaugusto: Removed all TRACE calls.

karapace/utils.py
hackaugusto: All loggers are now module scoped.

karapace/client.py
```diff
@@ -65,6 +64,7 @@ def __init__(
             self.ssl_mode = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
             self.ssl_mode.load_verify_locations(cafile=server_ca)
         self._client: Optional[ClientSession] = None
+        self._log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
hackaugusto: Most modules are now at the module level.

karapace/config.py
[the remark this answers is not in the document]
tvainika: Something like that was my point, but resolving this as matter of style, and differing opinions, not rules

karapace/config.py
```diff
@@ -108,12 +109,20 @@ def set_settings_from_environment(config: Config) -> None:
...
         env_val = os.environ.get(env_name)
         if env_val is not None:
-            LOG.debug(
-                "Populating config value %r from env var %r with %r instead of config file",
-                config_name,
-                env_name,
-                env_val,
-            )
+            if config_name not in SECRET_CONFIG_OPTIONS:
+                LOG.debug(
```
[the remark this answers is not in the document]
hackaugusto: For bug reports?

karapace/kafka_rest_apis/consumer_manager.py
hackaugusto: done

karapace/config.py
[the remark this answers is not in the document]
jjaakola-aiven: Yes, and in my opinion info level is the correct for printing the startup configuration.

jjaakola-aiven (commented):
LGTM

tvainika (approved):
LGTM, but pending merge until another pr merged as per pr description

PR #388, opened 2022-04-08
Schema reader loop improvements

hackaugusto:
Review after https://github.com/aiven/karapace/pull/391 (It fix port allocation for integration tests)

These are the main changes here:

- If creating the `Consumer` fails, the loop should wait a bit, just like it was done for the `AdminClient`.
   - The motivation for this was a hot loop where Karapace was constantly failing to run inside docker, because it could not resolve the hostname. I'm not sure what caused that bug, but I wanted to remove the hot loop (I have the feeling the hotloop is inside kafka-python library, but this one should be fixed anyways)
- The `close` of the consumer should be called even if the `close` of admin client fails
- Instead of using a flag now it uses an Event to stop the thread. It will be faster to stop an uninitialized thread that is doing a retry.

Also:

- I moved the instantiation out of the schema reader class. I tried to consolidate this before, but it is a lot of changes to get all the clients created by the same util (there are some inconsistencies in our flags because of this). This is just a small change that should make it easier to share utils to instantiate the clients from the config.
- This is a bit of ground work to isolate the thread initialization from the processing of messages. My idea is to move the complete state and the handling of message to a separate class, so that we can unit test it without using threads which is painful to use with a debugger and to properly assert on.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -58,6 +58,21 @@ def _create_consumer_from_config(config: Config) -> KafkaConsumer:
     )
 
 
+def _create_admin_client_from_config(config: Config) -> KafkaAdminClient:
+    return KafkaAdminClient(
+        api_version_auto_timeout_ms=constants.API_VERSION_AUTO_TIMEOUT_MS,
+        bootstrap_servers=config["bootstrap_uri"],
+        client_id=config["client_id"],
+        security_protocol=config["security_protocol"],
+        ssl_cafile=config["ssl_cafile"],
+        ssl_certfile=config["ssl_certfile"],
+        ssl_keyfile=config["ssl_keyfile"],
+        sasl_mechanism=config["sasl_mechanism"],
+        sasl_plain_username=config["sasl_plain_username"],
+        sasl_plain_password=config["sasl_plain_password"],
+    )
+
+
 class OffsetsWatcher:
     """Synchronization container for threads to wait until an offset is seen.
 
@@ -140,18 +155,7 @@ class KafkaSchemaReader(Thread):
 
     def init_admin_client(self) -> bool:
         try:
-            self.admin_client = KafkaAdminClient(
-                api_version_auto_timeout_ms=constants.API_VERSION_AUTO_TIMEOUT_MS,
-                bootstrap_servers=self.config["bootstrap_uri"],
-                client_id=self.config["client_id"],
-                security_protocol=self.config["security_protocol"],
-                ssl_cafile=self.config["ssl_cafile"],
-                ssl_certfile=self.config["ssl_certfile"],
-                ssl_keyfile=self.config["ssl_keyfile"],
-                sasl_mechanism=self.config["sasl_mechanism"],
-                sasl_plain_username=self.config["sasl_plain_username"],
-                sasl_plain_password=self.config["sasl_plain_password"],
-            )
+            self.admin_client = _create_admin_client_from_config(self.config)
             return True
         except (NodeNotReadyError, NoBrokersAvailable, AssertionError):
             LOG.warning("No Brokers available yet, retrying init_admin_client()")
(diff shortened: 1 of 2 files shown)


karapace/schema_reader.py
```diff
@@ -199,33 +175,62 @@ def get_schema_id(self, new_schema: TypedSchema) -> int:
...
-            except Exception as e:  # pylint: disable=broad-except
-                if self.stats:
+        with ExitStack() as stack:
+            while not self._stop.is_set() and self.admin_client is None:
+                try:
+                    self.admin_client = _create_admin_client_from_config(self.config)
+                    stack.enter_context(closing(self.admin_client))
+                except (NodeNotReadyError, NoBrokersAvailable, AssertionError):
+                    self.log.warning("[Admin Client] No Brokers available yet. Retrying")
+                    self._stop.wait(timeout=2.0)
```
HelenMel, line 188: I would rather make a `2.0` configurable property.

HelenMel:
@hackaugusto I agree with all stated points, but I would recommend to make a waiting time configurable.

karapace/schema_reader.py
hackaugusto: I thought about this, but what would be the benefit of making this configurable?

jjaakola-aiven: This `2.0` and the other `5.0` second timeout could be defined as constants.
```suggestion
                    self._stop.wait(timeout=CLIENT_CREATION_TIMEOUT_SECONDS)
                    # And for the topic creation TOPIC_CREATION_TIMEOUT_SECONDS
```

jjaakola-aiven:
```
KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS = 2.0
SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS = 5.0
```

HelenMel: > I thought about this, but what would be the benefit of making this configurable?

If I understand correctly, this number is just guesstimate, right? So we could tune it a bit later

hackaugusto: > If I understand correctly, this number is just guesstimate, right? So we could tune it a bit later

I don't know, I'm just moving the code around, these constants were already there. I understand we can tune the value, but I'm not sure if we need to make that a configuration setting, I think we should make configuration only values that we think are need to be tuned in production.

karapace/schema_reader.py
```diff
@@ -199,33 +178,62 @@ def get_schema_id(self, new_schema: TypedSchema) -> int:
...
+                except (NodeNotReadyError, NoBrokersAvailable, AssertionError):
+                    self.log.warning("[Admin Client] No Brokers available yet. Retrying")
+                    self._stop.wait(timeout=KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS)
+                except KafkaConfigurationError:
+                    self.log.exception("[Admin Client] Invalid configuration. Bailing")
+                    raise
+                except Exception as e:  # pylint: disable=broad-except
+                    self.log.exception("[Admin Client] Unexpected exception. Retrying")
+                    self.stats.unexpected_exception(ex=e, where="admin_client_instantiation")
+                    self._stop.wait(timeout=2.0)
```
jjaakola-aiven, line 198:
```suggestion
                    self._stop.wait(timeout=KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS)
```

karapace/schema_reader.py
```diff
@@ -199,33 +178,62 @@ def get_schema_id(self, new_schema: TypedSchema) -> int:
...
+
+            while not self._stop.is_set() and self.consumer is None:
+                try:
+                    self.consumer = _create_consumer_from_config(self.config)
+                    stack.enter_context(closing(self.consumer))
+                except (NodeNotReadyError, NoBrokersAvailable, AssertionError):
+                    self.log.warning("[Consumer] No Brokers available yet. Retrying")
+                    self._stop.wait(timeout=2.0)
```
jjaakola-aiven, line 206:
```suggestion
                    self._stop.wait(timeout=KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS)
```

tvainika (approved):

PR #389, opened 2022-04-08
Cleanup rest app

hackaugusto:
This PR has a few varied changes for issues I found while fixing the tests.

The biggest change is to remove the dead code from RestApp. It currently supports SOCKS requests, which is never used, and it contains two sessions, one with SSL validation and another without, only the one with validation is used.

I'm not sure if this was introduced while working on a feature but it was not finished. For now I'm removing it because I'm seeing SSL validation errors, and that reduces the number of things I need to check to fix that.

There is also some other small changes, for issues I found while debugging.

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -41,6 +41,7 @@ DEFAULTS = {
     "registry_port": 8081,
     "registry_ca": None,
     "log_level": "DEBUG",
+    "log_format": "%(name)-20s\t%(threadName)s\t%(levelname)-8s\t%(message)s",
     "master_eligibility": True,
     "replication_factor": 1,
     "security_protocol": "PLAINTEXT",
@@ -66,7 +67,6 @@ DEFAULTS = {
     "master_election_strategy": "lowest",
     "protobuf_runtime_directory": "runtime",
 }
-DEFAULT_LOG_FORMAT_JOURNAL = "%(name)-20s\t%(threadName)s\t%(levelname)-8s\t%(message)s"
 
 
 class InvalidConfiguration(Exception):
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -22,6 +22,7 @@ HOSTNAME = socket.gethostname()
 
 DEFAULTS = {
     "access_logs_debug": False,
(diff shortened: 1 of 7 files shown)


karapace/config.py
```diff
@@ -40,6 +41,7 @@
     "registry_port": 8081,
     "registry_ca": None,
     "log_level": "DEBUG",
+    "log_format": "%(name)-20s\t%(threadName)s\t%(levelname)-8s\t%(message)s",
```
hackaugusto, line 44: This is preparation work, another PR uses this to configure Karapace's logs.

tests/integration/utils/process.py
```diff
@@ -37,8 +37,12 @@ def wait_for_port_subprocess(
 
 def stop_process(proc: Optional[Popen]) -> None:
     if proc:
-        os.kill(proc.pid, signal.SIGKILL)
-        proc.wait(timeout=10.0)
+        try:
+            os.kill(proc.pid, signal.SIGKILL)
+            proc.wait(timeout=10.0)
+        # ProcessLookupError: Raised when the process is already gone
+        except ProcessLookupError:
```
hackaugusto, line 44: I experience this while working on another PR. I'm not 100% sure what was the race, one hypothesis is that the kill was fast enough that wait failed because the process was gone.

karapace/config.py
hackaugusto: This is used here:

https://github.com/aiven/karapace/pull/391/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R108-R111

It is useful for our CI, since we don't have systemd-journald adding the timestamps in that case.

tests/integration/utils/process.py
jjaakola-aiven: Why not use `Popen.kill`?

jjaakola-aiven (approved):
LGTM

tests/integration/utils/process.py
hackaugusto: No reason, it should be the same :)

PR #391, opened 2022-04-08
Fix flakiness

hackaugusto:
This probably doesn't fix all the issues, but it should handle the last spike in flakiness. The PR could definitely have better commits, I just ran out of time :P

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,7 @@ jobs:
       run: python3 -m pytest -s -vvv tests/unit/
 
     - name: Execute integration-tests
+      timeout-minutes: 10
       run: python3 -m pytest -s -vvv tests/integration/ --log-dir=/tmp/ci-logs --log-file=/tmp/ci-logs/pytest.log
 
     - name: Archive logs
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,7 @@ jobs:
       run: make karapace/version.py
 
     - name: Execute unit-tests
+      timeout-minutes: 2
       run: python3 -m pytest -s -vvv tests/unit/
 
     - name: Execute integration-tests
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -12,11 +12,10 @@ from dataclasses import asdict
 from filelock import FileLock
 from kafka import KafkaProducer
 from karapace.client import Client
-from karapace.config import set_config_defaults, write_config
+from karapace.config import Config, set_config_defaults, write_config
 from karapace.kafka_rest_apis import KafkaRest, KafkaRestAdminClient
-from karapace.schema_registry_apis import KarapaceSchemaRegistry
 from pathlib import Path
-from subprocess import Popen
+from tests.integration.utils.cluster import RegistryDescription, start_schema_registry_cluster
 from tests.integration.utils.config import KafkaConfig, KafkaDescription, ZKConfig
 from tests.integration.utils.kafka_server import (
     configure_and_start_kafka,
@@ -24,17 +23,19 @@ from tests.integration.utils.kafka_server import (
     maybe_download_kafka,
     wait_for_kafka,
 )
-from tests.integration.utils.network import get_random_port, KAFKA_PORT_RANGE, REGISTRY_PORT_RANGE, ZK_PORT_RANGE
+from tests.integration.utils.network import PortRangeInclusive
 from tests.integration.utils.process import stop_process, wait_for_port_subprocess
 from tests.integration.utils.synchronization import lock_path_for
 from tests.integration.utils.zookeeper import configure_and_start_zk
-from tests.utils import new_random_name, repeat_until_successful_request
-from typing import AsyncIterator, Iterator, Optional, Tuple
+from tests.utils import repeat_until_successful_request
+from typing import AsyncIterator, Iterator, List, Optional
 
 import asyncio
 import os
 import pathlib
 import pytest
+import re
+import string
 import time
 import ujson
 
@@ -46,6 +47,36 @@ KAFKA_SCALA_VERSION = "2.13"
 # stdout logger is disabled to keep the pytest report readable
 KAFKA_LOG4J = TEST_INTEGRATION_DIR / "config" / "log4j.properties"
 WORKER_COUNTER_KEY = "worker_counter"
+
+
+def _clear_test_name(name: str) -> str:
+    # Based on:
+    # https://github.com/pytest-dev/pytest/blob/238b25ffa9d4acbc7072ac3dd6d8240765643aed/src/_pytest/tmpdir.py#L189-L194
+    # The purpose is to return a similar name to make finding matching logs easier
+    return re.sub(r"[\W]", "_", name)[:30]
+
+
+@pytest.fixture(scope="session", name="port_range")
+def fixture_port_range() -> PortRangeInclusive:
+    """Container used by other fixtures to register used ports"""
+    # To find a good port range use the following:
+    #
+    #   curl --silent 'https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt' | \
+    #       egrep -i -e '^\s*[0-9]+-[0-9]+\s*unassigned' | \
+    #       awk '{print $1}'
+    #
+    start = 48700
+    end = 49000
+
+    # Split the ports among the workers to prevent port reuse
+    worker_name = os.environ.get("PYTEST_XDIST_WORKER", "0")
+    worker_id = int(worker_name.lstrip(string.ascii_letters))
+    worker_count = int(os.environ.get("PYTEST_XDIST_WORKER_COUNT", "1"))
+    total_ports = end - start
+    ports_per_worker = total_ports // worker_count
+    start_worker = (ports_per_worker * worker_id) + start
+    end_worker = start_worker + ports_per_worker - 1
+    return PortRangeInclusive(start_worker, end_worker)
 
 
 @pytest.fixture(scope="session", name="kafka_description")
@@ -70,6 +101,7 @@ def fixture_kafka_server(
     session_datadir: Path,
     session_logdir: Path,
     kafka_description: KafkaDescription,
+    port_range: PortRangeInclusive,
 ) -> Iterator[KafkaServers]:
     bootstrap_servers = request.config.getoption("kafka_bootstrap_servers")
 
@@ -88,20 +120,19 @@ def fixture_kafka_server(
     lock_file = lock_path_for(transfer_file)
 
     with ExitStack() as stack:
+        zk_client_port = stack.enter_context(port_range.allocate_port())
+        zk_admin_port = stack.enter_context(port_range.allocate_port())
+        kafka_plaintext_port = stack.enter_context(port_range.allocate_port())
+
         with FileLock(str(lock_file)):
             if transfer_file.exists():
                 config_data = ujson.loads(transfer_file.read_text())
                 zk_config = ZKConfig.from_dict(config_data["zookeeper"])
                 kafka_config = KafkaConfig.from_dict(config_data["kafka"])
-
-                # Count the new worker
-                config_data[WORKER_COUNTER_KEY] += 1
-                transfer_file.write_text(ujson.dumps(config_data))
+                config_data[WORKER_COUNTER_KEY] += 1  # Count the new worker
             else:
                 maybe_download_kafka(kafka_description)
 
-                zk_client_port = get_random_port(port_range=ZK_PORT_RANGE, blacklist=[])
-                zk_admin_port = get_random_port(port_range=ZK_PORT_RANGE, blacklist=[zk_client_port])
                 zk_config = ZKConfig(
                     client_port=zk_client_port,
                     admin_port=zk_admin_port,
@@ -114,7 +145,6 @@ def fixture_kafka_server(
                 # Make sure zookeeper is running before trying to start Kafka
                 wait_for_port_subprocess(zk_config.client_port, zk_proc, wait_time=20)
 
-                kafka_plaintext_port = get_random_port(port_range=KAFKA_PORT_RANGE, blacklist=[])
                 data_dir = session_datadir / "kafka"
                 log_dir = session_logdir / "kafka"
                 data_dir.mkdir(parents=True)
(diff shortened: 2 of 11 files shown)


tests/integration/test_master_coordinator.py
```diff
@@ -44,82 +40,82 @@ def has_master(mc: MasterCoordinator) -> bool:
...
-        # Make sure the end configuration is as expected
-        master_url = f'http://{master.config["host"]}:{master.config["port"]}'
-        assert master.sc.election_strategy == strategy
-        assert slave.sc.election_strategy == strategy
-        assert master.sc.master_url == master_url
-        assert slave.sc.master_url == master_url
-
-
-def test_no_eligible_master(kafka_servers: KafkaServers) -> None:
+    with port_range.allocate_port() as port1, port_range.allocate_port() as port2:
```
hackaugusto, line 45: I didn't change the test, only added the context manager for the port allocation.

tests/integration/test_master_coordinator.py
```diff
@@ -44,82 +40,82 @@ def has_master(mc: MasterCoordinator) -> bool:
...
-
-    with closing(init_admin(config_aa)) as mc:
-        # Wait for the election to happen, ie. flag is not None
-        while not mc.sc or mc.sc.are_we_master is None:
-            time.sleep(0.3)
-
-        # Make sure the end configuration is as expected
-        assert mc.sc.are_we_master is False
-        assert mc.sc.master_url is None
+    with port_range.allocate_port() as port:
```
hackaugusto, line 99: same here, no changes to the test, just port allocation

tests/integration/test_schema.py
```diff
@@ -863,12 +863,12 @@ async def assert_schema_versions(client: Client, trail: str, schema_id: int, exp
...
     assert res.status_code == 200
+    registered_schemas = res.json()
 
     # Schema Registry doesn't return an ordered list, Karapace does.
     # Need to check equality ignoring ordering.
-    assert len(res.json()) == len(expected)
-    for e in ({"subject": e[0], "version": e[1]} for e in expected):
-        assert e in res.json()
+    result = [(schema["subject"], schema["version"]) for schema in registered_schemas]
+    assert set(result) == set(expected)
```
hackaugusto, line 871: this is not a fix. Just a better debugging assert (the previous would fail in the length check without giving any useful info on what was different)

tests/integration/utils/network.py
```diff
@@ -29,41 +30,33 @@ def next_range(self, number_of_ports: int) -> "PortRangeInclusive":
 
         return PortRangeInclusive(next_start, next_end)
 
+    @contextmanager
+    def allocate_port(self) -> int:
```
hackaugusto, line 34: I decided to use a context manager because we need to release the ports back to the pool of available ones, otherwise we run out of ports to use

tests/utils.py
```diff
@@ -175,6 +175,8 @@ async def new_consumer(c, group, fmt="avro", trail=""):
 
 
 def new_random_name(prefix: str) -> str:
+    # A hyphen is not a valid character for Avro identifiers. Use only the
+    # first 8 characters of the UUID.
     suffix = str(uuid.uuid4())[:8]
```
hackaugusto, line 180: Just a bit of documentation. It is odd at first glance that this is being done (should probably do something else, but I had enough changes on this PR)

tests/integration/conftest.py
```diff
@@ -48,6 +49,127 @@
...
+
+
+@dataclass(frozen=True)
+class RegistryDescription:
+    endpoint: RegistryEndpoint
+    schemas_topic: str
+
+
+@asynccontextmanager
+async def start_schema_registry_cluster(
```
hackaugusto, line 76: This is the main fix. It will start a cluster with a new schemas topic and a new group id, which prevents most of the issues with flakiness. I also tried to be careful to now leak any servers behind, as that would cause subsequent runs to fail too

tests/integration/conftest.py
```diff
@@ -88,56 +211,59 @@ def fixture_kafka_server(
...
-                    "kafka": asdict(kafka_config),
-                    WORKER_COUNTER_KEY: 1,
-                }
-                transfer_file.write_text(ujson.dumps(config_data))
+        # must be before the transfer_file is checked
+        stack.enter_context(FileLock(str(lock_file)))
+
+        zk_client_port = stack.enter_context(port_range.allocate_port())
+        zk_admin_port = stack.enter_context(port_range.allocate_port())
+        kafka_plaintext_port = stack.enter_context(port_range.allocate_port())
```
hackaugusto, line 219: Didn't change much the code below, just fixed the usage of the ports.

tests/integration/utils/network.py
```diff
@@ -29,41 +30,33 @@ def next_range(self, number_of_ports: int) -> "PortRangeInclusive":
...
+            raise RuntimeError(f"no free ports available. Range {self}, allocated: {self.allocated}")
+
+        port = random.randint(self.start, self.end)
+        while port in self.allocated:
+            port = random.randint(self.start, self.end)
+        self.allocated.add(port)
+
+        yield port
+
+        self.allocated.remove(port)
```
hackaugusto, line 59: I suppose in this case this is the best approach, if an exception is thrown on the yield something went wrong, and we can't be sure the port was properly released

tests/integration/utils/network.py
jjaakola-aiven: Could the getting free port use binding to port 0, set it reusable and return the number?
Something like
```python
import socket
free_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
free_socket.bind('localhost', 0)
free_socket(socket.SOL_SOCKET, socket.SO_REUSE_ADDR, 1)
free_socket.getsockname() # returns tuple (addr, socket)
```

hackaugusto: I tried that in the past (different project), and it was a flaky solution. Behavior of REUSE_ADDR and REUSE_PORT is not compatible across OSs, and IIRC the application also needs to set the flag, otherwise we have issues with lingering sockets. but if you want we can give a try.

.github/workflows/tests.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       run: python3 -m pytest -s -vvv tests/unit/
 
     - name: Execute integration-tests
+      timeout-minutes: 10
```
hackaugusto, line 41: I introduced a bug on the fixtures that made the tests run forever, without this the build runs for hours:

image

tests/integration/conftest.py
```diff
@@ -48,6 +49,36 @@
...
+    #
+    #   curl --silent 'https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt' | \
+    #       egrep -i -e '^\s*[0-9]+-[0-9]+\s*unassigned' | \
+    #       awk '{print $1}'
+    #
+    start = 48700
+    end = 49000
+
+    # Split the ports among the workers to prevent port reuse
+    worker_name = os.environ.get("PYTEST_XDIST_WORKER", "0")
```
hackaugusto, line 72: The default is for runs with `-n0`

tests/integration/test_schema_backup.py
```diff
@@ -73,7 +75,12 @@ async def test_backup_restore(
             fp,
         )
 
-    config = set_config_defaults({"bootstrap_uri": kafka_servers.bootstrap_servers})
```
hackaugusto, line 76: This was a source of flakiness, the backup test would overwrite the `_schemas` topic used by other tests. With these changes each test gets its own topic so this doesn't happen.

tests/integration/utils/network.py
```diff
@@ -2,68 +2,71 @@
...
+            get the same ports for *a single test run*.
+
+            Because of that, the port range should be chosen such that there is no
+            system service in the range. Also note that running two sessions of the
+            tests with the same range is not supported and will lead to flakiness.
+        """
+        if len(self._maybe_available) == 0:
+            raise RuntimeError(f"No free ports available. start: {self.start} end: {self.end}")
+
+        filtered_ports = ((pos, port) for pos, port in enumerate(self._maybe_available) if not is_time_wait(port))
```
hackaugusto, line 55: This was kinda annoying, at first I assumed that 300 sockets would be good enough, but then tests failed with `address already in use` errors. I thought it could be an issue with a port outside of the test run, so I added a check using `psutil` to make sure the port was not used, turn out the socket was on TIME_WAIT state. So I tried to change this to a FIFO queue, to give time for the socket to close, but that was not sufficient for every socket. So I changed it to search for the first socket that is not on TIME_WAIT state.

Now, the obvious suggestion here would be to do what @jjaakola-aiven mentioned above. I generally find _both_ of these solutions bad. But let me know what you think.

hackaugusto:
I did a few local runs:

image

tests/integration/conftest.py
```diff
@@ -137,30 +167,32 @@ def fixture_kafka_server(
                     "kafka": asdict(kafka_config),
                     WORKER_COUNTER_KEY: 1,
                 }
-                transfer_file.write_text(ujson.dumps(config_data))
-
-        # Make sure every test worker can communicate with kafka
-        kafka_servers = KafkaServers(bootstrap_servers=[f"127.0.0.1:{kafka_config.plaintext_port}"])
-        wait_for_kafka(kafka_servers, KAFKA_WAIT_TIMEOUT)
```
hackaugusto, line 144: I didn't change Kafka's config here, but maybe it makes sense to remove the `auto_create` flag, there may be some tests still using the default `_schemas` topic which would cause flakiness.

tests/integration/utils/network.py
jjaakola-aiven: I think we can assume that sockets are BSD enough. And also discard Windows behavior SO_REUSEADDR as an issue here, these are used only for tests. I think the idiom I propose is fairly standard practice to get free port.

.github/workflows/tests.yml
jjaakola-aiven: This should be also on the unit tests.

tests/integration/utils/network.py
hackaugusto: I was not referring to differences between BSD or Windows, but Linux and BSD. It seems that this is working ATM, I would prefer to skip that rabbit hole for now, that can be done on a follow up.

hackaugusto:
The CI still has some issues that this PR is not fixing:

```
[gw1] ERROR tests/integration/test_client_protobuf.py::test_remote_client_protobuf2[pyloop] 
INTERNALERROR> def worker_internal_error(self, node, formatted_error):
INTERNALERROR>         """
INTERNALERROR>         pytest_internalerror() was called on the worker.
INTERNALERROR>     
INTERNALERROR>         pytest_internalerror() arguments are an excinfo and an excrepr, which can't
INTERNALERROR>         be serialized, so we go with a poor man's solution of raising an exception
INTERNALERROR>         here ourselves using the formatted message.
INTERNALERROR>         """
INTERNALERROR>         self._active_nodes.remove(node)
INTERNALERROR>         try:
INTERNALERROR> >           assert False, formatted_error
INTERNALERROR> E           AssertionError: Traceback (most recent call last):
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/_pytest/main.py", line 269, in wrap_session
INTERNALERROR> E                 session.exitstatus = doit(config, session) or 0
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/_pytest/main.py", line 323, in _main
INTERNALERROR> E                 config.hook.pytest_runtestloop(session=session)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR> E                 return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR> E                 return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR> E                 return outcome.get_result()
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR> E                 raise ex[1].with_traceback(ex[2])
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_callers.py", line 39, in _multicall
INTERNALERROR> E                 res = hook_impl.function(*args)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/xdist/remote.py", line 72, in pytest_runtestloop
INTERNALERROR> E                 self.run_one_test(torun)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/xdist/remote.py", line 89, in run_one_test
INTERNALERROR> E                 self.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR> E                 return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR> E                 return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR> E                 return outcome.get_result()
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR> E                 raise ex[1].with_traceback(ex[2])
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pluggy/_callers.py", line 34, in _multicall
INTERNALERROR> E                 next(gen)  # first yield
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pytest_timeout.py", line 104, in pytest_runtest_protocol
INTERNALERROR> E                 timeout_setup(item)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pytest_timeout.py", line 207, in timeout_setup
INTERNALERROR> E                 signal.signal(signal.SIGALRM, handler)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/signal.py", line 47, in signal
INTERNALERROR> E                 handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
INTERNALERROR> E             ValueError: signal only works in main thread
INTERNALERROR> E           assert False
INTERNALERROR> 
INTERNALERROR> /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/xdist/dsession.py:187: AssertionError
```

.github/workflows/tests.yml
hackaugusto: done

tvainika (approved):
LGTM

PR #395, opened 2022-04-20
Correct schema comparison when posting to subject

jjaakola-aiven:

After refactoring schema storage to use non-parsed versions an
error was introduced to schema comparison when posting to subject.
The original Avro schema string can differ from the parsed schema, e.g.
names are not necessary for comparison. Example below:

Stored schema string: {"type":"int","name":"example_name"}
-> parsed schema: "int"

New schema posted to subject: {"type":"int"}
-> parsed new schema: "int"

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -1526,6 +1526,41 @@ async def test_schema_same_subject(registry_async_client: Client, trail: str) ->
     assert json == {"id": schema_id, "subject": subject, "schema": ujson.loads(schema_str), "version": 1}
 
 
+async def test_schema_same_subject_unnamed(registry_async_client: Client) -> None:
+    """
+    The same schema JSON should be returned when checking the same schema str against the same subject
+    """
+    subject_name_factory = create_subject_name_factory("test_schema_same_subject_unnamed")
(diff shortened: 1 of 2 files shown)


tests/integration/test_schema.py
```diff
@@ -1526,6 +1526,41 @@ async def test_schema_same_subject(registry_async_client: Client, trail: str) ->
     assert json == {"id": schema_id, "subject": subject, "schema": ujson.loads(schema_str), "version": 1}
 
 
+async def test_schema_same_subject_unnamed(registry_async_client: Client) -> None:
+    """
+    The same schema JSON should be returned when checking the same schema str against the same subject
```
hackaugusto, line 1531: This could be a bit better, e.g.:

```
Avro equality doesn't depend on the value of `name`.
```

tvainika (approved):

tests/integration/test_schema.py
tvainika: There seems to be implicit assumption about "schema" is "avro schema" unless otherwise mentioned. E.g. filename is test_schema, as avro schemas where first, and only later jsonschema and protobuf was added. Maybe some bigger review of these could be done later?

PR #396, opened 2022-04-20
performance: added schema reused based on content

hackaugusto:
for usage patterns where the same instance is registered regularly, the
increase in memory pressure is measurable and can cause Karapace to be
OOM'ed. This commits improves the situation by reusing existing schemas
based on their content, instead of having multiple copies of the same
schema in memory.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -167,6 +167,13 @@ class KafkaSchemaReader(Thread):
         # set by another thread (e.g. `KarapaceSchemaRegistry`)
         self._stop = Event()
 
+        # Content based deduplication of schemas. This is used to reduce memory
+        # usage when the same schema is produce multiple times to the same or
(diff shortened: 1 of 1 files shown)


tvainika (commented):
If a schema is deleted via API, it is still left into `_hash_to_schema` dictionary, right? So basically leaking memory there even if saving much memory in other cases...

hackaugusto:
@tvainika unfortunately yes, we had an issue for this https://github.com/aiven/karapace/issues/283 ,  which I tried to fix with https://github.com/aiven/karapace/pull/284 , but it was reverted with some other modifications by https://github.com/aiven/karapace/pull/295 due to issues with duplicated global IDs.

This PR doesn't try to fix the memory leak, just optimize the memory usage.

tvainika (approved):

Issue #346, opened 2022-03-02
Add support for advertised_port to Karapace Rest Proxy


YuWang24:
# What is currently missing?

I am currently using a reverse proxy server (Nginx) in front of Karapace Rest Proxy. I am using the docker environment, and Nginx and Karapace are listening on different ports. When creating a consumer using Karapace:

```
curl -XPOST https://<domain>:<nginx-port>/consumers/consumer1 -u <username>:<password> \
--data '{"name": "my_consumer_instance", "format": "json", "auto.offset.reset": "earliest"}' \
--header "Content-Type: application/vnd.kafka.json.v2+json
```
The `base_uri` in response just gives me the karapace's port rather than the nginx port.
The response: 
`{"base_uri":"http://<ip>:<karapace-port>/consumers/consumer1/instances/my_consumer_instance","instance_id":"my_consumer_instance"}`

After checking the possible configs karapace has, it seems that there is not a config for `advertised_port`.

# How could this be improved?

Support of `advertised_port` could be added to the available configs to resolve such issues. When `advertised_port` is set, the response could display the `advertised_port` rather than `port`. By doing this, it can show the correct port in the consumer's response if Nginx is used along with Karapace. 

# Is this a feature you would work on yourself?

No

hackaugusto:
@YuWang24 thank you for opening the issue. I will add this to our internal black log and tag this as `good first issue`.

YuWang24:
Hey @hackaugusto, thanks for your reply. I am thinking doing it myself. I can add a boolean config `advertised_port` with default value of `None`. When it is set, this line [](https://github.com/aiven/karapace/blob/f6ac8f4b7c28be53eb7a6559ee258eb3b928eed4/karapace/kafka_rest_apis/consumer_manager.py#L32) will use `advertised_port` rather than `port`. 

Also, I found that the protocol in the `base_uri` response is using `http` as default, I am think of making this configurable if a secured proxy is used together with Karapace (for example my case). How about a config called `use_secured_proxy` or `behind_secured_proxy`? Default as `False`, when set `True`, this line [](https://github.com/aiven/karapace/blob/f6ac8f4b7c28be53eb7a6559ee258eb3b928eed4/karapace/kafka_rest_apis/consumer_manager.py#L32) will use `https` rather than `http`.

How do these sound?

tvainika:
Fixing PR has been merged.

PR #400, opened 2022-04-27
Add support for new configs including advertised_port and use_secured_proxy

YuWang24:
I am currently using a reverse proxy server (Nginx) in front of Karapace Rest Proxy. I am using the docker environment, and Nginx and Karapace are listening on different ports. When creating a consumer using Karapace:

```
curl -XPOST https://<domain>:<nginx-port>/consumers/consumer1 -u <username>:<password> \
--data '{"name": "my_consumer_instance", "format": "json", "auto.offset.reset": "earliest"}' \
--header "Content-Type: application/vnd.kafka.json.v2+json
```

The `base_uri` in response just gives me the karapace's port rather than the nginx port.
The response:
`{"base_uri":"http://<ip>:<karapace-port>/consumers/consumer1/instances/my_consumer_instance","instance_id":"my_consumer_instance"}`

Also, the protocol in the `base_uri` response is using `http` as default and cannot be changed because there are no configs to change that.

This PR adds support for new configs including `advertised_port` and `use_secured_proxy` to solve these problems.

References: #346

# Why this way

Adding these two configs makes the `port` and `protocol` returned in the `base_uri` response configurable, users can set these configs if they are using a reverse proxy like me.


karapace/config.py
```diff
@@ -35,6 +35,8 @@
     "group_id": "schema-registry",
     "host": "127.0.0.1",
     "port": 8081,
+    "advertised_port": None,
+    "use_secured_proxy": False,
```
hackaugusto, line 39: nitpick: maybe name this `advertised_protocol`.
request: Could you please add an `advertised_host` too?

YuWang24: Thanks for your reply @hackaugusto. Yeah, I can change the name to `advertised_protocol`. But for `advertised_host`, there is already a config called `advertised_hostname`, please see (https://github.com/aiven/karapace/blob/eb5a7720475b66de5011722db4799e3815fdacaa/karapace/config.py#L26). Is `advertised_host` still required if there is an `advertised_hostname`?

tvainika: Existing `advertised_hostname` is fine, but maybe sort those such that all three `advertised_` are together in the list. Also now `use_secure_proxy`was renamed to `advertised_protocol` its type and logic need to change from boolean True/False to a string (`http` by default)

YuWang24: Thanks @tvainika, changes are done. Please review it again, thanks a lot.

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -34,7 +34,11 @@ def new_name() -> str:
 class ConsumerManager:
     def __init__(self, config: dict) -> None:
         self.config = config
-        self.hostname = f"http://{self.config['advertised_hostname']}:{self.config['port']}"
+        self.protocol = self.config["advertised_protocol"]
```
hackaugusto, line 37: hey, do we need this new attribute? could it be just a standalone variable, or using `self.config["advertised_protocol"]` directly in the f-string?

If we don't need the attribute anywhere I would remove it, rit makes refactoring a bit easier.

karapace/config.py
hackaugusto: > But for advertised_host, there is already a config called advertised_hostname, please see 

My bad, thanks for pointing it out :)

karapace/kafka_rest_apis/consumer_manager.py
YuWang24: Yeah, moving it to the f-string is better, I have changed that, thanks for pointing out this.

tvainika (changes requested):
Overall looks good now, but lint failure due formatting. We ensure code formatting with automated CI pipeline to preserve common style. If you know how to use precommit tool, it should help you to automate things locally, but in any  case please fix the formatting and we will merge after that.

YuWang24:
> Overall looks good now, but lint failure due formatting. We ensure code formatting with automated CI pipeline to preserve common style. If you know how to use precommit tool, it should help you to automate things locally, but in any case please fix the formatting and we will merge after that.

Hey @tvainika, I have checked the error, it is because the long is too long:
`karapace/kafka_rest_apis/consumer_manager.py:42:0: C0301: Line too long (139/125) (line-too-long)`
If it does matter, I can make the line shorter by breaking it into two lines.

YuWang24:
@tvainika I have made the changes to fix the failure in lint. I guess it should be able to pass now, thanks for pointing this out. By the way, will there be a new release after merging this PR? Or it will be bundled together with some other PRs? If that's the case, when will be the estimated time to see my changes in a release?

tvainika:
I recommend trying precommit tool as with that you get automated local lint run whenever you're committing with git. It avoids slow cycle of CI runs waiting for feedback with lint failures. The requirements-dev.txt includes pre-commit so if you have virtualenv setup, then pre-commit is available, and you can run it manually with `pre-commit run --all-files` to get same lint checks, and also additionally you may install its git hooks with `pre-commit install`, and have the lint run automatically for you whenever you commit.

YuWang24:
@tvainika Hey, now it is complaining about the format of the code. I have run `black consumer_manager.py`  in my local, it seems there is no reformatting related to my changes anymore. But it's still failing on the black check. Could you spot where's could be wrong with my current changes? Thanks in advance.

tvainika:
We have `black` as automated code formatter, which keeps the style. Black has soft-max for line length, it tries to make lines at max 88 chars by length, but allows exceptions based on other rules. Our pylint has hard-limit of 125 for line length, and to resolve that one sometimes needs to use different style of line formatting to pass that. We have tried to make this easy with precommit as it will give you feedback immediately, but there seems to be somewhat corner case with these black & pylint line lengths.

YuWang24:
> We have `black` as automated code formatter, which keeps the style. Black has soft-max for line length, it tries to make lines at max 88 chars by length, but allows exceptions based on other rules. Our pylint has hard-limit of 125 for line length, and to resolve that one sometimes needs to use different style of line formatting to pass that. We have tried to make this easy with precommit as it will give you feedback immediately, but there seems to be somewhat corner case with these black & pylint line lengths.

Hey @tvainika, thanks for replying. Apparently, I can't solve this issue by running black locally, or there is no very useful information in pylint. Looking at my changes, what could be done to resolve this issue?

YuWang24:
How about reverting to commit (https://github.com/aiven/karapace/pull/400/commits/b01eebb66f985c1ca398e5e71107b1b6378f83c0)? This passed all the tests.

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -34,7 +34,16 @@ def new_name() -> str:
 class ConsumerManager:
     def __init__(self, config: dict) -> None:
         self.config = config
-        self.hostname = f"http://{self.config['advertised_hostname']}:{self.config['port']}"
+        if self.config["advertised_port"] is None:
+            self.hostname = (
+                f"{self.config['advertised_protocol']}://"
+                f"{self.config['advertised_hostname']}:{self.config['port']}"
+            )
```
hackaugusto, line 41: To fix the linter complain:
```suggestion
            self.hostname = (
                f"{self.config['advertised_protocol']}://{self.config['advertised_hostname']}:{self.config['port']}"
            )
```

---

As a note, I recommend getting `pre-commit` installed on your git repo:

https://pre-commit.com/#usage

That way the tool will run locally and you will get faster feedback cycles

YuWang24: Thanks a lot! I have installed and tried the pre-commit, it gave me exactly the same suggestion you sent. After changing it, it passed on my local. Appreciate that, will commit the change very soon.

tvainika (approved):

karapace/kafka_rest_apis/consumer_manager.py
hackaugusto: Thank you for the contribution :)

tvainika:
> By the way, will there be a new release after merging this PR? Or it will be bundled together with some other PRs? If that's the case, when will be the estimated time to see my changes in a release?

We're trying to speed up the time it takes to create new releases, but currently we're not planning to make a new release just for this. But let's say that within some weeks, or at least within a month from now there will be a new release.

Issue #381, opened 2022-04-01
Website update


lornajane:
Karapace has a web presence at https://karapace.io which is good but minimal. The plan (comments welcome) is to do the following:
 - move the website into this repository so that it's easier to keep both in sync
 - consider if the tool/build of the site needs updating, and if so, do that
 - move some of the usage examples and installation instructions from the README into website documentation, under "Installation" and "Examples" pages or sections
 - revisit all the wording on the existing site, and expand the "Home" and "About" pages

See also: #181 , and include API documentation on the site when we have it.

(should this be an epic, it's pretty extensive?)

PR #402, opened 2022-04-27
Replace Karapace's website with a landing page and some docs

lornajane:
Adds a simple sphinx setup in a `website/` folder that builds a static HTML website. Also adds styles, templates, and content.

Fixes: #381 

# Why this way
We already have Sphinx sites inside the business, and Python dependencies in this project.


lornajane:
With help from my friend (thanks Ioan!)
[Karapace.zip](https://github.com/aiven/karapace/files/9192341/Karapace.zip)
, I got some static assets to use as a template

lornajane:
Looking okay on desktop (I think!) but the navigation menu does not work for small screens, so it will need attention

website/source/_templates/page.html
```diff
@@ -0,0 +1,96 @@
...
+						<small class="trademark">Apache Kafka is either a registered trademark or a trademark of the Apache Software Foundation in the United States and/or other countries.</small>
+					</div>
+				</div>
+                {% endblock footer %}
+                </div>
+            </footer>
+        </div>
+    </div>
+</div>
+{%- endblock %}
```
blomqma, line 96: nit: newline

lornajane:
replaced by #489

PR #405, opened 2022-05-17
Remove ujson dependency

jjaakola-aiven:

Remove `ujson` dependency. After refactoring of deepcopy optimizations the ujson is not needed.

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -19,11 +19,10 @@ from tests.utils import (
 )
 from typing import List, Tuple
 
-import json as jsonlib
+import json
 import os
 import pytest
 import requests
-import ujson
 
 baseurl = "http://localhost:8081"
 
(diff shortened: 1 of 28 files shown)


tests/integration/test_schema.py
```diff
@@ -2414,15 +2409,15 @@ async def test_invalid_schema_should_provide_good_error_messages(registry_async_
...
     test_subject = subject_name_factory()
 
-    schema_str = ujson.dumps({"type": "string"})
+    schema_str = json.dumps({"type": "string"})
     res = await registry_async_client.post(
         f"subjects/{test_subject}/versions",
         json={"schema": schema_str[:-1]},
     )
-    assert res.json()["message"] == "Invalid AVRO schema. Error: Expecting ',' delimiter: line 1 column 17 (char 16)"
+    assert res.json()["message"] == "Invalid AVRO schema. Error: Expecting ',' delimiter: line 1 column 18 (char 17)"
```
juha-aiven, line 2417: Interesting change.... But I guess this doesn't matter?

juha-aiven (commented):
It looks that there's just the `ujson` to `json` change. Some places `ValueError` to `json.JSONDecodeError`. Looking good to me.

I'm not approving (yet), perhaps somebody else would like to have a look too.

tests/integration/test_schema.py
jjaakola-aiven: https://github.com/aiven/karapace/pull/330/files#diff-7ce48b7ba6aaf19c9eb9f548cd2fe0c0dad95c5a7d0957147477ed5ada247d1bL2317
The deepcopy optimization which used `ujson` introduced the change in the message. So this is reverting it back.

tvainika (approved):
LGTM and makes life simpler :shipit:

PR #406, opened 2022-05-20
Schema registry auth

tvainika:

Add authentication and authorization for Karapace Schema Registry API

# Why this way

Allow optional auth configuration file to define users and access control rules for Karapace Schema Registry API. Default mode allows any access in backward compatible manner.


karapace/schema_registry_apis.py
```diff
@@ -52,9 +60,110 @@ class SchemaErrorMessages(Enum):
     )
 
 
+@unique
```
jjaakola-aiven, line 63: Could all auth related models and functions live in `karapace/auth.py`?

karapace.auth.json
```diff
@@ -0,0 +1,33 @@
+{
```
jjaakola-aiven, line 1: Maybe this file could be under `tests` and serve as a template from there for final auth database files. Reduces the risk of running Karapace with default auth database.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+from karapace.client import Client
+from karapace.schema_models import SchemaType, ValidatedTypedSchema
+from tests.utils import new_random_name, schema_avro_json
+from urllib.parse import quote
+
+import aiohttp
+
+
+async def test_remote_client_auth(registry_async_client_auth: Client) -> None:
```
jjaakola-aiven, line 15: This tests the cases:
* no auth
* aladdin with WRITE to `cave-`, implicit to have READ also

I'd test also cases:
* aladdin tries to WRITE to `nick-cave-`
* aladdin to READ from `nick-cave-`
* plainuser to READ from `general` and  `cave-`
* plainuser to WRITE to `general` and `cave-`
* admin READ and WRITE

karapace/schema_registry_apis.py
```diff
@@ -16,11 +18,17 @@
 from typing import Any, Dict, NoReturn, Optional, Tuple
 
 import aiohttp
+import aiohttp.web
 import async_timeout
 import asyncio
+import crypt
```
jjaakola-aiven, line 24: Crypt will be deprecated in Python 3.11. Recommendation is to use `hashlib` instead.

jjaakola-aiven (changes requested):

karapace/schema_registry_apis.py
```diff
@@ -52,9 +60,110 @@ class SchemaErrorMessages(Enum):
...
+            )
+        except Exception as ex:
+            raise InvalidConfiguration("Auth configuration is not a valid JSON") from ex
+
+    def check_authorization(user: User, operation: Operation, resource: str) -> bool:
+        def check_operation(operation: Operation, acl: ACL) -> bool:
+            if operation == Operation.Read:
+                return True
+            if acl.operation == Operation.Write:
+                return True
```
juha-aiven, line 106: Does this mean that write access implies read access?

juha-aiven (changes requested):
First round of review.

karapace/schema_registry_apis.py
tvainika: My idea was that write implies read access as it makes no sense to have only write without read...

karapace.auth.json
```diff
@@ -0,0 +1,33 @@
...
+            "password": "$6$saltsalt$qFmFH.bQmmtXzyBY0s9v7Oicd2z4XSIecDzlB5KiA2/jctKu9YterLp8wwnSq.qc.eoxqOmSuNp2xS0ktL3nh/"
+        },
+        {
+            "username": "aladdin",
+            "password": "$6$salt$pRdN0h.pnEuLBJ2cux3aTtOa4OasTBXa3VUUeg5PI1dStiGZKA50HUJekUdVl0C6ZFt2TWqPT7l3P6QLcZHuY/"
+        }
+    ],
+    "acls": [
+        {
+            "principal": "plainuser",
```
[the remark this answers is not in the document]
tvainika: Done

karapace/schema_registry_apis.py
```diff
@@ -64,10 +74,10 @@ def __init__(self, config: Config) -> None:
         self.schema_lock = asyncio.Lock()
         self._master_lock = asyncio.Lock()
 
-        self._forward_client = None
+        self._forward_client: Optional[aiohttp.ClientSession] = None
```
jjaakola-aiven, line 77: When forwarding the authentication header needs to be passed.

karapace/schema_registry_apis.py
```diff
@@ -348,6 +421,26 @@ async def schemas_get(self, content_type, *, schema_id):
             )
         with self.ksr.id_lock:
             schema = self.ksr.schemas.get(schema_id_int)
+
+        def _has_subject_with_id() -> bool:
```
juha-aiven, line 425: Maybe do this before getting `schema`? If this returns `False`, there's no need to get the `schema`, it's set `None` below in that case. Also, maybe use something else than `schema` in the `for schema in val["schemas"].values():` below. Now the same variable name is used for different things in this method, which is slightly confusing.

karapace/schema_registry_apis.py
```diff
@@ -294,8 +360,15 @@ def send_delete_subject_message(self, subject, version):
...
 
-    async def compatibility_check(self, content_type, *, subject, version, request):
+    async def compatibility_check(
+        self, content_type: str, *, subject: str, version: str, request: HTTPRequest, user: Optional[User] = None
+    ) -> None:
         """Check for schema compatibility"""
+
+        if self._auth:
+            if not self._auth.check_authorization(user, Operation.Write, f"Subject:{subject}"):
+                self.r(body={"message": "Forbidden"}, content_type=JSON_CONTENT_TYPE, status=HTTPStatus.FORBIDDEN)
```
juha-aiven, line 370: This stuff should really be a method. It's repeated in so many places. The extracted method would take in the `Operation` and the resource. I'd also create do this  `f"Subject:{subject}"` in a single place in the code. So the method would take in the resource type (subject, config) and the resource "specifier", `None` in case of config and the subject name in the case of subject.

karapace/schema_registry_apis.py
```diff
@@ -52,9 +60,110 @@ class SchemaErrorMessages(Enum):
     )
 
 
+@unique
+class Operation(Enum):
+    Read = "Read"
+    Write = "Write"
```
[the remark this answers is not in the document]
tvainika: Conclusion was `Write` implies `Read`

tests/integration/test_schema_registry_auth.py
tvainika: `test_sr_list_subjects` now tests with three users to cover this

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+
+            last_loaded = os.path.getmtime(self._auth_filename)
+
+            while True:
+                try:
+                    await asyncio.sleep(5)
+                    last_modified = os.path.getmtime(self._auth_filename)
+                    if last_loaded < last_modified:
+                        self._load_authfile()
+                        last_loaded = last_modified
```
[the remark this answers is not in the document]
tvainika: It is true that this causes `FATAL` log message every 5 seconds if the file is not valid. However logging `FATAL` once might go unnoticed too easily, and is this causes old users & acl rules to be in use after trying to change those, it could be justified to keep logging this continuously... but you raised a good point to think still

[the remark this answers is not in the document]
tvainika: It is true that this causes `FATAL` log message every 5 seconds if the file is not valid. However logging `FATAL` once might go unnoticed too easily, and is this causes old users & acl rules to be in use after trying to change those, it could be justified to keep logging this continuously... but you raised a good point to think still

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+log = logging.getLogger(__name__)
+
+
+@unique
+class Operation(Enum):
+    Read = "Read"
+    Write = "Write"
+
+
+def hash_password(algorithm: str, salt: str, password: str) -> str:
```
juha-aiven, line 31: Make `algorithm` an `enum`. No need then to ever throw `NotImplementedError`.

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+        )
+    raise NotImplementedError(f"Hash algorithm '{algorithm}' is not implemented")
+
+
+@dataclass
+class User:
+    username: str
+    algorithm: str
+    salt: str
+    password: str = field(repr=False)  # hashed
```
juha-aiven, line 49: What about passing plaintext password around in variable called `password` and hashed one in `password_plaintext`. It'd always be obvious which flavor each variable / field is and no need for comments like this `# hashed`.

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def compare_password(self, plaintext: str) -> bool:
+        return compare_digest(self.password, hash_password(self.algorithm, self.salt, plaintext))
+
+
+@dataclass(frozen=True)
+class ACL:
```
juha-aiven, line 56: Nit: this in an access control list entry, not an ACL. Also, later in the code class is used in an array called `acls`. That is not not acls, as in access control lists, but a list of acl entries. Thus this would be more correct:

```python
class ACLEntry:
    ....

acl = [ACLEntry(x, y, z), ACLEntry(a, b, c)
```

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+    username: str
+    operation: Operation
+    resource: re.Pattern
+
+
+class HTTPAuthorizer:
+    def __init__(self, filename: str):
```
juha-aiven, line 63: Return type missing. Isn't `mypy` enabled?

juha-aiven (changes requested):
1st check, for `karapace/auth.py` only. I'll check the other files later.

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+    args = parser.parse_args()
+    salt: str = args.salt or secrets.token_urlsafe(nbytes=16)
+    result = {}
+    if args.user:
+        result["username"] = args.user
+    result["algorithm"] = args.algorithm
+    result["salt"] = salt
+    result["password"] = hash_password(args.algorithm, salt, args.password)
+    print(json.dumps(result, indent=4))
+    return 0
```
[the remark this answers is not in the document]
tvainika: See setup.py `console_scripts` and README for usage documentation

karapace/auth.py
tvainika: I don't want to support storing plaintext passwords on the file, those are only passed around internally for password checking. But you think this should be `hashed_password` or something?

karapace/schema_registry_apis.py
```diff
@@ -903,22 +1061,29 @@ def write_new_schema_local(
         )
         self.r({"id": schema_id}, content_type)
 
-    async def _forward_request_remote(self, *, body, url, content_type, method="POST"):
+    async def _forward_request_remote(
```
juha-aiven, line 1064: To keep the API tighter, this could have instead of full `request` just `authorization: Optional[str]`.

juha-aiven (changes requested):
Some notes:

* Test the case when the authfile is missing. 
* Test the case when the authfile is there but it's not valid JSON

Left comments, mostly in `karapace/schema_registry_apis.py`. I mostly concentrated on that in the review.

karapace/auth.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class HTTPAuthorizer:
+    def __init__(self, filename: str) -> None:
+        self._auth_filename: str = filename
+        self._refresh_auth_task: Optional[asyncio.Task] = None
+        # Once first, can raise if file not valid
+        self._load_authfile()
+
+    async def start_refresh_task(self, app: aiohttp.web.Application) -> None:  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
tvainika: Yes: `# pylint: disable=unused-argument`

karapace/auth.py
juha-aiven: Yea, my point is just that if there's a variable or field called `password`, it kind of suggests a _plaintext_ password. At least there's nothing hinting it's a hash. Thus `password_hash` or something like that would be more obvious. And then the parameter `plaintext` in `compare_password` could be called `password`. Again, `password` hints to the direction it's the password as such, not a hash.

karapace/schema_registry_apis.py
```diff
@@ -92,12 +103,18 @@ def _create_producer(self) -> KafkaProducer:
...
+            if not self._auth.check_authorization(user, operation, resource):
+                self.r(body={"message": "Forbidden"}, content_type=JSON_CONTENT_TYPE, status=HTTPStatus.FORBIDDEN)
+
     def _add_schema_registry_routes(self) -> None:
         self.route(
             "/compatibility/subjects/<subject:path>/versions/<version:path>",
             callback=self.compatibility_check,
             method="POST",
             schema_request=True,
+            auth=self._auth,
```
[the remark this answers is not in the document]
tvainika: I think this won't work with REST API if Karapace runs both schema registry and REST proxy due single instance of karapace with multi-inheritance.

karapace/schema_registry_apis.py
```diff
@@ -92,12 +103,18 @@ def _create_producer(self) -> KafkaProducer:
                 self.log.exception("Unable to create producer, retrying")
                 time.sleep(1)
 
+    def _check_authorization(self, user: User, operation: Operation, resource: str) -> None:
```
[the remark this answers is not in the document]
tvainika: IMHO it works currently just fine. `/schemas/types` is not authenticated (returns static response based on what is implemented on Karapace), and it does not have `auth=` in route, but also it does not call `self._check_authorization` at all, and this keeps those in sync just fine.

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+                self.userdb = users
+                self.acls = acls
+                log.info(
+                    "Loaded schema registry ACL rules: %s",
+                    [(acl.username, acl.operation.value, acl.resource.pattern) for acl in acls],
+                )
+            except Exception as ex:
+                raise InvalidConfiguration("Auth configuration is not a valid JSON") from ex
+
+    def check_authorization(self, user: Optional[User], operation: Operation, resource: str) -> bool:
```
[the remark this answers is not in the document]
tvainika: I think it safer to keep this `Optional[User]` and related condition there, even if it is not currently called without user.

karapace/schema_registry_apis.py
tvainika: I think passing around `request` simpler instead of querying `Authorization` header in every call invocation.

tvainika:
>     * Test the case when the authfile is missing.
>     * Test the case when the authfile is there but it's not valid JSON

I'm not sure how relevant these kind of tests are as those would basically only test that karapace always returns same error message. It is currently not very stylistic error message as I think Karapace as server-software has users who prefer more "raw" style of error message. However thinking this I found case if authfile is removed, it was not handling os error for last modified check.

karapace/auth.py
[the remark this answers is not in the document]
jjaakola-aiven: This does not get reported anywhere as the `Exception` is catched and not re-raised.
See: https://github.com/aiven/karapace/blob/36575e34230ce7a4c6d44e452d07cfe88a9c1cf9/karapace/rapu.py#L384

Only way to notice this from user perspective is that authentication fails for new credentials. Then the logs need to be reviewed.

I'd say the reattempt is ok and the logging is ok but the error handling scenario is not complete.

karapace/auth.py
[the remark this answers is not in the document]
jjaakola-aiven: I agree with @tvainika. I streamlined a bit and `if user is None` returns immediately with `False`.

karapace/auth.py
jjaakola-aiven: Changed to `password_hash` and `plaintext_password`. Metavar for command line is `password` but stored in `plaintext_password` of args map.

README.rst
```diff
@@ -424,6 +428,82 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
      - ``lowest``
      - Decides on what basis the Karapace cluster master is chosen (only relevant in a multi node setup)
 
+Karapace Schema Registry authorization file is an optional JSON configuration, which contains a list of authorized users in ``users`` list and a set of access control rules in ``permissions`` list.  Karapace uses HTTP Basic Authentication for user authentication.
```
juha-aiven, line 431: `list of access control rules`, not a set.

Better like this

which contains a list of authorized users in ``users`` and a list of access control rules in ``permissions``.

juha-aiven: > Karapace uses HTTP Basic Authentication for user authentication.

I don't think it's 100% clear from the documentation that Karapace uses this _only_ if the auth file is there. Also, the files is used only in Schema Registry side.

I think it'd be cleaner to dedicate a chapter to Karapace Schema Registry authentication and authorization to have these things in the correct context.

README.rst
```diff
@@ -401,6 +401,10 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
      - Kafka Registry CA certificate, used by Kafka Rest for Avro related requests.
        If this is set, Kafka Rest will use HTTPS to connect to the registry.
        If running both in the same process, it should be left to its default value
+   * - ``registry_authfile``
+     - ``/path/to/authfile.json``
+     - Filename to specify users and access control rules for Karapace Schema Registry.
+       If this is set, schema registry requires authorized user to connect to the registry.
```
juha-aiven, line 407:
```suggestion

       If this is set, Schema Segistry requires authentication for most of the endpoints and applies per endpoint authorization rules.

File: README.rst [hunk 24]
```diff
@@ -424,6 +428,82 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
...
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``username``
+     - A string
+   * - ``algorithm``
+     - One of supported hashing algorithms, ``scrypt``, ``sha1``, ``sha256``, or ``sha512``
+   * - ``salt``
+     - Salt used for hashing the password
```
juha-aiven, line 445: How exactly is it used?

README.rst
```diff
@@ -424,6 +428,82 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
...
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``username``
+     - A string to match against authenticated user
+   * - ``operation``
+     - Exact value of ``Read`` or ``Write``. Write implies also read permissions. Write includes all mutable operations, e.g. deleting schema versions
+   * - ``resource``
+     - A regular expression used to match against accessed resource of form ``Subject:<subject_name>`` or ``Config:<optional_subject_name>``
```
juha-aiven, line 471: What does the "form" mean here - resource can be any valid Python regexp, as can be seen in the example below.

README.rst
```diff
@@ -424,6 +428,82 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
...
+
+Example of complete authorization file
```
juha-aiven, line 473: As said earlier, configuring the auth file should have a heading and this would then be a sub-header.

README.rst
juha-aiven: It looks it's missing how exactly authorizations works. I don't think it's clear at all from the documentation.

We have this: 

> A regular expression used to match against accessed resource of form Subject:<subject_name> or Config:<optional_subject_name>

There are at least these things missing:

* How global ``Config:`` resource works
* Mentioning about content filtering, not just limiting access or allowing

Maybe this logic should be opened a little? Maybe not explaining each endpoint separately, but the general idea like if the user has been granted read access to a subject X, he has access to all "read" endpoints that have that subject in the URL and the endpoints that returns schemas will filter in the subject X. Etc.

juha-aiven (dismissed):
Reviewed README.rst

karapace/auth.py
```diff
@@ -0,0 +1,185 @@
...
+    parser.add_argument("password", help="Password to hash", type=str)
+    parser.add_argument("salt", help="Salt for hashing, random generated if not given", nargs="?", type=str)
+    args = parser.parse_args()
+    salt: str = args.salt or secrets.token_urlsafe(nbytes=16)
+    result = {}
+    if args.user:
+        result["username"] = args.user
+    result["algorithm"] = args.algorithm
+    result["salt"] = salt
+    result["password"] = hash_password(args.algorithm, salt, args.password)
```
juha-aiven: `password_hash` instead?

README.rst
```diff
@@ -424,6 +428,82 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
...
+
+::
+
+    {
+        "users": [
+            {
+                "username": "admin",
+                "algorithm": "scrypt",
+                "salt": "5nq6IYKAwodZEsEs4UElYw",
+                "password": "1Z+NCmV6nYFfrTPugqftgP8uWGeIAFOWMpLnjVJ3kEoQdsbEgM+33jdzBF+DR6KcMDl/7RJpKz4Olh3sC5VvuA=="
```
juha-aiven: `password` -> `password_hash`, the same in the other entry.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,138 @@
...
+
+admin = aiohttp.BasicAuth("admin", "admin")
+aladdin = aiohttp.BasicAuth("aladdin", "opensesame")
+reader = aiohttp.BasicAuth("reader", "secret")
+
+
+async def test_sr_auth(registry_async_client_auth: Client) -> None:
+    subject = new_random_name("cave-")
+
+    res = await registry_async_client_auth.post(f"subjects/{quote(subject)}/versions", json={"schema": schema_avro_json})
```
juha-aiven, line 25: `quote(subject)` is not needed, `subject` works fine. The format is something like `cave-5cfff223` that doesn't need `quote` call:

```
quote("cave/5cfff223")
'cave/5cfff223'
```

Calling `quote` makes things look like it's needed.

juha-aiven (changes requested):
General note of test coverage. If I looked correctly, `tests/integration/test_schema_registry_auth.py` tests the following Schema Registry endpoints:

* test_sr_auth: /subject/<subject>/versions, /subject/<subject>/versions/latest 
* test_sr_list_subjects: /subject/<subject>/versions,  /subject
* test_sr_ids: /subject/<subject>/versions, /schemas/ids/ filters based on user

Thus the endpoints tested are: /subject/<subject>/versions POST, /subject/<subject>/versions/latest, /subjects and /schemas/ids/. Maybe I missed something.

But in there are ~15 endpoints that enable authorization. There's no testing for most of the endpoints? I think we should have at least basic authorization testing for all, e.g. a simple case where authorization forbids the endpoint or filters out the response and a success case.

tests/integration/test_schema_registry_auth.py
tvainika: Using `quote` is correct here, although `new_random_name` does not currently generate special characters. However if there would be any such special characters in subject name, then `quote` does the right thing.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,215 @@
...
+    wait_time = 0.5
+    for compat in ["FULL", "BACKWARD", "FORWARD", "NONE"]:
+        resp = requests.put(f"{replica_url}/config", json={"compatibility": compat}, auth=auth)
+        assert resp.ok
+        while True:
+            if counter >= max_tries:
+                raise Exception("Compat update not propagated")
+            resp = requests.get(f"{primary_url}/config", auth=auth)
+            if not resp.ok:
+                print(f"Invalid http status code: {resp.status_code}")
```
ivanyu, line 191: Probably it's worth bringing a logger in here.

README.rst
```diff
@@ -424,6 +434,105 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
...
+
+.. list-table::
+   :header-rows: 1
+
+   * - Resource
+     - Description
+   * - ``Subject:<subject_name>``
+     - Controls authorization to subject. The ``<subject_name>`` is a regular expression to match against the accessed subject.
+   * - ``Config:``
+     - Controls authorization to global schema registry configuration.
```
juha-aiven, line 481: It'd be more logical to change the order of these two tables. First, explain what the element of each entry are and then explain the format of `resource`.

tests/integration/conftest.py
```diff
@@ -278,14 +279,85 @@ async def get_client() -> ClientSession:
...
+    # so would cause this node to join the existing group and participate in
+    # the election process. Without proper configuration for the listeners that
+    # won't work and will cause test failures.
+    rest_url = request.config.getoption("rest_url")
+    if rest_url:
+        yield None
+        return
+
+    registry = urlparse(registry_async_client_auth.server_uri)
+    config = config = set_config_defaults(
```
juha-aiven, line 300:
```
config = set_config_defaults(
```

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,215 @@
...
+                continue
+            if data["compatibilityLevel"] != compat:
+                print(f"Bad compatibility: {data}")
+                counter += 1
+                await asyncio.sleep(wait_time)
+                continue
+            break
+
+
+async def test_rest_api_with_sr_auth(rest_async_client_registry_auth: Client, admin_client: KafkaRestAdminClient) -> None:
```
ivanyu, line 207: Just checking: this tests that we can produce via Kafka REST without credentials when SR auth is enabled, right?

ivanyu (changes requested):
Not a full review. A couple of comments/questions here, mostly minor, but I think the question about the status code when not authorized is important.

karapace/config.py
```diff
@@ -39,9 +39,11 @@
     "port": 8081,
     "server_tls_certfile": None,
     "server_tls_keyfile": None,
+    "registry_url": None,
```
juha-aiven, line 42: README doesn't mention `registry_url`. What's the difference that and `registry_host`?

juha-aiven (dismissed):
Left a few comment.

Additionally, a comment related to default config and README.

* Docker image should be up-to-date with main. Docker image should support setting authorization related conf parameters in `start.sh`.
* Quickstart in README should be up-to-date with at least Docker image, preferably with main too. I'm not sure if now Karapace Schema Registry is started from main, will it use `karapace.auth.json`? Or what's the usage of that file. If it's used, the `curl` commands won't work since authorization is enabled
 
I think the examples in Quickstart should rather use authentication than not to illustrate how it works.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,215 @@
...
+    res = await registry_async_client_auth.get(f"subjects/{quote(carpetsubject)}/versions", auth=admin)
+    assert res.status_code == 200
+    assert [sc_id] == res.json()
+
+    res = await registry_async_client_auth.get("subjects", auth=aladdin)
+    assert res.status_code == 200
+    assert [cavesubject] == res.json()
+
+    res = await registry_async_client_auth.get(f"subjects/{quote(carpetsubject)}/versions", auth=aladdin)
+    assert res.status_code == 403
```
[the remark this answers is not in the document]
tvainika: The spec (in my understanding) only refers to `GET /subjects` which lists all available subjects, not all `/subjects/something` prefixed endpoints. This exact endpoint does not leak any information, because access is checked first based on only requested path without reading the database.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,215 @@
...
+admin = aiohttp.BasicAuth("admin", "admin")
+aladdin = aiohttp.BasicAuth("aladdin", "opensesame")
+reader = aiohttp.BasicAuth("reader", "secret")
+
+
+async def test_sr_auth(registry_async_client_auth: Client) -> None:
+    subject = new_random_name("cave-")
+
+    res = await registry_async_client_auth.post(f"subjects/{quote(subject)}/versions", json={"schema": schema_avro_json})
+    assert res.status_code == 401
```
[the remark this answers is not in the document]
tvainika: Most of the existing tests use numbers for HTTP error codes instead of `HTTPStatus`, so maybe this needs general guideline which way we want it to be?

karapace/config.py
tvainika: My bad, did originally `registry_url` containing user & pw, but changed it after discussion to current form. This was leftover from first iteration.

tests/integration/test_schema_registry_auth.py
tvainika: Added comment why this test exists

tests/integration/test_schema_registry_auth.py
tvainika: Eliminated these prints as useless

tvainika:
> Additionally, a comment related to default config and README.
> 
>     * Docker image should be up-to-date with main. Docker image should support setting authorization related conf parameters in `start.sh`.

`start.sh` only has environment variable tuning for backward compatibility. New configuration settings do not need such bw-compatibility, but can be set via environment directly. `start.sh` does not need any modifications.

>     * Quickstart in README should be up-to-date with at least Docker image, preferably with main too. I'm not sure if now Karapace Schema Registry is started from main, will it use `karapace.auth.json`? Or what's the usage of that file. If it's used, the `curl` commands won't work since authorization is enabled

To setup authorization with Docker image, one needs to simply set the config via environment variable as documented in README, and make authorization config file available into container (e.g. mount), but Karapace README should not document Docker usage (Docker itself has a good documentation), but only contain simply quickstart guide for simple setup. Any complex setup anyway has special needs, and understanding, and trying to cover such here is IMHO simply impossible, and causes simple things became unnecessary complex instead.

> I think the examples in Quickstart should rather use authentication than not to illustrate how it works.

I disagree with this. It is a bad security practise to set up dummy passwords, so rather without a password than a bad password.

tests/integration/test_schema_registry_auth.py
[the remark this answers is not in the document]
ivanyu: Ok, if so let's ignore this for now. (I vote for named constants over literals.:)

tests/integration/test_schema_registry_auth.py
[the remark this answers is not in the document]
ivanyu: > This exact endpoint does not leak any information, because access is checked first based on only requested path without reading the database.

Ah, I see. So it means if a subject doesn't exist (and because of this we don't have an ACL record for it), it will also be 403, right?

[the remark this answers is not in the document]
tvainika: Yes, this is now 403 if no access regardless if a subject exists or not. Subject data is only checked if the user has access to it.

README.rst
ivanyu: Do we really need to provide the details info? I'd probably just post a link to https://en.wikipedia.org/wiki/Salt_(cryptography)

ivanyu (changes requested):
Not the final review, but leaving a couple of comments at this point

karapace/auth.py
```diff
@@ -0,0 +1,210 @@
...
+        return user
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(prog="karapace_mkpasswd", description="Karapace password hasher")
+    parser.add_argument("-u", "--user", help="Username", type=str)
+    parser.add_argument(
+        "-a", "--algorithm", help="Hash algorithm", choices=["sha1", "sha256", "sha512", "scrypt"], default="sha512"
+    )
+    parser.add_argument(metavar="password", dest="plaintext_password", help="Password to hash", type=str)
```
ivanyu, line 195: Just a note:
Probably fine for now, but later we need to ask a password interactively with `getpass` if this parameter is omitted, for better security.

ivanyu (dismissed):
I finished the review. Generally LGTM, but considering there are comments from other reviewers.

karapace/auth.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+        try:
+            auth = aiohttp.BasicAuth.decode(auth_header)
+        except ValueError:
+            # pylint: disable=raise-missing-from
+            raise aiohttp.web.HTTPUnauthorized(
+                headers={"WWW-Authenticate": 'Basic realm="Karapace Schema Registry"'},
+                text='{"message": "Unauthorized"}',
+                content_type=JSON_CONTENT_TYPE,
+            )
```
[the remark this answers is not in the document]
tvainika: AFAIU the header can be also other than `Basic`, e.g. `Digest` and valid formed, but not supported. In such cases also execution flow would end up into this exception. I think HTTP's Unauthorized is fine here, and no need to separate cases more.

karapace/auth.py
```diff
@@ -0,0 +1,210 @@
...
+        except Exception as ex:
+            raise InvalidConfiguration("Auth configuration is not valid") from ex
+
+    def check_authorization(self, user: Optional[User], operation: Operation, resource: str) -> bool:
+        if user is None:
+            return False
+
+        def check_operation(operation: Operation, aclentry: ACLEntry) -> bool:
+            if operation == Operation.Read:
+                return True
```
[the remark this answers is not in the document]
tvainika: Added docstring and changed ifs a bit.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,213 @@
...
+
+    res = await registry_async_client_auth.get("schemas/ids/1/versions")
+    assert res.status_code == 401
+
+    res = await registry_async_client_auth.get("schemas/types")
+    assert res.status_code == 200
+
+    res = await registry_async_client_auth.get("schemas/types", auth=admin)
+    assert res.status_code == 200
```
juha-aiven, line 83: Why test with `admin`? Also, the test explanation is not exactly correct, it says endpoints require authentication, but `schemas/types` doesn't need that. Maybe split testing that to another function. That way this test function would be consistently testing only the endpoints requiring authentication.

tests/integration/test_schema_registry_auth.py
```diff
@@ -0,0 +1,213 @@
...
+    res = await registry_async_client_auth.get(f"subjects/{quote(subject)}/versions/latest", auth=aladdin)
+    assert res.status_code == 200
+    assert sc_id == res.json()["id"]
+    assert ValidatedTypedSchema.parse(SchemaType.AVRO, schema_avro_json) == ValidatedTypedSchema.parse(
+        SchemaType.AVRO, res.json()["schema"]
+    )
+
+
+# Test each endpoint without authorization to verify those require authentícation
+async def test_sr_auth_endpoints(registry_async_client_auth: Client) -> None:
```
juha-aiven, line 56:
```python
async def test_sr_auth_endpoints(registry_async_client_auth: Client) -> None:
    """Test each endpoint without authorization to verify those require authentícation."""
```

juha-aiven (changes requested):
A few very small things.

karapace/auth.py
```diff
@@ -0,0 +1,211 @@
...
+                    "Loaded schema registry users: %s",
+                    users,
+                )
+                self.permissions = permissions
+                log.info(
+                    "Loaded schema registry access control rules: %s",
+                    [(entry.username, entry.operation.value, entry.resource.pattern) for entry in permissions],
+                )
+        except Exception as ex:
+            raise InvalidConfiguration("Auth configuration is not valid") from ex
```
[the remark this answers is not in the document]
tvainika: This is used both during the initialization and refresh task, but I reworded the exception message.

juha-aiven (approved):
LGTM

PR #407, opened 2022-05-23
Fix test logging for Kafka download

tvainika:

Fix logging in integration tests, which causes failure if log level enabled


juha-aiven (approved):
LGTM

PR #409, opened 2022-05-27
Fix subject name escaping

tvainika:

Fix subject name escaping

# Why this way

Handle special characters in subject name by using proper JSON formatting.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -265,7 +265,7 @@ class KarapaceSchemaRegistry(KarapaceBase):
         version: int,
         deleted: bool,
     ):
-        key = '{{"subject":"{}","version":{},"magic":1,"keytype":"SCHEMA"}}'.format(subject, version)
+        key = json.dumps({"subject": subject, "version": version, "magic": 1, "keytype": "SCHEMA"}, separators=(",", ":"))
         if schema:
             valuedict = {
                 "subject": subject,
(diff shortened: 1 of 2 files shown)


hackaugusto (commented):
I'm not sure if this should be merged. Per the JSON spec:

> An object is an unordered set of name/value pairs

https://www.json.org/json-en.html

Which means we can't guarantee consistent ordering of the keys, and if the key doesn't match byte-by-byte then compression doesn't work.

There is another ticket related to this issue: https://github.com/aiven/karapace/issues/347

Maybe just do the `json.dumps` for the values you want to escape?

tvainika:
> Which means we can't guarantee consistent ordering of the keys

https://docs.python.org/3/library/json.html#json.dumps  has argument `sort_keys=False`, thus it is fixed ordering.

hackaugusto:
> https://docs.python.org/3/library/json.html#json.dumps has argument sort_keys=False, thus it is fixed ordering.

That agrees with what I said, no? The documentation states what the behavior is for when the flag is `True`:

> If sort_keys is true (default: False), then the output of dictionaries will be sorted by key; this is useful for regression tests to ensure that JSON serializations can be compared on a day-to-day basis.

The behavior for `False` is not really documented. I would assume they don't do anything special, just iterate over the dictionary. (which seems to [be the case](https://github.com/python/cpython/blob/5893b5db98b38b17750c0572c7209774a5034898/Modules/_json.c#L1549-L1555)).

So I guess you could rely on the insertion order of the keys:

> Dictionaries preserve insertion order, meaning that keys will be produced in the same order they were added sequentially over the dictionary. Replacing an existing key does not change the order, however removing a key and re-inserting it will add it to the end instead of keeping its old place.

But I can't the what is the insertion order for [dictionary displays](https://docs.python.org/3/reference/expressions.html#dictionary-displays). Your call if you want to rely on the above.

HelenMel (approved):

PR #413, opened 2022-06-08
Init asyncio locks when event loop is available

jjaakola-aiven:

Asyncio locks need to be attached to correct thread. If locks are created in constructor the loop is not available, instead init locks from `aiohttp.web.Application` startup callback.

# Why this way

When calling produce API from REST proxy the first requests fail as the lock acquire future is attached to main loop.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -16,6 +16,7 @@ from karapace.serialization import InvalidMessageSchema, InvalidPayload, SchemaR
 from karapace.utils import convert_to_int, KarapaceKafkaClient
 from typing import List, Optional, Tuple
 
+import aiohttp.web
 import asyncio
 import base64
 import json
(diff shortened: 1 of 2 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -159,6 +167,9 @@ def _add_kafka_rest_routes(self) -> None:
         self.route("/topics/<topic:path>", callback=self.topic_publish, method="POST", rest_request=True)
 
     async def _maybe_create_async_producer(self) -> AIOKafkaProducer:
+        if self._async_producer is not None:
+            return self._async_producer
+
```
juha-aiven, line 172: Is this just an optimization? After all if `_async_producer is None`, the `while` loop below is never executed and `self._async_producer` is returned.

OR

Is this to avoid using the lock. If so, why?

jjaakola-aiven: Avoid using the lock as this is feasible optimization route. The `_async_producer` is created one, never set back to `None`.

tvainika (approved):

PR #414, opened 2022-06-14
Improve deserialization error message

juha-aiven:
The error message tell whether the failure was when deserializing key or value. Log the requested format.

Add a dedicated error message for key and value deserialization error and log the requested type. This helps the end user to see what exactly failed.


hackaugusto (approved):
LGTM

PR #418, opened 2022-06-20
Performance: schema string hash to schema id

jjaakola-aiven:

Fast path check of schema existence on subject, based on hash of subject and schema string.

Why this way
==========

* The schema, subject pairs are hashed to key for getting schema id, this avoids version list iteration and comparison of input schema to each one in the list like in the original code.
* The schema existence on subject can be done without locking, this is the fast path.
 * If schema is not found a lock is acquired and existence is rechecked.
 * If schema is found the id is returned, concurrently another request has POSTed the schema.
 * If schema is not found after second round of checks it is a new schema.

The round-trip to wait the schema message to be sent to Kafka and receiving it in the schema reader is the longest path and during this time the schema write operation is locked. No optimization is done to it.

Performance test comparison
=======================

The following graphs have both versions run sequantially with the same test scenario on development laptop. The `_schemas` topic is deleted and recreated between the test runs.
Numbers in clustered environment are likely different, but this shows the performance difference of original and modified code.

Test creates 2000 schemas to two subjects, total is 4000. Detailed response time statistics are below.

The original code can achieve 10-13 RPS constantly when unique new schema is POSTed.
The modified code can achieve 20-25 RPS constantly when unique new schema is POSTed.

The original code can achieve 60 RPS constantly when existing schema is POSTed.
The modified code can achieve 2200 RPS quite consistently when existing schema is POSTed. The cause RPS drops in the graphs is unknown, low point of the drop is 1200 RPS.

modified-vs-original-schema-post

Stats for code before the change, no fast path, with list iteration and locking
----------------------------------------------------------------------------------------------------

Locust report:

    Response time percentiles (approximated)
    Type     Name                                                                                  50%    66%    75%    80%    90%    95%    98%    99%  99.9% 99.99%   100% # reqs
    --------|--------------------------------------------------------------------------------|--------|------|------|------|------|------|------|------|------|------|------|------
    POST     /subjects/test-subject-1/versions?EXISTING                                            170    190    200    220    250    290    330    350    410    560    710  19928
    POST     /subjects/test-subject-1/versions?NEW                                                 420    540    660    760    860   1000   1300   1700   2200   2200   2200   2000
    POST     /subjects/test-subject-2/versions?EXISTING                                            170    190    200    210    250    290    330    350    400    650    790  19207
    POST     /subjects/test-subject-2/versions?NEW                                                 430    550    660    770    850   1000   1300   1700   2100   2200   2200   2000
    --------|--------------------------------------------------------------------------------|--------|------|------|------|------|------|------|------|------|------|------|------
             Aggregated                                                                            180    200    210    230    300    400    750    850   1600   2200   2200  43135

Stats for code with schema hash to schema id, locking only if not found
-----------------------------------------------------------------------------------------------

Locust report:

    Response time percentiles (approximated)
    Type     Name                                                                                  50%    66%    75%    80%    90%    95%    98%    99%  99.9% 99.99%   100% # reqs
    --------|--------------------------------------------------------------------------------|--------|------|------|------|------|------|------|------|------|------|------|------
    POST     /subjects/test-subject-1/versions?EXISTING                                              4      4      4      5      6      7      8      9     13     24    550 803315
    POST     /subjects/test-subject-1/versions?NEW                                                 280    360    490    520    600    680    900   1100   1100   1100   1100   2000
    POST     /subjects/test-subject-2/versions?EXISTING                                              4      4      4      5      6      7      8      9     13     23    370 802742
    POST     /subjects/test-subject-2/versions?NEW                                                 280    360    490    520    600    680    900   1100   1100   1100   1100   2000
    --------|--------------------------------------------------------------------------------|--------|------|------|------|------|------|------|------|------|------|------|------
             Aggregated                                                                              4      4      4      5      6      7      8     10    330    720   1100 1610057

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -26,6 +26,7 @@ Version = int
 Schema = Dict[str, Any]
 # Container type for a subject, with configuration settings and all the schemas
 SubjectData = Dict[str, Any]
+SchemaId = int
 
 # The value `0` is a valid offset and it represents the first message produced
 # to a topic, therefore it can not be used.
@@ -173,6 +174,7 @@ class KafkaSchemaReader(Thread):
         # instead of the ids to handle corrupt data (where the ids are equal
         # but the schema themselves don't match)
         self._hash_to_schema: Dict[int, TypedSchema] = {}
+        self._hash_to_schema_id_on_subject: Dict[int, SchemaId] = {}
 
     def get_schema_id(self, new_schema: TypedSchema) -> int:
         with self.id_lock:
@@ -181,6 +183,15 @@ class KafkaSchemaReader(Thread):
                     return schema_id
             self.global_schema_id += 1
             return self.global_schema_id
+
+    def get_schema_id_if_exists(self, *, subject: Subject, schema: TypedSchema) -> Optional[SchemaId]:
+        return self._hash_to_schema_id_on_subject.get(hash((subject, schema.__str__())), None)
+
+    def _set_schema_id_on_subject(self, *, subject: Subject, schema: TypedSchema, schema_id: SchemaId) -> None:
+        self._hash_to_schema_id_on_subject.setdefault(hash((subject, schema.__str__())), schema_id)
+
+    def _delete_from_schema_id_on_subject(self, *, subject: Subject, schema: TypedSchema) -> None:
+        self._hash_to_schema_id_on_subject.pop(hash((subject, schema.__str__())), None)
 
     def close(self) -> None:
         LOG.info("Closing schema_reader")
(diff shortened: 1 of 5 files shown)


karapace/schema_reader.py
```diff
@@ -182,6 +184,12 @@ def get_schema_id(self, new_schema: TypedSchema) -> int:
             self.global_schema_id += 1
             return self.global_schema_id
 
+    def get_schema_id_if_exists(self, *, subject: Subject, schema: TypedSchema) -> Optional[int]:
+        return self._hash_to_schema_id_on_subject.get(hash((subject, schema.__str__())), None)
+
+    def _delete_from_schema_id_on_subject(self, *, subject: Subject, schema: TypedSchema) -> None:
+        self._hash_to_schema_id_on_subject.pop(hash((subject, schema.schema_str)), None)
```
tvainika, line 191: Get is using `TypedSchema`'s `__str__()`, but deleting using `schema.schema_str`. Why is that? Shouldn't this always use normalized string representation?

karapace/schema_reader.py
```diff
@@ -362,25 +371,17 @@ def _handle_msg_schema(self, key: dict, value: Optional[dict]) -> None:
...
-                schema_type=schema_type_parsed,
-                schema_str=schema_str,
-            )
-            schema = {
-                "schema": typed_schema,
-                "version": schema_version,
-                "id": schema_id,
-                "deleted": schema_deleted,
-            }
+            self._hash_to_schema_id_on_subject.setdefault(hash((schema_subject, schema_str)), SchemaId(schema_id))
```
tvainika, line 374: Inside this `with` statement, beginning has this line setting value into hashmap, but if it was actually `schema_deleted=True`, then this same(?) is removed some lines later?

Also, for get & delete there is function to do those, but not for setting.

tvainika (commented):
I went throught this, and had some questions

karapace/schema_reader.py
jjaakola-aiven: I'll add the function.

karapace/schema_reader.py
jjaakola-aiven: I decided not to add documentation as now all subject+schema->id mapping functions are close together.

tvainika (approved):
LGTM

PR #419, opened 2022-06-21
Remove schema registry logic from API controller to module

jjaakola-aiven:

The `karapace/schema_registry_apis.py` contained application logic
which is mostly moved to module `karapace/schema_registry.py`.
The API controller is now responsible of mapping the requests,
responses and authentication/authorization.

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -175,7 +175,7 @@ class MasterCoordinator(Thread):
         self.sc.master_eligibility = self.config["master_eligibility"]
         self.schema_coordinator_ready.set()
 
-    def get_master_info(self) -> Tuple[bool, Optional[str]]:
+    def get_master_info(self) -> Tuple[Optional[bool], Optional[str]]:
         """Return whether we're the master, and the actual master url that can be used if we're not"""
         self.schema_coordinator_ready.wait()
         return self.sc.are_we_master, self.sc.master_url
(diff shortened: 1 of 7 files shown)


karapace/master_coordinator.py
```diff
@@ -175,7 +175,7 @@ def init_schema_coordinator(self):
         self.sc.master_eligibility = self.config["master_eligibility"]
         self.schema_coordinator_ready.set()
 
-    def get_master_info(self) -> Tuple[bool, Optional[str]]:
+    def get_master_info(self) -> Tuple[Optional[bool], Optional[str]]:
```
mdedetrich, line 178: Is the reason why the `bool` is `Optional` because its representing an absence of a value? An `Optional[bool]` can be seen as misleading and it can be argued that just having a bool (if in context there is no real difference between absence of value and value being false) or class with subtypes representing the different states of the value?

karapace/schema_registry.py
```diff
@@ -0,0 +1,427 @@
...
+        if subject_data.get("compatibility"):
+            ret["compatibility"] = subject_data.get("compatibility")
+        return ret
+
+    async def write_new_schema_local(
+        self,
+        subject: Subject,
+        new_schema: ValidatedTypedSchema,
+    ) -> int:
+        """Since we're the master we get to write the new schema"""
```
mdedetrich, line 241: This feels more like a comment than a docstring

mdedetrich (approved):
Looks good to me, just need to fix the merge conflict

HelenMel (approved):

PR #423, opened 2022-07-01
Kafka authorization over REST proxy

tvainika:

Delegates REST caller principal to Kafka for authorization inside Kafka.

# Why this way

Delegation of caller makes it possible for Kafka to check authorization of REST API caller using already existing Kafka ACL entries & principals.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -364,6 +364,7 @@ class UserRestProxy:
         self.kafka_timeout = kafka_timeout
         self.serializer = serializer
         self._cluster_metadata = None
+        self._cluster_metadata_complete = False
         self._metadata_birth = None
         self.metadata_max_age = self.config["admin_metadata_max_age"]
         self.admin_client = None
@@ -523,16 +524,23 @@ class UserRestProxy:
             if self._metadata_birth is None or time.monotonic() - self._metadata_birth > self.metadata_max_age:
                 self._cluster_metadata = None
 
-            if self._cluster_metadata and topics is None:
-                return self._cluster_metadata
+            if self._cluster_metadata:
+                # Return from metadata only if all queried topics have cached metadata
+                if topics is None:
+                    if self._cluster_metadata_complete:
+                        return self._cluster_metadata
+                elif all(topic in self._cluster_metadata["topics"] for topic in topics):
+                    return {
+                        **self._cluster_metadata,
+                        "topics": {topic: self._cluster_metadata["topics"][topic] for topic in topics},
+                    }
 
             try:
                 metadata_birth = time.monotonic()
                 metadata = self.admin_client.cluster_metadata(topics)
-                # Only store cache if fetched full metadata
-                if topics is None:
-                    self._metadata_birth = metadata_birth
-                    self._cluster_metadata = metadata
+                self._metadata_birth = metadata_birth
+                self._cluster_metadata = metadata
+                self._cluster_metadata_complete = topics is None
             except NodeNotReadyError:
                 log.exception("Could not refresh cluster metadata")
                 KafkaRest.r(
@@ -750,13 +758,7 @@ class UserRestProxy:
                 content_type=content_type,
                 sub_code=RESTErrorCodes.PARTITION_NOT_FOUND.value,
             )
-        except (UnknownTopicOrPartitionError, TopicAuthorizationFailedError):
-            KafkaRest.not_found(
-                message=f"Partition {partition} not found",
-                content_type=content_type,
-                sub_code=RESTErrorCodes.PARTITION_NOT_FOUND.value,
-            )
-        except KeyError:
+        except (KeyError, UnknownTopicOrPartitionError, TopicAuthorizationFailedError):
             KafkaRest.not_found(
                 message=f"Topic {topic} not found",
                 content_type=content_type,
@@ -765,14 +767,15 @@ class UserRestProxy:
         return {}
 
     async def get_topic_info(self, topic: str, content_type: str) -> dict:
-        metadata = await self.cluster_metadata()
-        if topic not in metadata["topics"]:
+        try:
+            metadata = await self.cluster_metadata([topic])
+            return metadata["topics"][topic]
+        except (KeyError, UnknownTopicOrPartitionError, TopicAuthorizationFailedError):
             KafkaRest.not_found(
-                message=f"Topic {topic} not found in {list(metadata['topics'].keys())}",
+                message=f"Topic {topic} not found",
                 content_type=content_type,
                 sub_code=RESTErrorCodes.TOPIC_NOT_FOUND.value,
             )
-        return metadata["topics"][topic]
 
     @staticmethod
     def all_empty(data: dict, key: str) -> bool:
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -7,6 +7,7 @@ from http import HTTPStatus
 from kafka.errors import (
     BrokerResponseError,
     KafkaTimeoutError,
+    NoBrokersAvailable,
     NodeNotReadyError,
     TopicAuthorizationFailedError,
     UnknownTopicOrPartitionError,
@@ -221,31 +222,37 @@ class KafkaRest(KarapaceBase):
 
     def get_user_proxy(self, request: HTTPRequest) -> "UserRestProxy":
         key = ""
-        if self.config.get("rest_authorization", False):
-            auth_header = request.headers.get("Authorization")
+        try:
+            if self.config.get("rest_authorization", False):
+                auth_header = request.headers.get("Authorization")
 
-            if auth_header is None:
-                raise HTTPResponse(
-                    body='{"message": "Unauthorized"}',
-                    status=HTTPStatus.UNAUTHORIZED,
-                    content_type=JSON_CONTENT_TYPE,
-                    headers={"WWW-Authenticate": 'Basic realm="Karapace REST Proxy"'},
-                )
-            key = auth_header
-            if self.proxies.get(key) is None:
-                auth = aiohttp.BasicAuth.decode(auth_header)
-                config = self.config.copy()
-                config["bootstrap_uri"] = config["sasl_bootstrap_uri"]
-                config["security_protocol"] = (
-                    "SASL_SSL" if config["security_protocol"] in ("SSL", "SASL_SSL") else "SASL_PLAINTEXT"
-                )
-                config["sasl_mechanism"] = "PLAIN"
-                config["sasl_plain_username"] = auth.login
-                config["sasl_plain_password"] = auth.password
-                self.proxies[key] = UserRestProxy(config, self.kafka_timeout, self.serializer)
-        else:
-            if self.proxies.get(key) is None:
-                self.proxies[key] = UserRestProxy(self.config, self.kafka_timeout, self.serializer)
+                if auth_header is None:
+                    raise HTTPResponse(
+                        body='{"message": "Unauthorized"}',
+                        status=HTTPStatus.UNAUTHORIZED,
+                        content_type=JSON_CONTENT_TYPE,
+                        headers={"WWW-Authenticate": 'Basic realm="Karapace REST Proxy"'},
+                    )
+                key = auth_header
+                if self.proxies.get(key) is None:
+                    auth = aiohttp.BasicAuth.decode(auth_header)
+                    config = self.config.copy()
+                    config["bootstrap_uri"] = config["sasl_bootstrap_uri"]
+                    config["security_protocol"] = (
+                        "SASL_SSL" if config["security_protocol"] in ("SSL", "SASL_SSL") else "SASL_PLAINTEXT"
+                    )
+                    config["sasl_mechanism"] = "PLAIN"
+                    config["sasl_plain_username"] = auth.login
+                    config["sasl_plain_password"] = auth.password
+                    self.proxies[key] = UserRestProxy(config, self.kafka_timeout, self.serializer)
+            else:
+                if self.proxies.get(key) is None:
+                    self.proxies[key] = UserRestProxy(self.config, self.kafka_timeout, self.serializer)
+        except NoBrokersAvailable:
+            # This can be caused also due misconfigration, but kafka-python's
+            # KafkaAdminClient cannot currently distinguish those two cases
+            log.exception("Failed to connect to Kafka with the credentials")
+            self.r(body={"message": "Forbidden"}, content_type=JSON_CONTENT_TYPE, status=HTTPStatus.FORBIDDEN)
         return self.proxies[key]
 
     async def list_brokers(self, content_type: str, *, request: HTTPRequest):
@@ -379,7 +386,10 @@ class UserRestProxy:
             acks = int(self.config["producer_acks"])
 
         async with self._async_producer_lock:
-            while self._async_producer is None:
+            for retry in [True, True, False]:
+                if self._async_producer is not None:
+                    break
+
                 log.info("Creating async producer")
 
                 # Don't retry if creating the SSL context fails, likely a configuration issue with
@@ -404,7 +414,11 @@ class UserRestProxy:
                 try:
                     await producer.start()
                 except KafkaConnectionError:
-                    log.exception("Unable to connect to the bootstrap servers, retrying")
+                    if retry:
+                        log.exception("Unable to connect to the bootstrap servers, retrying")
+                    else:
+                        log.exception("Giving up after trying to connect to the bootstrap servers")
+                        raise
                     await asyncio.sleep(1)
                 else:
                     self._async_producer = producer
@@ -532,7 +546,7 @@ class UserRestProxy:
             return metadata
 
     def init_admin_client(self):
-        while True:
+        for retry in [True, True, False]:
             try:
                 self.admin_client = KafkaRestAdminClient(
                     bootstrap_servers=self.config["bootstrap_uri"],
@@ -550,7 +564,11 @@ class UserRestProxy:
                 )
                 break
             except:  # pylint: disable=bare-except
-                log.exception("Unable to start admin client, retrying")
+                if retry:
+                    log.exception("Unable to start admin client, retrying")
+                else:
+                    log.exception("Giving up after failing to start admin client")
+                    raise
                 time.sleep(1)
 
     async def aclose(self) -> None:
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -1050,6 +1050,8 @@ class UserRestProxy:
         partition_id = self.validate_partition_id(partition_id, content_type)
         try:
             KafkaRest.r(await self.get_offsets(topic, partition_id), content_type)
+        except TopicAuthorizationFailedError:
+            KafkaRest.r(body={"message": "Forbidden"}, content_type=JSON_CONTENT_TYPE, status=HTTPStatus.FORBIDDEN)
         except UnknownTopicOrPartitionError as e:
             # Do a topics request on failure, figure out faster ways once we get correctness down
             metadata = await self.cluster_metadata()
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -635,7 +635,7 @@ class UserRestProxy:
         publish_results = await self.produce_messages(topic=topic, prepared_records=prepared_records)
         for publish_result in publish_results:
             if "error" in publish_result and status == HTTPStatus.OK:
-                status = HTTPStatus.INTERNAL_SERVER_ERROR
+                status = HTTPStatus.UNPROCESSABLE_ENTITY
             response["offsets"].append(publish_result)
         KafkaRest.r(body=response, content_type=content_type, status=status)
 
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -37,6 +37,7 @@ KNOWN_FORMATS = {"json", "avro", "protobuf", "binary"}
 OFFSET_RESET_STRATEGIES = {"latest", "earliest"}
 SCHEMA_MAPPINGS = {"avro": SchemaType.AVRO, "jsonschema": SchemaType.JSONSCHEMA, "protobuf": SchemaType.PROTOBUF}
 TypedConsumer = namedtuple("TypedConsumer", ["consumer", "serialization_format", "config"])
+IDLE_PROXY_TIMEOUT = 5 * 60
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


tvainika:
This is missing integration tests for the new code. The integration tests for this feature need quite some work due setting up Kafka with SASL authorization to run the test against.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -134,37 +146,217 @@ def _add_kafka_rest_routes(self) -> None:
...
+                auth = aiohttp.BasicAuth.decode(auth_header)
+                config = self.config.copy()
+                config["bootstrap_uri"] = config["sasl_bootstrap_uri"]
+                config["security_protocol"] = (
+                    "SASL_SSL" if config["security_protocol"] in ("SSL", "SASL_SSL") else "SASL_PLAINTEXT"
+                )
+                config["sasl_mechanism"] = "PLAIN"
+                config["sasl_plain_username"] = auth.login
+                config["sasl_plain_password"] = auth.password
+                self.proxies[key] = UserRestProxy(config, self.kafka_timeout, self.serializer)
```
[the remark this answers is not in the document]
juha-aiven: What about the following.

Change `self.proxies` like this:

```python
self.proxies: Dict[str, Dict[str, "UserRestProxy"]] = {}
```

The outer key is the username and the inner key is the password.

I think we don't ever need to store the "old" proxy when the user changes his password, or do we? 

If not, `get_user_proxy` should work like this.

* If there's a proxy for the username & password tuple, return that
* If there's no proxy for the username, create and store a new one and return that
* If there's a proxy for the username, but for a different password, "close" the proxy and create a new like in the above case.

Old proxies will use the old password and should not be able to connect to Kafka, though not sure how Kafka authentication works for existing connections when the password is changed. But no matter how it works, we _shouldn't_ talk to Kafka with old credentials.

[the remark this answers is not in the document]
jjaakola-aiven: Maybe `Dict[Username, UserRestProxy]`:
```suggestion
maybe_valid_proxy = proxies.get(username)
if maybe_valid_proxy is None:
    # If proxy is None a new must be created
    proxy = create_user_proxy(...)
elif maybe_valid_proxy.get_pw_hash() != hash(auth.password):
    # If password has changed, close the old, create new
    maybe_valid_proxy.close()
    proxy = create_user_proxy(...)
    proxies[username] = proxy
else:
    # existing will do
    proxy = maybe_valid_proxy
```

karapace/kafka_rest_apis/__init__.py
```diff
@@ -134,37 +146,214 @@ def _add_kafka_rest_routes(self) -> None:
...
+    async def topic_details(self, content_type: str, *, topic: str, request: HTTPRequest):
+        proxy = self.get_user_proxy(request)
+        await proxy.topic_details(content_type, topic=topic)
+
+    async def topic_publish(self, topic: str, content_type: str, *, request: HTTPRequest) -> None:
+        proxy = self.get_user_proxy(request)
+        await proxy.topic_publish(topic, content_type, request=request)
+
+
+class UserRestProxy:
```
jjaakola-aiven, line 340: Could this track if there is auth error? I am thinking that the proxy could use backoff so it would not try constantly?

karapace/kafka_rest_apis/__init__.py
```diff
@@ -40,36 +49,28 @@ def __init__(self, config: Config) -> None:
...
-        self._cluster_metadata = None
-        self._metadata_birth = None
-        self.metadata_max_age = self.config["admin_metadata_max_age"]
-        self.admin_client = None
-        self.admin_lock = None
-        self.metadata_cache = None
-        self.schemas_cache = {}
-        self.consumer_manager = ConsumerManager(config=config)
-        self.init_admin_client()
+        self.proxies: Dict[str, "UserRestProxy"] = {}
```
jjaakola-aiven, line 52: Any consideration to have cleanup on rarely used proxies?

juha-aiven: And maybe set a high enough max number on how many proxies there can be to prevent excessive resource usage? And when that is reached and a new one is created, close and remove one of the proxies, maybe in the oldest in LRU order?

karapace/kafka_rest_apis/__init__.py
```diff
@@ -40,36 +49,28 @@ def __init__(self, config: Config) -> None:
...
-        self.admin_lock = asyncio.Lock()
-        self._async_producer_lock = asyncio.Lock()
+            for proxy in self.proxies.values():
+                stack.push_async_callback(proxy.aclose)
 
     def _add_kafka_rest_routes(self) -> None:
         # Brokers
-        self.route("/brokers", callback=self.list_brokers, method="GET", rest_request=True)
+        self.route(
+            "/brokers", callback=self.list_brokers, method="GET", rest_request=True, with_request=True, json_body=False
```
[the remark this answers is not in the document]
tvainika: Autoformatter would revert such change...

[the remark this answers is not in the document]
jjaakola-aiven:
```suggestion
            "/brokers", callback=self.list_brokers, method="GET", rest_request=True, with_request=True, json_body=False,
```
Added comma to the end of the line, autoformatter will split.

karapace/kafka_rest_apis/__init__.py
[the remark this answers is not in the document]
tvainika: At least cannot close old proxy unless authentication with new password succeeds, otherwise malicious calls cause havoc.

But even for normal usage, closing old proxies automatically also causes destruction of consumers & subscriptions in uncontrolled fashion. At least I haven't found ultimate solution for password change cases, which would not cause some oddities one way or another.

juha-aiven:
Karapace should log on startup which mode it's running in, i.e. whether `rest_authorization' is `True` or `False`. This would help debugging (reading logs), one would know what Karapace is doing. The logging is naturally slightly different in both cases, but it'd be nice to have this explicitly written in the log.

juha-aiven:
> Karapace should log on startup which mode it's running in, i.e. whether `rest_authorization' is `True`or`False`. This would help debugging (reading logs), one would know what Karapace is doing. The logging is naturally slightly different in both cases, but it'd be nice to have this explicitly written in the log.

We now log all the conf in #443, so no need for this anymore.

jjaakola-aiven (changes requested):

karapace/kafka_rest_apis/__init__.py
```diff
@@ -297,19 +524,26 @@ async def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
...
             try:
                 metadata_birth = time.monotonic()
                 metadata = self.admin_client.cluster_metadata(topics)
-                # Only store cache if fetched full metadata
-                if topics is None:
-                    self._metadata_birth = metadata_birth
-                    self._cluster_metadata = metadata
+                self._metadata_birth = metadata_birth
+                self._cluster_metadata = metadata
+                self._cluster_metadata_complete = topics is None
```
juha-aiven, line 543: What's the reasoning behind the following logic.

If `topics is not None`. `cluster_metadata` is called, `_cluster_metadata` is queried but only for the given topics. This value is cached but the cache is "not complete" and is not used if `cluster_metadata` is called with `topics is None`.

Imagine the following happens quickly after each other

```python
cluster_metadata() # All topics are queried and cached, image topic t is not there in the cache
cluster_metadata(topics=["t"]) # topic t is queried and cached
cluster_metadata() # All topics are queried and cached, since `_cluster_metadata_complete` is `False`
``` 

To me sounds the following would make more sense:

```python
cluster_metadata() # All topics are queried and cached, image topic t is not there in the cache
cluster_metadata(topics=["t"]) # topic t is queried, if found add that to the cached value. If not found, do not touch the cache.
cluster_metadata() # return cached value
```

I mean. If querying for a single topic, never store in the cache (bu. If querying for all topics,

karapace/kafka_rest_apis/__init__.py
```diff
@@ -297,19 +524,26 @@ async def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
             if self._metadata_birth is None or time.monotonic() - self._metadata_birth > self.metadata_max_age:
                 self._cluster_metadata = None
 
-            if self._cluster_metadata and topics is None:
-                return self._cluster_metadata
+            if self._cluster_metadata:
+                # Return from metadata only if all queried topics have cached metadata
```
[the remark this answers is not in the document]
juha-aiven: `cluster_metadata` is called either with `topics` `None` or with with a list of single topic `str`. Let's simplify the type of `topics` to `topic: Optional[str]`?

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -145,19 +146,6 @@ def _illegal_state_fail(message: str, content_type: str) -> None:
...
-        # This is (should be?) equivalent to calling poll on the consumer.
-        # which would return 0 results, since the subscription we just created will mean
-        # a rejoin is needed, which skips the actual fetching. Nevertheless, an actual call to poll is to be avoided
-        # and a better solution to this is desired (extend the consumer??)
-
-        # pylint: disable=protected-access
-        consumer._coordinator.poll()
-        if not consumer._subscription.has_all_fetch_positions():
-            consumer._update_fetch_positions(consumer._subscription.missing_fetch_positions())
-        # pylint: enable=protected-access
```
juha-aiven, line 159: Is this removed because not needed with aiokafka or for some other reason?

juha-aiven (changes requested):

karapace/kafka_rest_apis/__init__.py
tvainika: Only one proxy per **valid** user, password pair is ever created. (Thus password change for user will create a new one.) Therefore I don't know if a max limit is really needed as number of users in Kafka already is one limit. And in any case removal can be done only if there is no active consumers. But of course passive proxies keep Kafka connection open, thus disconnecting those (if only producer/metadata fetch, not consumer) does release some resources.

karapace/kafka_rest_apis/__init__.py
tvainika: Cache is always best effort. Topics are IMHO rarely created, but one can create more than one topic at time. Your suggestion would fail in such case as "all topics" returned from cache would not contain second topic if such is created same time. Also it would make cache update logic more complex.

karapace/kafka_rest_apis/consumer_manager.py
tvainika: Yes.

1. This uses kafka-python's internal implementation, not available in aiokafka.
2. As far as I understood aiokafka automatically does needed partition assignments after `subscribe` or `assign` (the cases when this internal method was used with kafka-python)

juha-aiven: > This uses kafka-python's internal implementation, not available in aiokafka

Got it.

karapace/kafka_rest_apis/__init__.py
juha-aiven: What about caching per `(user,password)`. And when a new entry is created for a `user` (with a valid user + password tuple), the older ones are invalid because they use an old password. And since they're invalid, could they be purged?

karapace/kafka_rest_apis/__init__.py
juha-aiven: I've been looking at this for a while, and it's just hard to understand:

* as said earlier, it's very odd that it sometimes caches partial results (only for the queried topics) and sometimes full results. I think it's much clearer if full results only are cached.
* I think the method implicitly raises `UnknownTopicOrPartitionError` if the topic is not found, at least that's many callers of the method handle. This is not clear in the current implementation.

Here's a more reabable idea. Basic idea: 1st check if the cache needs to be updated. If yes, update it. Then get the response from the cache. Always cache full metadata.

The downside I can think of is that is getting full metadata is more costly than getting for a single topic that doesn't exist in the cache. But if that happens the cache is fully refreshed, the follow-up calls for all topics or for that new topic are cheap because the data is in the cache. 

```python
    async def cluster_metadata(self, topic: Optional[str] = None) -> dict:
        async with self.admin_lock:
            refresh_cache = False
            if self._cluster_metadata is None:
                # First time or cache update has failed
                refresh_cache = True
            if time.monotonic() - self._metadata_birth > self.metadata_max_age:
                # cache expired
                refresh_cache = True
            elif topic is not None and topic not in self._cluster_metadata["topics"]:
                # Getting for a single topic that is not in the cache.
                # It's possible the topic has been created after the cache was populated.
                # Force cache update to not return false negative.
                refresh_cache = True

            if refresh_cache:
                self._cluster_metadata = None
                self._metadata_birth = None
                try:
                    self._cluster_metadata = self.admin_client.cluster_metadata()
                    self._metadata_birth = time.monotonic()
                except NodeNotReadyError:
                    log.exception("Could not refresh cluster metadata")
                    KafkaRest.not_found(
                        body={
                            "message": "Kafka node not ready",
                            "code": RESTErrorCodes.HTTP_INTERNAL_SERVER_ERROR.value,
                        },
                        content_type="application/json",
                    )

            if topic is None:
                return self._cluster_metadata
            else:
                topic_metadata = self._cluster_metadata["topics"].get(topic)
                if topic_metadata is not None:
                    return {
                        **self._cluster_metadata,
                        "topics": {topic: topic_metadata},
                    }
                else:
                    raise UnknownTopicOrPartitionError()
```

karapace/kafka_rest_apis/__init__.py
tvainika: There is now janitor task for idle proxies. It will release proxy after five minutes of idle if the proxy does not have consumers. If the proxy has consumers, then there is a configuration value which destructs consumers, the default for this is one hour, but feature can be disabled with the configuration value.

karapace/kafka_rest_apis/__init__.py
tvainika: IMHO the current implementation is shorter and more readable than your suggestion here.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -523,16 +524,23 @@ async def cluster_metadata(self, topics: Optional[List[str]] = None) -> dict:
...
-                # Only store cache if fetched full metadata
-                if topics is None:
-                    self._metadata_birth = metadata_birth
-                    self._cluster_metadata = metadata
+                self._metadata_birth = metadata_birth
+                self._cluster_metadata = metadata
+                self._cluster_metadata_complete = topics is None
             except NodeNotReadyError:
                 log.exception("Could not refresh cluster metadata")
                 KafkaRest.r(
```
jjaakola-aiven, line 598: Note, gateway error would be better here than internal server error.

tests/integration/test_rest.py
```diff
@@ -365,6 +379,17 @@ async def test_partitions(rest_async_client, admin_client, producer):
     res = await rest_async_client.get("/topics/fooo/partitions")
     assert res.status_code == 404
     assert res.json()["error_code"] == 40401
+
+    # Fill cache
+    res = await rest_async_client.get("/brokers")
```
jjaakola-aiven, line 384: Should the metadata cache tests be separate set of tests, possibly using the `topic_details` API?

karapace/kafka_rest_apis/__init__.py
```diff
@@ -51,16 +51,48 @@ def __init__(self, config: Config) -> None:
...
+        while True:
+            await asyncio.sleep(IDLE_PROXY_TIMEOUT / 2)
+
+            key, proxy = None, None
+            async with self._proxy_lock:
+                # Always clean one at time per iteration, don't mutate dict while iterating
+                for _key, _proxy in self.proxies.items():
+                    # If UserRestProxy has consumers with state, disconnecting loses state
+                    if _proxy.num_consumers() > 0:
+                        timeout = self.config.get("consumer_idle_disconnect_timeout", 0)
```
jjaakola-aiven, line 80: Would be nice to have a tests.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -51,16 +51,48 @@ def __init__(self, config: Config) -> None:
...
+                        if timeout > 0 and _proxy.last_used + timeout < time.monotonic():
+                            key, proxy = _key, _proxy
+                            log.warning("Disconnecting idle consumers for %s", _proxy)
+                            break
+                    # If there are no consumers, connection can be recreated without losing any state
+                    else:
+                        if _proxy.last_used + IDLE_PROXY_TIMEOUT < time.monotonic():
```
jjaakola-aiven, line 87: Not sure, but shouldn't this follow also the configuration option of disconnect time? If set to zero / cleaning disabled keep the proxy in the dictionary?

karapace/kafka_rest_apis/__init__.py
```diff
@@ -648,96 +858,97 @@ async def produce_messages(self, *, topic: str, prepared_records: List) -> List:
...
+            KafkaRest.not_found(
                 message=f"Topic {topic} not found",
                 content_type=content_type,
                 sub_code=RESTErrorCodes.TOPIC_NOT_FOUND.value,
             )
 
     async def partition_details(self, content_type: str, *, topic: str, partition_id: str):
-        self.log.info("Retrieving partition details for topic %s and partition %s", topic, partition_id)
+        log.info("Retrieving partition details for topic %s and partition %s", topic, partition_id)
         p = await self.get_partition_info(topic, partition_id, content_type)
```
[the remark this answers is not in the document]
tvainika: Yes/no, `get_partition_info` catches `TopicAuthorizationFailedError` and throws topic not found http level error.

karapace/kafka_rest_apis/__init__.py
tvainika: I went without configuration option for this because without consumers the proxy is stateless, ie. it can be recreated, and the API user does not see any difference.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -648,96 +858,97 @@ async def produce_messages(self, *, topic: str, prepared_records: List) -> List:
...
-        self.r(p, content_type)
+        KafkaRest.r(p, content_type)
 
     async def partition_offsets(self, content_type: str, *, topic: str, partition_id: str):
-        self.log.info("Retrieving partition offsets for topic %s and partition %s", topic, partition_id)
+        log.info("Retrieving partition offsets for topic %s and partition %s", topic, partition_id)
         partition_id = self.validate_partition_id(partition_id, content_type)
         try:
-            self.r(await self.get_offsets(topic, partition_id), content_type)
+            KafkaRest.r(await self.get_offsets(topic, partition_id), content_type)
```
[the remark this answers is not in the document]
tvainika: Good catch, missing from here. Seems also originally admin code did not handle errors with offsets, but returned dummy values...

karapace/kafka_rest_apis/__init__.py
tvainika: To test this with integration tests would require setting up Kafka with authorization. Most of the new code in this PR is not unit-testable, thus test coverage is quite low. To test this specific feature with the integration test, the test would have to wait up to minutes as this task is not so frequently checking for idle proxies, ie. 2.5 minutes per cycle. I'm kind out of tangible ideas how to really test this.

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -288,15 +276,23 @@ async def get_offsets(self, internal_name: Tuple[str, str], content_type: str, r
...
+                    )
+                if offset is None:
                     continue
                 response["offsets"].append(
                     {
                         "topic": tp.topic,
                         "partition": tp.partition,
-                        "metadata": commit_info.metadata,
-                        "offset": commit_info.offset,
+                        "metadata": "",
```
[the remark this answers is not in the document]
tvainika: aiokafka has issues with this, and I went with "not important, but compatible way". This needs to be addressed aiokafka upstream, and later add support here, but I didn't see any usage for this.

karapace/kafka_rest_apis/__init__.py
tvainika: Maybe address this separately, and see if there is anything similar in CSR from compatibility point of view?

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -208,21 +196,20 @@ async def create_consumer(self, group_name: str, request_data: dict, content_typ
...
 
     async def create_kafka_consumer(self, fetch_min_bytes, group_name, internal_name, request_data):
+        ssl_context = create_client_ssl_context(self.config)
         while True:
             try:
                 session_timeout_ms = self.config["session_timeout_ms"]
                 request_timeout_ms = max(
                     session_timeout_ms,
-                    KafkaConsumer.DEFAULT_CONFIG["request_timeout_ms"],
+                    305000,  # kafka-python's DEFAULT_CONFIG["request_timeout_ms"]
```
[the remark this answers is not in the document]
tvainika: This copies kafka-python's default, which the value there has been because aiokafka doesn't have this exposed, and I didn't want to import it from kafka-python here... I didn't get if you had any better suggestion for this.

karapace/kafka_rest_apis/consumer_manager.py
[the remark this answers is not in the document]
jjaakola-aiven: Thank you, explains this.

karapace/kafka_rest_apis/__init__.py
jjaakola-aiven: Sure, just notices this when reviewing and added as note.

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -461,7 +465,10 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
...
             poll_data = defaultdict(list)
             message_count = 0
-            while read_bytes < max_bytes and start_time + timeout / 1000 > time.monotonic():
+            # Read buffered records with calling getmany() with possibly zero timeout and max_records=1 multiple times
+            read_buffered = True
+            while read_bytes < max_bytes and (start_time + timeout / 1000 > time.monotonic() or read_buffered):
+                read_buffered = False
                 time_left = start_time + timeout / 1000 - time.monotonic()
                 bytes_left = max_bytes - read_bytes
                 LOG.info(
```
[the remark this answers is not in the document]
tvainika: Changed to debug level...

karapace/kafka_rest_apis/__init__.py
```diff
@@ -40,36 +50,70 @@ def __init__(self, config: Config) -> None:
...
 
-        self.app.on_startup.append(self.initialize_asyncio_locks)
+            for proxy in self.proxies.values():
+                stack.push_async_callback(proxy.aclose)
 
-    async def initialize_asyncio_locks(self, app: aiohttp.web.Application):  # pylint: disable=unused-argument
-        """The locks can be initialized when asyncio loop is available"""
-        self.admin_lock = asyncio.Lock()
-        self._async_producer_lock = asyncio.Lock()
+    async def _idle_proxy_janitor(self) -> None:
```
juha-aiven, line 69: Comments for `_idle_proxy_janitor`

* There are no tests nor other prove that this works as expected. In prod it will be hard to see if this is working as expected, since there's no user-visible error or something like that that would occur if this doesn't work.
* As a reviewer the code is not simple to read and reason about, there's nested `while`, `async with`, `for`, `if`, another `if`and inside there a break. Very  complex, and especially since there are no tests, hard to say whether this works or not.
* Two fixes available. 1) simplify the code so that it's easier to read and/or 2) add tests. Preferably both.
* I think the method could be split into two - 1) a function that gathers the key,proxy-tuples and 2) the delete close of these . The function can be pulled out from the `KafkaRest` class and unit tested.
* There's no reason except for potentially simpler code for "Always clean one at time". For the reader of the code it's unclear _why_ clean only one. Now this makes the reader wonder, is there a reason for this. Let's just clean all that need to be cleaned. That's what would be expected for such a cleaner.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -40,36 +50,70 @@ def __init__(self, config: Config) -> None:
...
+                            key, proxy = _key, _proxy
+                            log.info("Releasing unused connection for %s", _proxy)
+                            break
+                if key is not None:
+                    del self.proxies[key]
+            if proxy is not None:
+                try:
+                    await proxy.aclose()
+                except:  # pylint: disable=bare-except
+                    log.exception("Disconnecting proxy failure")
```
juha-aiven, line 97: Log the exception?

karapace/kafka_rest_apis/__init__.py
```diff
@@ -40,36 +50,70 @@ def __init__(self, config: Config) -> None:
...
-        """The locks can be initialized when asyncio loop is available"""
-        self.admin_lock = asyncio.Lock()
-        self._async_producer_lock = asyncio.Lock()
+    async def _idle_proxy_janitor(self) -> None:
+        IDLE_PROXY_TIMEOUT = 5 * 60
+        idle_consumer_timeout = self.config.get("consumer_idle_disconnect_timeout", 0)
+        while True:
+            await asyncio.sleep(IDLE_PROXY_TIMEOUT / 2)
+
+            key, proxy = None, None
```
juha-aiven, line 75: Maybe put one more `try` here. If something goes wrong in the code, I don't know what, the janitor will die. It'd be probably better if the exception would be logged and janitor would just sleep and try again.

juha-aiven (dismissed):
I think `_idle_proxy_janitor` needs simplification and/or tests.

karapace/kafka_rest_apis/__init__.py
tvainika: What do you mean?

juha-aiven: Ups, didn't notice it's `log.exception`, not `log.error`. All good.

karapace/kafka_rest_apis/__init__.py
tvainika: There is now simplification by splitting into two, and all catching try block

jjaakola-aiven (approved):
LGTM

Issue #428, opened 2022-07-19
JSON Schema Lookups Via `subject/<subject>` Failing


corinnafab:
# Problem

After upgrading Karapace versions, calls to look up a JSON schema version on the [ `subject/<subject>` POST endpoint](https://github.com/aiven/karapace/blob/746e65a0c65f52052648c4ef100b134a69a31074/karapace/schema_registry_apis.py#L853) no longer return the version of a known schema, instead they throw 40403 “Schema not found” errors.

Karapace python error stack: `ValidatedTypedSchema` > `parse` > `parse_jsonschema_definition`  > `json.loads` > `Throws`

# Investigation

I’ve included examples of calls on versions `2.1.3-3.3.0` below; each was sent the same request body but between version `3.1.0` and version `3.1.1` ([[diff](https://github.com/aiven/karapace/compare/3.1.0...3.1.1)](https://github.com/aiven/karapace/compare/3.1.0...3.1.1)) the expected behavior has broken and starts to throw `40403` errors. 

In each case I’m able to successfully publish a schema (via `subjects/<subject>/versions`), still get the expected output from `subjects/<subject>/version/<version>`, and `subjects/<subject>/version/<version>/schema` returns the schema I would expect to see. We’ve tried this with several versions of schema (ie set as `'true'`, with examples provided in the documentation, with simple strings, with objects, etc.) with the same result. 

I ran the same commands with an AVRO schema and *did* successfully run *all* of the commands above, including to `subject/<subject>` (example below), so this seems to be specific to JSON. 

 

**Initial POST to subject**

```
curl POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
  --data '{"schemaType": "JSON", "schema": "{\"type\": \"string\"}"}' \
  http://localhost:8081/subjects/jsontype-string/versions
```

**Then** **POST schema to subject to get version and id**

```
curl POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
  --data '{"schemaType": "JSON", "schema": "{\"type\": \"string\"}"}' \
  http://localhost:8081/subjects/jsontype-string/
```

### **2.1.3 //** ✅

```graphql
❯ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
  --data '{"schemaType": "JSON", "schema": "{\"type\": \"string\"}"}' \
  http://localhost:8081/subjects/jsontype-string/versions
{"id":7}%                                                                                           
[...]

PR #430, opened 2022-07-22
Fix JSON schema comparison when posting to subject

sebinsunny:
Fix for issue #428

JSON schema comparison was not validated correctly and threw a  “Schema not found” error when posting to a subject.
Posting a JSON schema
```
curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
  --data '{"schemaType": "JSON", "schema": "{\"type\": \"string\"}"}' \
  http://localhost:8081/subjects/local/versions http://localhost:8081/subjects/local

{"id": 1}{"id": 1, "schema": "{\"type\": \"string\"}", "schemaType": "JSON", "subject": "local", "version": 1}
```
Before
```
curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
--data '{"schemaType": "JSON", "schema": "{\"type\": \"string\"}"}' \
[http://localhost:8081/subjects/jsontype-string/](http://localhost:8081/subjects/jsontype-string/)
{"error_code": 40403, "message": "Schema not found"}
```

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -752,10 +752,11 @@ class KarapaceSchemaRegistryController(KarapaceBase):
             )
         for schema in subject_data["schemas"].values():
             validated_typed_schema = ValidatedTypedSchema.parse(schema["schema"].schema_type, schema["schema"].schema_str)
-            if (
-                validated_typed_schema.schema_type == new_schema.schema_type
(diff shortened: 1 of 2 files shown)


jjaakola-aiven (dismissed):
Could you add test?
I would test with two JSON schemas that have the same properties but in different order.

jjaakola-aiven (approved):

Issue #347, opened 2022-03-09
Schema key field format differs from Confluent SR


ivanyu:
# What happened?

Karapace uses this order for the key fields in Kafka records: `subject`, `version`, `magic`, `keytype`.
In Confluent SR, it's: `keytype`, `subject`, `version`, `magic` ([code](https://github.com/confluentinc/schema-registry/blob/52305272e4ef994bd61222b05382bfbb10091b39/core/src/main/java/io/confluent/kafka/schemaregistry/storage/SchemaKey.java#L26)).

This difference is significant because Kafka topic compaction (normally used on the `_schemas` topic) doesn't parse these key JSONs and compare them as byte strings. If a schema was written by CSR and later hard deleted by Karapace (i.e. a tombstone record with the null value was produced), the compaction will not happen and the old schema definition will remain in the topic.

It seems, this should not affect soft deletion and other operations, because to apply them, the JSON is parsed. However, the compaction still won't happen.

# What did you expect to happen?

Karapace would use the same order as Confluent SR.

hackaugusto:
Note: When this change is done, it should also take backwards compatibility with Karapace itself into account. Things can get messy, so my proposal is to run a "transaction" of sorts:

- on startup when schema reader is syncing. If the key format is incompatible:
  - add to a list of incompatible entries
- once startup is finished and the leader is elected. If the current node is leader:
  - First re-write all the existing entries with the compatible key format
  - Then send a tombstone for every entry with the old key format

the key format of the other messages need to be checked too

jjaakola-aiven:
Keys can be corrected when restoring: #447

PR #433, opened 2022-07-26
fix: Schema registry message key field format correction

jjaakola-aiven:

References: #347

Confluent Schema Registry message key uses field order of `keytype, subject, version, magic` and Karapace has used order of `subject, version, magic, keytype`. Kafka won't compact the topic correctly as the key field byte contents are different.

Confluent Schema Registry uses the most compact format of the JSON in the key, for example `{"keytype":"SCHEMA","subject":"foo","version":1,"magic":0}`. There is no whitespaces, indent nor linebreaks.

# Why this way

*First approach*
On Karapace startup the schema message topic is consumed and each key format checked. If format is not correct a tombstone message and corrected message is sent to the topic.

I used the approach @hackaugusto suggested in the ticket but decided to use the schema reader thread.
The limitation is that it cannot call the schema writing functions as those will trigger offset watching and would cause deadlock.
Another issue is with registries that have a lot of records and this has a significant memory pressure as invalid entries are first collected in memory and then processed. This was rejected and next approach taken.

*Second approach, current solution*
The correction processing is moved to background thread which is started when Karapace leader is ready. The correction is run per record and consumed from start. There is a special marker record sent to schemas topic when correction is run. This allows to identify if correction has already been run.

# Notes
 * Backup restore could also correct the keys for records.

### DIFF
--- a/karapace/key_format_corrector.py
+++ b/karapace/key_format_corrector.py
@@ -1,0 +1,219 @@
+"""
+karapace - Key correction
+
+Copyright (c) 2022 Aiven Ltd
+See LICENSE for details
+"""
+
+from aiokafka import AIOKafkaConsumer, AIOKafkaProducer
+from aiokafka.structs import TopicPartition
+from asyncio import Future
+from karapace.config import Config
+from karapace.typing import JsonData
+from karapace.utils import json_encode
+from typing import List, Optional
+
+import asyncio
+import json
+import logging
+import threading
+import time
+
+LOG = logging.getLogger(__name__)
+
+SCHEMA_KEY_ORDER = ["keytype", "subject", "version", "magic"]
+CONFIG_KEY_ORDER = ["keytype", "subject", "magic"]
+NOOP_KEY_ORDER = ["keytype", "magic"]
+
+CORRECT_KEY_ORDERS = [SCHEMA_KEY_ORDER, CONFIG_KEY_ORDER, NOOP_KEY_ORDER]
+
+
+def _get_key_in_correct_order(key: JsonData) -> bytes:
+    corrected_key = {
+        "keytype": key["keytype"],
+    }
+    if "subject" in key:
+        corrected_key["subject"] = key["subject"]
+    if "version" in key:
+        corrected_key["version"] = key["version"]
+    # Magic is the last element
+    corrected_key["magic"] = key["magic"]
+    return json_encode(corrected_key, sort_keys=False, binary=True, compact=True)
+
+
+class KeyFormatCorrector:
+    """Correct message key format to use "keytype, subject, version, magic" order of keys.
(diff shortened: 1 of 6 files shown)


juha-aiven (changes requested):
First set of comments. I didn't read the test code yet.

juha-aiven:
I think `KeyFormatCorrector` could be made slightly simpler by removing the flag whether the correction has been already run from that. `KeyFormatCorrector` could do exactly what its name state, correct the keys. And whether or not, the correction has been run should be outside of `KeyFormatCorrector` itself.

In practise
* `key_correction_done` flag would not exist. 
* If told start, it would start (if not running already). It's the problem of the caller if it has already been run.
* The marker handling (`_send_key_correction_done_marker`) would not be in `KeyFormatCorrector` but in the caller side. 
* When the `KeyFormatCorrector` is ready, it'd tell using a callback or some other mechanism to its "owner" that it's ready. That entity would then write the marker, i.e. call `_send_key_correction_done_marker`. 

The only problem I can see is that the caller would use a different Kafka producer to write the marker. And that could at least fail.

karapace/key_format_corrector.py
```diff
@@ -0,0 +1,219 @@
...
+    async def run_correction(self) -> None:
+
+        if self.failed:
+            LOG.warning("[KeyCorrector] Correction marked failed, not attempting again.")
+            return
+
+        latest_offset = 0
+        start_time = time.monotonic()
+        producer = AIOKafkaProducer(
+            bootstrap_servers=self.config["bootstrap_uri"],
```
tvainika, line 86: This will fail with secure kafka connection (ssl and/or sasl auth) because those parameters are not passed?

karapace/key_format_corrector.py
```diff
@@ -0,0 +1,219 @@
...
+            linger_ms=50,
+        )
+        consumer = AIOKafkaConsumer(
+            self.config["topic_name"],
+            bootstrap_servers=self.config["bootstrap_uri"],
+            group_id="key-corrector",
+            auto_offset_reset="earliest",
+            enable_auto_commit=False,
+            max_poll_records=5000,
+        )
```
tvainika, line 96: Same here, security & auth related config params not passed?

tvainika (commented):
I (also) think that key correction done should not be message in kafka topic. It can be internal signal.

Then bigger issue about my first reading: I didn't see any locking protection for mutations when key corrector is still running. Did I miss something? I'm thinking a scenario where after karapace is started and key corrector is running in another thread, then some new schema versions are registered or old versions deleted. These will be written with canonical order of keys, but then key corrector thread can re-write earlier messages from topic with canonical key order, which end up later in the topic than just registered mutations.

Issue #435, opened 2022-08-01
Schema in API response contains unknown field "compatibility"


Oduig:
# What happened?

Our Karapace Schema Registry, hosted with Aiven under the Kafka installation, is returning a field that should not be there according to the documentation, for a specific topic:

```curl -w '\n' -X GET https://avnadmin:redacted@redacted.aivencloud.com:10799/subjects/ourtopic-key/versions/latest
{"compatibility": "BACKWARD", "id": 44, "schema": "{\"fields\": [{\"name\": \"transactionid\", \"type\": \"string\"}], \"name\": \"OurTopicKey\", \"namespace\": \"com.example\", \"type\": \"record\"}", "subject": "ourtopic-key", "version": 1}
```

The field "compatibility" is not expected according to Schema Registry docs:
https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions-(versionId-%20version)

This causes the Confluent library kafka-schema-registry-client to break.

# What did you expect to happen?

There should not be a field `compatibility` in the schema.

# What else do we need to know?

We use Kafka 3.2 (latest) from Aiven. Updating the client library does not help - there is no version in history that supports the field "compatibility" as part of a schema.

It appears that it has something to do with the compatibility setting of the schema itself. The problem only exists when the topic has an explicit compatibility (not the global one). Changing the compatibility level back to the default value, or to NONE, unfortunately does not un-set the compatibility setting. In theory, we could clear the schema-level compatibility setting for an immediate fix using DELETE. But it looks like this API endpoint is disabled by Aiven's Schema Registry: 405: Method Not Allowed.

Even if we delete the schema with ?permanent=true, the underlying compatibility setting persists, so we can never un-set it after setting it.

I will continue to work on a workaround on our end, and also reported the issue via Aiven support. Due to the nature of the bug, I expect this is a bug in Karapace and not just in Aiven's Schema Registry. Please advise us if there is any quick fix we can attempt.

Oduig:
For completeness sake and Googlability, here is the stack trace from kafka-schema-registry-client.

```
22/08/01 14:33:29 ERROR ScalaDriverLocal: User Code Stack Trace: 
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "compatibility" (class io.confluent.kafka.schemaregistry.client.rest.entities.Schema), not marked as ignorable (4 known properties: "version", "schema", "id", "subject"])
[...]

PR #436, opened 2022-08-03
Remove the compatibility level information from the GET API response.

sebinsunny:
[PM-2036]

Fix for issue #435

The issue only occurs when we change the compatibility level for a subject. The fix is to remove the compatibility information from the GET API response.

[PM-2036]: https://aiven.atlassian.net/browse/PM-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -544,6 +544,8 @@ class KarapaceSchemaRegistryController(KarapaceBase):
 
         try:
             subject_data = await self.schema_registry.subject_version_get(subject, version)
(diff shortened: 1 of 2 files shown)


jjaakola-aiven (dismissed):
Fix is good, I propose to change the test a bit for stricter response checking.

jjaakola-aiven (approved):

PR #437, opened 2022-08-05
offset watcher: Use Condition with offset integer instead of Event dict.

jjaakola-aiven:

There are cases when data is fed to Karapace and offset watching is
enabled, namely future work of correcting record keys produced
by Karapace that are not compatible with Confluent Schema Registry.
This can create thousands of Event objects to OffsetWatcher that are
never consumed and removed and this leaks significant amount of memory.

The Condition object is used to monitor for predicate by producer
that offset is seen by consumer. The consumer side loop will call
offset_seen() for each consumed record and updates the OffsetWatcher
and notifies the waiting threads. Producer side will wait for the
expected offset for the timeout period.

# Why this way

Single Condition object can be used instead of Dict[int, Event].


tvainika (approved):
LGTM.

Much simpler code than original fix in #364.

Issue #442, opened 2022-08-19
 Protobuf Schema Lookups Via subject/<subject> Failing (also protobuf creation is not idempotent)


strokyl:
# What happened?

From version `3.1.0` to version `3.3.1` (the latest version at this ticket edition moment). The `subjects/<subject>` POST endpoint no longer returns the version of a known protobuf schema, instead, it throws a `40403` (Schema not found) error.
I also noticed while looking for this bug that POST `subjects/<subject>/versions` called twice with the same protobuf schema will create to version instead of just one.
This bug make `io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializer` fails.

# What did you expect to happen?

I think that when trying to create a new version with the same protobuf schema than the previous one we should not create a new version and just return the already existing id for this schema.
I also think user should be able to get the id of an existing protobuf version.

# What else do we need to know?

This issue is really similar to https://github.com/aiven/karapace/issues/428: (also maybe you should check the idempotent issue with JSON schema)
Here the script I used:
```bash
#!/bin/bash


protobuf_schema='syntax = \"proto3\";\n\nmessage Record {\nint32 id = 1;\nstring createdAt = 2;\nstring name = 3;\n}'

create_schema() {
	curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
		--data '{"schemaType": "PROTOBUF", "schema": "'"$protobuf_schema"'"}' \
		http://localhost:8081/subjects/protobuf_test/versions
}

retrieving_schema_id () {
	curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
		--data '{"schemaType": "PROTOBUF", "schema": "'"$protobuf_schema"'"}' \
		http://localhost:8081/subjects/protobuf_test/
	}

perform_test () {
	echo "Creating protobuf schema"
	create_schema 
	echo
	echo "Creating again same schema"
	create_schema 
	echo
	echo "retrieving back schema id"
	retrieving_schema_id
	echo
}

start_karapace() {
	VERSION=$1 docker-compose up -d >/dev/null 2>&1
	sleep 3 #wait karapace is really started
}

for version in 2.1.3 3.0.0 3.1.0 3.1.1 3.1.2 3.2.0 3.3.0 3.3.1
do
	echo "[START]"
	echo "Running test for version: $version"
	start_karapace "$version"
	perform_test
	echo "[STOP]"
	echo
	echo
done
```

The docker-compose file:
```yaml
version: '3'
[...]

PR #445, opened 2022-08-21
Fix PROTOBUF schema comparison when posting to subject

sebinsunny:
fix for issue #442
[PM-2111]

PROTOBUF schema comparison was not validated correctly and threw 40403 “Schema not found” errors when posting to a subject.

[PM-2111]: https://aiven.atlassian.net/browse/PM-2111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -19,6 +19,7 @@ from karapace.errors import (
     VersionNotFoundException,
 )
 from karapace.karapace import KarapaceBase
+from karapace.protobuf.compare_result import CompareResult
 from karapace.rapu import HTTPRequest, JSON_CONTENT_TYPE, SERVER_NAME
 from karapace.schema_models import SchemaType, TypedSchema, ValidatedTypedSchema
 from karapace.schema_registry import KarapaceSchemaRegistry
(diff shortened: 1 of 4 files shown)


karapace/schema_registry_apis.py
```diff
@@ -756,6 +757,10 @@ async def subjects_schema_post(
             validated_typed_schema = ValidatedTypedSchema.parse(schema["schema"].schema_type, schema["schema"].schema_str)
             if schema_type is SchemaType.JSONSCHEMA:
                 schema_valid = validated_typed_schema.to_dict() == new_schema.to_dict()
+            elif schema_type is SchemaType.PROTOBUF:
+                result = CompareResult()
+                validated_typed_schema.schema.compare(new_schema.schema, result)
+                schema_valid = result.is_compatible()
```
tvainika, line 763: `compare` + `is_compatible` are wrong here. Those compare against backward compatibility of schemas, not equality. E.g. addition of new (non-required) field is compatible change, but not equal schema anymore. This is wrong. But you correctly catched here, the original code seems to be wrong also...

PR #446, opened 2022-08-22
feature: detect message key format to use

jjaakola-aiven:

Prefer the canonical format on new installations and migrations from
Confluent Schema Registry.
Prefer Karapace key format on installations where this key format use
is detected. This applies to Karapace installations and installations
that have migrated from Confluent Schema Registry to Karapace and have
mixed key formats in schemas topic.

### DIFF
--- a/karapace/key_format.py
+++ b/karapace/key_format.py
@@ -1,0 +1,73 @@
+"""
+karapace - Key correction
+Copyright (c) 2022 Aiven Ltd
+See LICENSE for details
+"""
+
+from enum import Enum
+from karapace.typing import JsonData
+from karapace.utils import json_encode
+
+SCHEMA_KEY_ORDER = ["keytype", "subject", "version", "magic"]
+CONFIG_KEY_ORDER = ["keytype", "subject", "magic"]
+NOOP_KEY_ORDER = ["keytype", "magic"]
+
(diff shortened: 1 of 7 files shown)


karapace/key_format.py
```diff
@@ -0,0 +1,57 @@
...
+SCHEMA_KEY_ORDER = ["keytype", "subject", "version", "magic"]
+CONFIG_KEY_ORDER = ["keytype", "subject", "magic"]
+NOOP_KEY_ORDER = ["keytype", "magic"]
+
+CANONICAL_KEY_ORDERS = [SCHEMA_KEY_ORDER, CONFIG_KEY_ORDER, NOOP_KEY_ORDER]
+
+
+class KeyMode(Enum):
+    CANONICAL = 1
+    KARAPACE = 2
```
tvainika, line 21: It would be improvement to explain and rename this `KARAPACE` format, e.g. naming it as `OLD_KARAPACE` or `DEPRECATED_KARAPACE`.

tvainika (changes requested):
Looks good overall with the logic. Maybe some naming adjustment, and some part of information from commit message to be included as code comment to see why this keyformatter is there?

tvainika (approved):

PR #447, opened 2022-08-22
feature: do key correction on backup get and restore

jjaakola-aiven:

Format schema record keys in canonical format when doing a backup.
Format schema record keys in canonical format when restoring a backup.

The key formatting is restricted to `schemas` topic or when explicitly from configuration. The backup tool has dual use, it is not onyl used to backup Karapace but can be used to backup any Kafka topic.

### DIFF
--- a/karapace/schema_backup.py
+++ b/karapace/schema_backup.py
@@ -11,9 +11,11 @@ from kafka.errors import NoBrokersAvailable, NodeNotReadyError, TopicAlreadyExis
 from karapace import constants
 from karapace.anonymize_schemas import anonymize_avro
 from karapace.config import Config, read_config
+from karapace.key_format import KeyFormatter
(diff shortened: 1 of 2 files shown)


tvainika (changes requested):
The implementation looks technically good, but due to reformatting key when doing backup, it is risky. KeyFormatter ignores keys unknown to it. Therefore if this backup feature is wrongly used against other data using `force_key_correction` flag it actually loses data from backup as those other than expected keys of json object are eliminated. I think that needs to be changed. Either change KeyFormatter to also include unknown keys, or simply not do any reformatting when creating a backup file, only on restore. For restore the risk is not like this because the backup file can be restored again with different parameters if the flag is wrongly used.

tvainika (approved):

PR #448, opened 2022-08-23
Clean dependencies

jjaakola-aiven:

Removed aiohttp_socks and python_socks.
Moved aiohttp to direct dependencies.
Moved lz4 and requests to indirect dependencies.
Added python-snappy and zstandard to indirect dependencies as those
were in the setup.py for enabling these compression algorithms.

StatsClient comments for ignored exception in Sentry config also cleaned.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,14 +1,13 @@
 # PyPI dependencies
 accept-types==0.4.1
-aiohttp-socks==0.5.5
+aiohttp==3.8.1
 aiokafka==0.7.2
-jsonschema==3.2.0
-lz4==3.0.2
-requests==2.27.1
-networkx==2.5
-python-dateutil==2.8.2
-protobuf==3.19.4
 avro==1.11.0
+jsonschema==3.2.0
+networkx==2.5
+protobuf==3.19.4
+python-dateutil==2.8.2
+
 # Patched dependencies
 #
 # Note: It is important to use commits to reference patched dependencies. This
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -24,14 +24,12 @@
     install_requires=[
         "accept-types",
         "aiohttp",
-        "aiohttp_socks",
         "aiokafka",
         "avro",
         "jsonschema",
-        "kafka-python",
```
tvainika, line 31: Even if `kafka-python` gets installed as indirect dependency of aiokafka, it is also used as direct dependency e.g. admin client. I don't think it should be removed from here. And it should be also in requirements.txt?

requirements.txt
```diff
@@ -1,14 +1,13 @@
 # PyPI dependencies
 accept-types==0.4.1
-aiohttp-socks==0.5.5
+aiohttp==3.8.1
 aiokafka==0.7.2
+avro==1.11.0
 jsonschema==3.2.0
-lz4==3.0.2
-requests==2.27.1
```
tvainika, line 7: Requests library seems to be directly used in tests code, so instead of removing, moving this to requirements-dev.txt?

tvainika (commented):
I'm not sure if all those were correct. Please check my inline comments.

requirements.txt
jjaakola-aiven: Requests is in the indirect requirements. I'll add it also to in the dev requirements.

setup.py
jjaakola-aiven: I'll add it back, it needs to follow the version requirements of aiokafka.

requirements.txt
jjaakola-aiven: No, I won't add it back as the `kafka-python` is already in the file: https://github.com/aiven/karapace/pull/448/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R19

The `kafka-python` is installed from a fork.

setup.py
jjaakola-aiven: No, I won't add it back as the kafka-python is already in the file: https://github.com/aiven/karapace/pull/448/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R19

The kafka-python is installed from a fork.

tvainika (approved):

PR #451, opened 2022-08-24
Improve Karapace API compatibility with Confluent Schema Registry API

jjaakola-aiven:

Set of commits that improve Karapace API compatibility with Confluent Schema Registry.


tvainika (approved):
LGTM. Reviewed per commit and also together.

PR #452, opened 2022-08-26
Cache normalized string representation

tvainika:

Cache normalized string representation

# Why this way

Some use cases calculate hash of `TypedSchema` multiple times for fast
comparison.  Cache normalized string form of AVRO & JSON Schemas
similar to ProtoBuf for performance improvement.

This might need some follow-up later to use normalized forms of schemas more consistently.

### DIFF
--- a/karapace/schema_models.py
+++ b/karapace/schema_models.py
@@ -79,6 +79,8 @@ class TypedSchema:
         self.schema_str = schema_str
         self.max_id: Optional[int] = None
 
+        self._str_cached: Optional[str] = None
+
     def to_dict(self) -> Dict[str, Any]:
         if self.schema_type is SchemaType.PROTOBUF:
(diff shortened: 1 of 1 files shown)


jjaakola-aiven (approved):

karapace/schema_models.py
```diff
@@ -79,6 +79,8 @@ def __init__(self, schema_type: SchemaType, schema_str: str):
         self.schema_str = schema_str
         self.max_id: Optional[int] = None
 
+        self._str_cached: Optional[str] = None
```
juha-aiven, line 82: This doubles the memory usage? Does that matter?

tvainika: There are other memory usage and this does not apply to all schema types, so it does not double all memory usage. But definately this needs followed, and if there comes unifying changes for normalization of schemas, it reduces memory usage of original schema definitions here.

PR #453, opened 2022-08-26
metrics: report Karapace processing metrics

jjaakola-aiven:

* report count of processed records per keymode, counter
* report number of of processed records per keymode, gauge
* report number of of schemas and subjects, gauge
* report number of live and soft deleted schema versions, gauge

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -38,6 +38,14 @@ KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS = 2.0
 SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS = 5.0
 
 
+# Metric names
+METRIC_SCHEMA_TOPIC_RECORDS_PROCESSED_COUNT = "karapace_schema_reader_records_processed"
+METRIC_SCHEMA_TOPIC_RECORDS_PER_KEYMODE_GAUGE = "karapace_schema_reader_records_per_keymode"
+METRIC_SCHEMAS_GAUGE = "karapace_schema_reader_schemas"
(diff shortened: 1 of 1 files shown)


tvainika (commented):
The code looks good to me. However two things to check before merging:

1. Should interim gauges during startup reported? There is `ready` flag that could be easily used to skip gauge reporting during startup. E.g. if Karapace is restarted, current implementation would cause gauge metrics to drop, and then go back up.
2. Per subject version gauges: Is it too much data for metrics? If there are many subjects registered, and this increased cardinality of metrics. Is there need for per subject metrics?

jjaakola-aiven:
1. I would not interrupt, the gauge going up shows progress when starting.
2. There is two things to consider, one is the cardinality and you're right the `subject` tag is to be removed. Other is the actual calculation which loops through all subjects and schema versions. I am more concerned on the latter issue on larger schema registries.

karapace/schema_reader.py
```diff
@@ -285,9 +300,77 @@ def handle_messages(self) -> None:
...
+                self.stats.gauge(
+                    metric=METRIC_SUBJECT_DATA_SCHEMA_VERSIONS_GAUGE,
+                    value=versions,
+                    tags={"state": "live"},
+                )
+                self.stats.gauge(
+                    metric=METRIC_SUBJECT_DATA_SCHEMA_VERSIONS_GAUGE,
+                    value=soft_deleted_versions,
+                    tags={"state": "soft_deleted"},
+                )
```
tvainika, line 372: After removal of `subject` from tags, these calls should be called once, not per `for ... self.subjects.item()` loop iteration, right? Otherwise value overrides other subject's value.

tvainika (approved):

PR #455, opened 2022-09-02
Use Sentry SDK

jjaakola-aiven:

The Raven client is obsolete. Some refactoring is included to
separate the Sentry usage from StatsClient. StatsClient will use
either no-op Sentry client or actual Sentry client is Sentry SDK
is installed.

The Sentry SDK version to be greater than or equal to 1.6.0.

### DIFF
--- a/karapace/statsd.py
+++ b/karapace/statsd.py
@@ -9,11 +9,11 @@ Copyright (c) 2019 Aiven Ltd
 See LICENSE for details
 """
 from contextlib import contextmanager
+from karapace.sentry import get_sentry_client
 from typing import Any, Dict, Iterator, Optional
 
 import datetime
 import logging
-import os
 import socket
 import time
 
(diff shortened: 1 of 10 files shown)


karapace/statsd.py
```diff
@@ -23,80 +23,42 @@
...
-        self._send(metric, b"c", inc_value, tags)
+        send_tags = self._tags.copy()
+        send_tags.update(tags or {})
+        self._send(metric, b"c", inc_value, send_tags)
 
     def timing(self, metric: str, value: float, tags: Optional[Dict] = None) -> None:
-        self._send(metric, b"ms", value, tags)
+        send_tags = self._tags.copy()
+        send_tags.update(tags or {})
+        self._send(metric, b"ms", value, send_tags)
```
tvainika, line 51: Isn't this duplication of logic from `_send` that combines default tags?

tvainika (approved):
LGTM

PR #459, opened 2022-09-15
Fix key JSON compact form

tvainika:

Fix key JSON compact form

References: #458 triggered a case where JSON key was written in non-compact form.


tvainika:
This regression originates from refactoring in #419

juha-aiven (approved):
LGTM. There's no reason to not to use `compact` forrm. Thus this PR does fix the issue when `compact` was apparently accidentally set to `False` in `send_schema_message`.

PR #464, opened 2022-10-04
Use unittest.mock in REST proxy janitor tests and in serialization tests

jjaakola-aiven:

Improve REST proxy janitor tests with mocking.
Improve serialization tests by using mocks instead of stub objects.


tvainika:
This is more like question of scoping, but there is `MockClient` and `MockProtobufClient` also used in unittests currently. Those could use same kind of refactoring to use `unittest.mock` instead with same benefits. Do you think those could be included in same PR?

jjaakola-aiven:
@tvainika I can include.

jjaakola-aiven:
@tvainika Improved testing also by asserting the method calls. The tests will fail if unexpected calls are made.

PR #466, opened 2022-10-05
Fix Internal Server Error deleting unknown subject

tvainika:

Fix Internal Server Error deleting unknown subject

# Why this way

Catch exception if subject know found correctly, and add integration tests for this behaviour.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -613,6 +613,15 @@ class KarapaceSchemaRegistryController(KarapaceBase):
             try:
                 version_list = await self.schema_registry.subject_delete_local(subject=subject, permanent=permanent)
                 self.r(version_list, content_type, status=HTTPStatus.OK)
+            except (SubjectNotFoundException, SchemasNotFoundException):
+                self.r(
+                    body={
+                        "error_code": SchemaErrorCodes.SUBJECT_NOT_FOUND.value,
+                        "message": SchemaErrorMessages.SUBJECT_NOT_FOUND_FMT.value.format(subject=subject),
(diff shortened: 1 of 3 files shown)


karapace/schema_registry_apis.py
```diff
@@ -613,6 +613,15 @@ async def subject_delete(
             try:
                 version_list = await self.schema_registry.subject_delete_local(subject=subject, permanent=permanent)
                 self.r(version_list, content_type, status=HTTPStatus.OK)
+            except (SubjectNotFoundException, SchemasNotFoundException):
```
jjaakola-aiven, line 616: I think in case of `SchemasNotFoundException` the subject should be deleted regardless and operation to succeed.

jjaakola-aiven (changes requested):

karapace/schema_registry_apis.py
tvainika: The test case here triggers behaviour which without this change causes 500 Internal Server Error after hard-deleting a subject. I'm not ever sure what is the difference between `SubjectNotFoundException` and `SchemasNotFoundException`. It can be only internal structuring.

However in this case there is nothing to delete any more, ie. 404 not found is fine error message.

tvainika: Modified hard delete functionality to do more cleanup of internal data structures. In the test case extended here this another exception is no longer thrown, but I still kept this catch here for the safety.

Also related compatibility check API results were fixed in case of non-existing subject, and the test case extended there.

jjaakola-aiven (approved):

Issue #195, opened 2021-04-15
Handle schema references


lornajane:
I ran into an error using https://github.com/riferrei/srclient which is a Go client for Kafka with Confluent Schema Registry support. That (since v5.5 I believe) has support for schema references, however Karapace sees this as an unknown field and gives an error:

```
422 Unprocessable Entity: Unrecognized field: references
```

Looking at what the client is sending, it's an empty `[]` field named `references` at the same level as `schema` and `schemaType`.

What I expected: Karapace should ignore/tolerate this field, even if it doesn't have support for schema references quite yet.

lornajane:
Here's what the client sent to Karapace:

```
{
    "schema": "{     \"namespace\": \"io.aiven.example\",     \"type\": \"record\",     \"name\": \"MachineSensor\",     \"fields\": [         {\"name\": \"machine\", \"type\": \"string\", \"doc\": \"The machine whose sensor this is\"},         {\"name\": \"sensor\", \"type\": \"string\", \"doc\": \"Which sensor was read\"},         {\"name\": \"value\", \"type\": \"float\", \"doc\": \"Sensor reading\"},         {\"name\": \"units\", \"type\": \"string\", \"doc\": \"Measurement units\"}     ] } ",
    "schemaType": "AVRO",
    "references": []
}
```

(in case an example is useful)

alpreu:
Any updates? We would like to depend on karapace for license reasons but need to have feature compatibility regarding the schema-references

hackaugusto:
@alpreu currently we are not working on this feature. PRs are welcome :)

gklijs:
For https://crates.io/crates/schema_registry_converter references are also supported for almost a year now. It would be nice if they were supported by karapace as well.

perj:
confluent-kafka-python also sends `"references": []` for json or protobuf schemas.
```
confluent_kafka.schema_registry.error.SchemaRegistryError: Unrecognized field: references (HTTP status code 422, SR code 422)
```

ChristosMantzouranisMT:
Hi, a bit confused here. My understanding is that there is no support for schema references, so is there an alternative in order to avoid redundant embedded schemas?

gklijs:
@ChristosMantzouranisMT you can use Confluent Schema registry, if that's an option and you want to use references. Or you could build some tooling so you can embed them dynamically.

ChristosMantzouranisMT:
[...]

PR #473, opened 2022-10-20
Protobuf schema references support

jjaakola-aiven:

Add support for protobuf schema references

References: #195 

# Why this way

Initial work is in pull request #467 
Original #467 pull request description:
```
These changes add basic support of references/dependencies for protobuf schema type.
This code contains much common code which allows adding support of references/dependencies to other types.
Limitations:
The compatibility check does not support references/dependencies.
Serialization/Deserialization of protobuf does not support references/dependencies.
```


libretto:
@jjaakola-aiven do You think it is a good idea to develop two different ways of the same functionality? I just have already a workaround for compatibility(compare) functionality in the Instaclustr repository and planning to release it ASAP.

karapace/protobuf/proto_file_element.py
```diff
@@ -99,8 +99,13 @@ def __eq__(self, other: "ProtoFileElement") -> bool:  # type: ignore
...
 
-    def compare(self, other: "ProtoFileElement", result: CompareResult) -> CompareResult:
-
+    def compare(
+        self,
+        other: "ProtoFileElement",
+        result: CompareResult,
+        self_dependency_types: Optional[List[TypeElement]] = None,
+        other_dependency_types: Optional[List[TypeElement]] = None,
+    ) -> CompareResult:
```
libretto, line 108: We have a type for dependencies... and my code which I working on already use there exactly this type. please not that Your fixes will interfere with my code logic which I just debugging. You can see it at https://github.com/instaclustr/karapace/pull/24

jjaakola-aiven:
@libretto, this PR does not implement the same functionality in different way. Your PR and commit is the basis and my commits are on top of it. I understand this creates a conflict to your implementation, but I have to consider the total quality of the feature and therefore decided to add tests and relevant changes.

libretto:
Better add your tests to my latest code where most bugs (including reported by You) are fixed.

libretto:
> @libretto, this PR does not implement the same functionality in different way. Your PR and commit is the basis and my commits are on top of it. I understand this creates a conflict to your implementation, but I have to consider the total quality of the feature and therefore decided to add tests and relevant changes.

@jjaakola-aiven Actually, it did.  You did the same work which I did too You can see the code in my PR.  It adds to me more work for me for resolving conflicts and rewriting code which I did in time You did Your changes. You did something in a different way. For example, I have 2 classes one for References and one for Dependencies. I understand You like use List[Reference] instead of References class. It is ok and I will replace it (actually it is not a bad style because karapace has another example of using similar array classes, but ok). By logic, it is better to have Dependency in a separate class because dependencies are resolved references. I understand You want to improve the code style. But why You can not simply tell me to do that by pointing out what I must change? I understand You want to add more tests but why You cannot add them in separate files to avoid conflicts with my code?

PR #474, opened 2022-10-21
fix: correct schema id resolving

jjaakola-aiven:

Karapace 2.1.2 has compared Avro schemas by comparing the parsed
Avro schema objects. Karapace 3.3.2 compares schema strings instead
of the parsed object.
The commit contains a test with data that Karapace 2.1.2 may have
produced and verifies the expected behavior with Karapace 3.3.2 and
later versions.

Schemas with duplicate id are skipped and rollback record sent to
schemas topic.

# LImitations

When schema topic data has a single record for a subject and the records is considered as duplicate, the subject is not registered. If there is data in the actual data topic of the subject and it references the schema:
 * endpoint  `/schemas/ids/<SCHEMA_ID>` returns the equivalent schema registered from the first encountered record from the schemas topic.
 * in case when producer side registers the schema for the subject again, the subject gets created and version 1 of the schema registered. The schema will be the first encountered schema from the schemas topic.


karapace/schema_reader.py
```diff
@@ -185,8 +162,37 @@ def get_schema_id(self, new_schema: TypedSchema) -> int:
...
-        return self._hash_to_schema_id_on_subject.get(hash((subject, schema.__str__())), None)
+    def get_schema_id_if_exists(self, *, subject: Subject, schema: ValidatedTypedSchema) -> Optional[SchemaId]:
+        sid = self._hash_to_schema_id_on_subject.get(hash((subject, schema.__str__())), None)
+        if sid is not None:
+            return sid
+        # In case if the hashing does not return id check with parsed schema if
+        # equivalent schema can be found. In Avro e.g. "string" == {"type":"string"}.
+        # If found set the equivalent schema to cache.
+        maybe_id = None
+        for existing_id, existing_schema in self.schemas.items():
```
tvainika, line 173: This is missing locking. `get_schema_id_if_exists` can be called from API, but schema reader thread handling Kafka messages may also mutate `self.schemas` during iteration

tests/integration/test_schema.py
```diff
@@ -2903,3 +2947,266 @@ async def test_invalid_schema_should_provide_good_error_messages(registry_async_
...
+        assert existing_version == 1, "Should get existing version 1"
+        assert existing_schema_id == 2, "Should get existing id 2"
+
+        # Consumer requests the schema with id 2
+        res = await registry_async_client.get("/schemas/ids/2")
+        schema_res = res.json().get("schema")
+        assert schema_res == first_schema
+
+        # Consumer requests the schema with id 3
+        res = await registry_async_client.get("/schemas/ids/2")
```
tvainika, line 3169: Numbers do not match

tests/integration/test_schema.py
```diff
@@ -2903,3 +2947,266 @@ async def test_invalid_schema_should_provide_good_error_messages(registry_async_
...
+        admin_client.close()
+
+    first_subject = create_subject_name_factory("first_subject_record_with_good_id")()
+    first_message_key = {
+        "keytype": "SCHEMA",
+        "subject": first_subject,
+        "version": 1,
+        "magic": 1,
+    }
+    first_schema = '"string"'
```
tvainika, line 3034: Maybe renaming `first_schema` to something like `primitive_schema` so it is comparable to `schema_with type`? Also readability would be improved if both schemas are introduced first next to each other.

karapace/schema_record_producer.py
```diff
@@ -0,0 +1,161 @@
...
+class SchemaRecordProducer:
+    def __init__(self, config: Config, key_formatter: KeyFormatter, offsets_watcher: OffsetsWatcher) -> None:
+        self.kafka_timeout = 10
+
+        self.config = config
+        self.producer = self._create_producer()
+        self.key_formatter = key_formatter
+        self.offsets_watcher = offsets_watcher
+
+    def _create_producer(self) -> KafkaProducer:
```
tvainika, line 28: I know this is just refactoring of the code, but how much work it would be to switch using aiokafka here? Quick look seemed to be that all these messages are already produced from async methods as caller.

tvainika (changes requested):
At least the locking issue is a blocker, other notes less important

karapace/schema_record_producer.py
jjaakola-aiven: I'd recommend not to refactor more in this PR.

tests/integration/test_schema.py
jjaakola-aiven: The issue here is not the primitive but the order of schemas, therefore I named as first and second etc.

tests/integration/test_schema.py
jjaakola-aiven: Left over, will remove.

karapace/schema_reader.py
jjaakola-aiven: Fixed, added a function in `schema_registry` to handle locking when called from `schema_registry_api`.

tvainika: This is now protected by `asyncio.Lock` but I think schema registry API and schema reader's message handler are running in two different threads. Therefore protection should be using `threading.Lock`?

jjaakola-aiven: Done.

jjaakola-aiven:
Closing this, outdated.

PR #478, opened 2022-10-27
feature: forward requests to primary node when not ready

jjaakola-aiven:

When Karapace is not ready and still replaying the state from `_schemas` the instance should forward every request to primary node.

### DIFF
--- a/karapace/karapace.py
+++ b/karapace/karapace.py
@@ -8,13 +8,13 @@ See LICENSE for details
 from functools import partial
 from http import HTTPStatus
 from karapace.config import Config
-from karapace.rapu import HTTPResponse, RestApp
-from typing import NoReturn, Union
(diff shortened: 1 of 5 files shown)


juha-aiven (commented):
It looks it's proposed that we do the forwarding in 1 + N places, this new one (1) that is global and then in each existing "write"-handler (N) such as `subject_delete`. Could these be unified so that this global logic would do all the forwardings and the per-handler ones would be removed? The global handler could look at the URL and forward in the cases when the local Karapace is not ready and in the case of a write operation also when the local Karapace is not the master. And it would fail if no `master_url`.

tvainika (dismissed):
There seems to be a failure with logic in case running single instance Karapace which is not ready: It will forward requests back to itself if I'm reading this right...

tvainika (approved):

PR #484, opened 2022-11-03
Handling None properly during backup

AnatolyPopov:

A follow-up PR for https://github.com/aiven/karapace/pull/472
Putting `null` string instead of `b''` to be able to properly restore from the backup.


jjaakola-aiven (dismissed):

karapace/schema_backup.py
```diff
@@ -240,9 +240,9 @@ def export(self, export_func) -> None:
             LOG.info("Schema export written to stdout")
         self.close()
 
-    def encode_key(self, key: Optional[Union[JsonData, str]]) -> bytes:
-        if not key:
-            return b""
+    def encode_key(self, key: Optional[Union[JsonData, str]]) -> Optional[bytes]:
+        if key == "null":
```
jjaakola-aiven, line 244: I would prefer null byte for null data. And would solve the literal string "null" in the `_restore_backup_version_1` as that is the source for those.
Essentially means that both internal restore functions would need to handle the key and value encoding, and pass key and value as separate arguments to `_handle_restore_message`.

tests/integration/test_data/test_restore_non_schema_topic_v2.log
```diff
@@ -3,3 +3,5 @@
 6B65795F74776F	76616C75655F74776F
 	76616C75655F7468726565
 6B65795F666F7572	
+6B65795F666F7572	null
```
jjaakola-aiven, line 6: With null byte encoded to b16 this line would be `6B65795F666F7572\t00`

karapace/schema_backup.py
jjaakola-aiven: My proposal to use null byte is not feasible. It would make impossible to distinguish between actual null byte payload and tombstone marker. The `null` string is good here as it can be safely recognized as actual null payload.

jjaakola-aiven (approved):

PR #486, opened 2022-11-09
Primary node selection prefers ready node

jjaakola-aiven:

The primary node selection prefers node that is ready, has replayed the schemas topic. The selection uses the strategy, i.e. lowest ready or highest ready. Selected node will stay as primary until a new group selection is run and new (or the same) node is selected as primary based on the strategy.

Fallback will be to select primary based on the strategy from the full set of nodes.

# Why this way

In case when there is restarts in the Karapace cluster the writing is enabled faster when first ready node is selected as primary.
When a Karapace node is not ready it has been possible to select unready node as a primary. If the schemas topic to replay is large it may take significant amount of time for Karapace to become ready.

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -30,7 +30,7 @@ def get_member_url(scheme: str, host: str, port: str) -> str:
     return "{}://{}:{}".format(scheme, host, port)
 
 
-def get_member_configuration(*, host: str, port: int, scheme: str, master_eligibility: bool, ready: bool) -> JsonData:
+def get_member_configuration(*, host: str, port: int, scheme: str, master_eligibility: bool) -> JsonData:
     return {
         "version": 2,
         "karapace_version": __version__,
@@ -38,7 +38,6 @@ def get_member_configuration(*, host: str, port: int, scheme: str, master_eligib
         "port": port,
         "scheme": scheme,
         "master_eligibility": master_eligibility,
-        "ready": ready,
     }
 
 
@@ -63,7 +62,17 @@ class SchemaCoordinator(BaseCoordinator):
         self.port = port
         self.scheme = scheme
         self.master_eligibility = master_eligibility
-        self.schema_reader_ready = False
+        self._schema_reader_ready = False
+        self._need_rejoin = False
+
+    def schema_reader_ready(self, ready: bool) -> None:
+        self._schema_reader_ready = ready
+        if self.rejoining:
+            LOG.debug("Rejoin ongoing, mark needing rejoin.")
+            self._need_rejoin = True
+        else:
+            LOG.debug("Rejoin requested based on readiness change.")
+            self.request_rejoin()
 
     def protocol_type(self) -> str:
         return "sr"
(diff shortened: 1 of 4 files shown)


karapace/master_coordinator.py
```diff
@@ -64,7 +66,7 @@ def _perform_assignment(self, leader_id, protocol, members):
         fallback_urls = {}
         for member_id, member_data in members:
             member_identity = json.loads(member_data.decode("utf8"))
-            if member_identity["master_eligibility"] is True:
+            if member_identity["master_eligibility"] is True and member_identity.get("ready", False) is True:
```
jjaakola-aiven, line 69: Note the default `False` for ready. Effectively this means that Karapace that does not provide `ready` is never allowed to become a primary as the state of those is unknown. This is important to know in scenarios where rolling update to newer version is done.

juha-aiven:
`election_strategy` `lowest` would no longer mean the lowest, but lowest ready? This is functional change. Also the naming is no more logical. 

Do we want to change existing functionality? To not change existing functionality, there could be

* new strategy `lowest_ready` and `lowest` would work as so far, OR
* a new flag enabling this "ready"-filtering

jjaakola-aiven:
The strategy as currently is broken as non-ready can be selected as primary. I would not like add a new name or new flag for enabling ready filtering. Fallback is to select lowest as a primary if there is no ready nodes available. I can add one more branch to selection to favor nodes that are eligible for primary role.

tvainika (changes requested):
This needs rebase to pass CI (due ci check upstream brokeness)

tvainika (commented):
This needs a bit discussions before I can approve things even if I did not notice anything wrong with the approach.

On my observation: Schema reader becoming ready does **not** trigger primary selection again. In other words if there is Karapace node joining, which is by url sorting designed to become the primary, but when it joins, it marks itself as not yet ready, then it will take a long time - actually some new node joining or existing leaving - until this designated node becomes primary due ready flag not triggering re-election. I think this is fine if intentional. What do you think?

Also another observation is that `ready` flag overall could be simply replaced with `master_eligibility` flag being True only if also node is ready regardless of setting, but this is of course opinionated if one wants explicit ready flag or not.

tvainika:
> Also another observation is that `ready` flag overall could be simply replaced with `master_eligibility` flag being True only if also node is ready regardless of setting, but this is of course opinionated if one wants explicit ready flag or not.

A thought related to this: if group coordinator runs older version of Karapace, then incorporating readiness into `master_eligibility` would actually make safer choices as primary, because older Karapace would look from `master_eligibility: False` that non-ready node cannot be assigned as primary.

karapace/schema_reader.py
```diff
@@ -309,6 +309,8 @@ def handle_messages(self) -> None:
         raw_msgs = self.consumer.poll(timeout_ms=self.timeout_ms)
         if self.ready is False:
             self.ready = self._is_ready()
+            if self.master_coordinator:
+                self.master_coordinator.ready(True)
```
tvainika, line 313: The logic fails if `self._is_ready()` returns `False` as this has explicit `True` value passed as readiness.

Comparison for `self.master_coordniator` elsewhere has `is not None` explictly.

karapace/master_coordinator.py
tvainika: After latest refactoring the `member_identity` struct no longer contains `ready` flag at all...

PR #490, opened 2022-11-11
Fix asynch fetch polling without delay

### LINKED ISSUE #488: Karapace 3.4.x consuming all CPU
# What happened?

In Karapace 3.4.x it is possible to send HTTP requests to the Kafka REST Proxy causing Karapace process to take almost all CPU (a single core). Additionally Karapace causes high load ~30-50% in Kafka. The situation remain even if the process is left alone (no traffic sent to it) for hours.

The issue was found out when running some manual testing using consumers that are not clean up correctly. The exact steps are quite hard to explain here, since the tester and the exact test steps are a part of a larger project that cannot be shared here,

Karapace 3.4.x has the issue. It's unknown whether older versions have it.

tvainika:

Fix asynch fetch polling without delay in REST proxy

This is regression from #423 when REST consumer was ported from kafka-python to aiokafka.

Fixes #488

# Why this way

First fetch min bytes negative values was misconfiguration even for kafka-python earlier, but worsened with aiokafka.  Add extra configuration for aiokafka timeout settings tuning.

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -35,7 +35,7 @@ DEFAULTS = {
     "consumer_request_timeout_ms": 11000,
     "consumer_request_max_bytes": 67108864,
     "consumer_idle_disconnect_timeout": 0,
-    "fetch_min_bytes": -1,
+    "fetch_min_bytes": 1,
     "group_id": "schema-registry",
     "host": "127.0.0.1",
     "port": 8081,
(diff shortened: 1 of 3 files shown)


karapace/config.py
```diff
@@ -35,7 +35,7 @@
     "consumer_request_timeout_ms": 11000,
     "consumer_request_max_bytes": 67108864,
     "consumer_idle_disconnect_timeout": 0,
-    "fetch_min_bytes": -1,
+    "fetch_min_bytes": 1,
```
jjaakola-aiven, line 38: This is correct, valid values are `> 0` and default is `1`.
https://kafka.apache.org/documentation/#consumerconfigs_fetch.min.bytes

jjaakola-aiven (approved):
I testes manually and by setting the `fetch.min.bytes` to `-1` the CPU issue can be reproduced.
With the fix proposed here I cannot reproduce the issue.

PR #493, opened 2022-11-21
Allow specifying version directly during build

ivanyu:

Sometimes the Git metadata is not available during build. In this case, we can rely on externally specified version using the `KARAPACE_VERSION` environment variable.


ivanyu:
Hint: hiding whitespaces in the review view makes the diff much smaller visually.

jjaakola-aiven (approved):

PR #497, opened 2022-11-25
chore: bump mypy, move config from pre-commit

aiven-anton:

Bumps mypy to latest version, 0.991.

Config for mypy is moved from the pre-commit hook definition to mypy.ini, and in addition `pretty = True` is set to show more human friendly error messages.

# Why this way

Keep project up-to-date.


jjaakola-aiven (approved):
:crown:

PR #503, opened 2022-12-08
INS-20428 | Karapace Rest Proxy goes in infinite loop while retrieving broker details

debaditya-netapp:
The fix is to stop KafkaAdminClient from going on an infinite loop while fetching broker metadata.

# Why this way

The sequence of events which cause this problem are below:

1. connection to bootstarp-0 is being made.
2. connection to Node-1 is being made.
3. bootstarp-0 is connected. MetaData request is still not in-flight.
4. Node-1 is connected.
5. Node-1 closes all bootstrap node connections. (See: https://github.com/aiven/kafka-python/blob/9feeb79140ed10e3a7f2036491fc07573740c231/kafka/client_async.py#L301)
6. Connection to Node-1 is closed due to being idle or no request in-flight and an EVENT_READ is registered(see https://github.com/aiven/kafka-python/blob/9feeb79140ed10e3a7f2036491fc07573740c231/kafka/client_async.py#L666)
7. And the process repeats.

We believe this is because of the overridden implementation of _maybe_refresh_metadata in Karapace(see: https://github.com/aiven/karapace/blob/fdfcb73c2e649f4ab61ec96a7ad154dac7f7ccae/karapace/utils.py#L174)
This is always preferring bootstarp node over least loaded node. Hence causing this problem (due to step 5).

The fix for this would be just to use the _maybe_refresh_metadata from aiven/kafka-python (see: https://github.com/aiven/kafka-python/blob/9feeb79140ed10e3a7f2036491fc07573740c231/kafka/client_async.py#L807)
which always makes the request to the least loaded node.


Claudenw:
While this seems OK I note that the removed code has the comment `Lifted from the parent class with the caveat that the node id will always belong to the bootstrap node,  thus ensuring we do not end up in a stale metadata loop`, how does the new code ensure that we do not end up in a stale metadata loop?

debaditya-netapp:
> While this seems OK I note that the removed code has the comment `Lifted from the parent class with the caveat that the node id will always belong to the bootstrap node, thus ensuring we do not end up in a stale metadata loop`, how does the new code ensure that we do not end up in a stale metadata loop?

Hi @Claudenw , 
While going through the commit which introduces this change, I found this was made to "fix a bug where metadata calls would end up continuously being made to no longer valid connections, resulting in a stale metadata" (commit: 4335886e597a1eaa6d55f2837e60a4f30bc09b62).
And I also found a later commit in aiven/kafka-python which introduces `_send_request_to_least_loaded_node`, where it is being checked "if the target node is
available in the cluster metadata" after every poll() (commit: 726066465e5effa6d3b182844b3037d4ded30644).
So I think the later already safeguards the problem with least_loaded_node, hence it would be alright not to use the overridden implementation in Karapace.

Claudenw (approved):
Looks good to me.  Thanks for the explanation.

debaditya-netapp:
Hi @tvainika ,
While https://github.com/aiven/kafka-python/commit/1b95333c9628152066fb8b1092de9da0433401fd safeguards against stale cluster metadata loop problem, this does not solve the problem I mentioned in this PR. 
The problem in this case occurs from consistently trying to make  a connection to bootstrap node and other nodes in parallel, where connection to any non bootstrap node, once connected, kills all bootstrap node connections in the connection callback method(https://github.com/aiven/kafka-python/blob/8006e65a3f92d0131441600a3e875459a1b3cffb/kafka/client_async.py#L303) and thereby going in an infinite loop(more on this in the PR description).
I think as karapace now uses https://github.com/aiven/kafka-python/commit/1b95333c9628152066fb8b1092de9da0433401fd, we can safely use the _maybe_refresh_metadata from aiven/kafka-python instead of the overridden implementation in karapace. Especially as the overridden implementation was only preferring bootstrap node without any checks to make sure if the bootstrap node is part of the broker metadata.

tvainika (approved):

debaditya-netapp:
Thanks @tvainika . Could you please tell me when can we expect this to be released?

tvainika:
@debaditya-netapp Karapace 3.4.3 has been finally released containing this fix.

PR #508, opened 2022-12-09
fix: Karapace instance primary eligiblity added to group protocol

jjaakola-aiven:

Group coordinator leader is selected from the whole set of members of the group. This can also be member that is not primary eligible. This change corrects the group protocol data and Karapace primary selection to use the actual instance configured primary eligibility. The code before used the primary eligibility of the selected group leader which can be false.

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -9,11 +9,12 @@ from kafka.coordinator.base import BaseCoordinator
 from kafka.errors import NoBrokersAvailable, NodeNotReadyError
 from kafka.metrics import MetricConfig, Metrics
 from karapace import constants
+from karapace.config import Config
 from karapace.typing import JsonData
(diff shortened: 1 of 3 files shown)


tvainika (commented):
I can see that this is kind of minimal changes needed to fix the issue found. However what about considering also method naming and signatures while we go? I find it confusing that we have `get_identity` and `get_identity_url` which I mix with each other, or is it just me? Second note is about type signature (return type) of `get_identity_url`: It has two modes with one returning json and another dict. What about separating those cases so that one can add valid return type annotation?

jjaakola-aiven:
Removed the dual-return-type.
Change the naming to `get_member_url` and `get_member_configuration`.

jjaakola-aiven:
Added also typing, not perfect some casts left. The configuration should be moved to dataclass from plain dictionary/JSON.

tvainika (approved):

PR #516, opened 2022-12-29
Disable Avro name and enum validation when parsing existing schemas

jjaakola-aiven:

When parsing existing Avro schemas validation for name and enum default is disabled. This is for compatibility as it has been possible to produce data that is not compliant with Avro specification.

Include also changes for future Avro release compatibility:
 * Error message format
 * new exception in Avro for invalid binary encoding.

# Why this way

The Avro SDK dependency is taken from Aiven fork of Avro. When upstream Avro releases the support for validation disable it is possible to move the dependency back.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@
 accept-types==0.4.1
 aiohttp==3.8.3
 aiokafka==0.7.2
-avro==1.11.0
 jsonschema==3.2.0
 networkx==2.5
 protobuf==3.19.5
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -22,6 +21,9 @@ watchfiles==0.15.0
 #
 git+https://github.com/aiven/kafka-python.git@1b95333c9628152066fb8b1092de9da0433401fd
 
+git+https://github.com/aiven/avro.git@branch-1.11#subdirectory=lang/py
```
tvainika, line 24: Everything else looks good, but IMHO this should point to git sha, not branch name (as branch can evolve) ie. replacing `branch-1.11` here with current branch head sha1sum. `branch-1.11` can be used as a comment here explain why this optionally. What do you think?

tvainika (approved):

PR #518, opened 2023-01-16
fix: use sane sleep on primary coordinator main loop

jjaakola-aiven:

The primary coordinator main loop shall not use zero sleep when heartbeat has been missed. Use half of the heartbeat interval in that case.

Issue symptom is visible with debug log level when `We're master ...` is logged multiple times.

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -227,6 +227,8 @@ class MasterCoordinator(Thread):
             try:
                 if not self.kafka_client:
                     if self.init_kafka_client() is False:
+                        # If Kafka client is not initialized sleep a bit
+                        time.sleep(0.5)
                         continue
                 if not self.sc:
                     self.init_schema_coordinator()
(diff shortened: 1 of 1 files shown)


karapace/master_coordinator.py
```diff
@@ -236,7 +238,11 @@ def run(self) -> None:
                 self.sc.ensure_active_group()
                 self.sc.poll_heartbeat()
                 LOG.debug("We're master: %r: master_uri: %r", self.sc.are_we_master, self.sc.master_url)
-                time.sleep(min(_hb_interval, self.sc.time_to_next_heartbeat()))
+                # In cases when heartbeat is missed the sleep min sleep time would be 0
+                # from `time_to_next_heartbeat`. In that case halve the heartbeat interval for
+                # some sane sleep instead of running the loop without sleep for a while.
+                sleep_time = min(_hb_interval, self.sc.time_to_next_heartbeat()) or _hb_interval / 2
```
juha-aiven, line 244: Could this be written as an `if`-statement for readability reasons?

jjaakola-aiven: Done.

juha-aiven (approved):
Looks good to me - all code paths in the `while`-loop do `time.sleep()` (for non-zero) time.

PR #519, opened 2023-01-17
chore: update GitHub action versions

jjaakola-aiven:

Node 12 is deprecated and removed soon. This affects multiple GitHub actions.

The Protobuf compiler action is not updated, it is missing the release for Node 16 support.


RyanSkraba (approved):
LGTM, thanks for the bumps!

PR #520, opened 2023-01-19
Fix `allocate_port()` on Mac

Fleshgrinder:
The `PortRangeInclusive.allocate_port()` function uses `psutil` under the hood to find out if a port is allocatable. This leads to an `AccessDenied` error on Mac. Effectively making it impossible to run any of the integration tests.

This patch changes the implementation to not rely on `psutil` at all, it also removes it from the dependencies since this is the only usage. Instead, we simply try to bind to a port and see if it is free, while keeping track of the ports we already predestined for binding by one of our own components.

This does not fix any possible race conditions that might arise from trying to find a free port on a system where other processes are running that might bind to ports. There is no way to achieve this since any free port can be bound by any process. The only way to achieve this would be if the software in question itself has direct support for finding a suitable port. It would be possible to add this functionality to Karapace (e.g., if port is set to `-1` automatically find a suitable one). However, I did not go in that direction, as it would require extensive changes to various parts. The patch is only meant to allow integration test execution on Mac, without any visible API changes.

A detail I think worth mentioning that might not be clear from the patch itself is that it removes the `maybe_available` collection in favor of a `bound` collection. The `maybe_available` collection is created from the given `start` and `end` parameters, that range might be big (up to 64,511 integers to be precise) and consume an unreasonable amount of memory. The `bound` approach does the opposite of the `maybe_available` and keeps track of those ports that were identified as being free, hence, the collection is small throughout the entire execution time.

Last, but not least, the `port_is_listening()` was inline as it had only a single usage and the `ipv6` flag was removed as it was unused. IPv6 support would require a more global approach for enabling if needed.

### DIFF
--- a/tests/integration/utils/network.py
+++ b/tests/integration/utils/network.py
@@ -2,82 +2,18 @@
 Copyright (c) 2022 Aiven Ltd
 See LICENSE for details
 """
-from contextlib import contextmanager
-
-import psutil
 import socket
 
 
-def is_time_wait(port: int) -> bool:
-    """True if the port is still on TIME_WAIT state."""
-    return any(conn.laddr.port == port for conn in psutil.net_connections(kind="inet"))
+def inet_port() -> int:
+    """Returns a free and usable IPv4 port.
 
-
-class PortRangeInclusive:
-    PRIVILEGE_END = 2**10
-    MAX_PORTS = 2**16 - 1
-
-    def __init__(
-        self,
-        start: int,
-        end: int,
-    ) -> None:
-        # Make sure the range is valid and that we don't need to be root
-        assert end > start, "there must be at least one port available"
-        assert end <= self.MAX_PORTS, f"end must be lower than {self.MAX_PORTS}"
-        assert start > self.PRIVILEGE_END, "start must not be a privileged port"
-
-        self.start = start
-        self.end = end
-        self._maybe_available = list(range(start, end + 1))
-
-    def next_range(self, number_of_ports: int) -> "PortRangeInclusive":
-        next_start = self.end
-        next_end = next_start + number_of_ports
-        return PortRangeInclusive(next_start, next_end)
-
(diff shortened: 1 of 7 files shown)


tests/integration/utils/network.py
```diff
@@ -3,81 +3,72 @@
...
-    except socket.error:
-        return False
+        sock = socket.socket()
+        for port in range(self.__start, self.__end + 1):
+            # We gave this port to a previous caller, and thus cannot use it, even if it is free. This is always a small
+            # collection, as we are not expecting tests to bind to hundreds of ports.
+            if port not in self.__bound:
+                self.__bound.add(port)
+                try:
+                    sock.bind(("127.0.0.1", port))
```
jjaakola-aiven, line 64: What about something like this, without looping and would also remove the need for min and max port?
```suggestion
import closing
with closing(socket.socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
    sock.bind(("127.0.0.1", 0))
    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    port = sock.getsockname()[1]
yield port
```

Fleshgrinder: We could get rid of the entire port range machinery and instead just search the entire port space, yes. I had a look at the whole range logic, but decided to not touch too much. But, we can do that. Less code is always better. 😊

Fleshgrinder: Thinking of it, this actually is not going to work. From what I understand the tests are executed in parallel with different processes. This means that one process might find a free port, hand it out, and another process finds the same port to be free since it was not bound to in the other process just yet. Now we have two components that want to bind to the same port. I think the logic as it is is necessary.

Fleshgrinder:
Nothing seems to work on both Linux and Mac. I guess running things in a container or VM is the way to go here.

PR #522, opened 2023-01-19
Fix `make copyright`

Fleshgrinder:
The `make copyright` goal always lists all Python files that have a copyright notice, and since we have more than one file with a copyright notice it always exits with success. This does not seem right, given that there is a GHA job that is supposed to check that all files (except `__init__.py`) have a copyright notice. This changes the code used in the goal to a single git invocation that does exactly that. The `!` in front inverts the exit code. Hence, if a Python file other than `__init__.py` is found without a copyright notice it exists with a non-zero exit code and lists the paths that miss the copyright notice relative to the repository root.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ clean:
 karapace/version.py: version.py
 	$(PYTHON) $^ $@
 
+# Lists the Python files that do not have an Aiven copyright. Exits with a
+# non-zero exit code if any are found.
 .PHONY: copyright
 copyright:
-	grep -EL "Copyright \(c\) 20.* Aiven" $(shell git ls-files "*.py" | grep -v __init__.py)
+	! git grep --untracked -ELm1 'Copyright \(c\) 20[0-9]{2} Aiven' -- '*.py' ':!*__init__.py'
--- a/karapace/anonymize_schemas/anonymize_avro.py
+++ b/karapace/anonymize_schemas/anonymize_avro.py
@@ -1,7 +1,7 @@
 """
 karapace - anonymize avro
 
-Copyright (c) 2022 Aiven Ltd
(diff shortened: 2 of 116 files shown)


Fleshgrinder:
Updated all Python files with the copyright notice, also updated the year since I had to touch the all anyway. Is there a reason why we have the copyright notice everywhere? It is not required, the `LICENSE` file in the root is sufficient. Removing the existing license headers would be simpler than this patch. 😉

Makefile
```diff
@@ -16,6 +16,8 @@ clean:
 karapace/version.py: version.py
 	$(PYTHON) $^ $@
 
+# Lists the Python files that do not have an Aiven copyright. Exits with a
+# non-zero exit code if any are found.
 .PHONY: copyright
 copyright:
-	grep -EL "Copyright \(c\) 20.* Aiven" $(shell git ls-files "*.py" | grep -v __init__.py)
+	! git grep --untracked -ELm1 'Copyright \(c\) 20[0-9]{2} Aiven' -- '*.py' ':!*__init__.py'
```
Fleshgrinder, line 23: This is the only code change in this patch. Notice that the `--untracked` is important for the goal to make sense locally.

jlprat:
While no strictly needed, some best practices still recommend to add the copyright notices on the files. I guess this makes more sense in projects whose files can be easily copied over to other projects. This way its easier for the ones copying to apply the proper attribution.

Makefile
RyanSkraba: This LGTM: `git grep` finding a `*.py` file (excluding `__init__.py`) without the regex causes the target to fail.

RyanSkraba (approved):

RyanSkraba:
Thanks for the contribution!

PR #523, opened 2023-01-25
fix: reset watch of authfile after delete event

giuseppelillo:
`awatch` disables file watching if a delete event for that file is issued. This may result in the impossibility to read new users and permissions data if the authfile is replaced.

# Why this way
Resetting file watch after delete event allows notifications related to the new file to be received.

### DIFF
--- a/karapace/auth.py
+++ b/karapace/auth.py
@@ -3,6 +3,7 @@ Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
 from base64 import b64encode
+from collections import defaultdict
 from dataclasses import dataclass, field
 from enum import Enum, unique
 from hmac import compare_digest
@@ -10,7 +11,7 @@ from karapace.config import InvalidConfiguration
 from karapace.rapu import JSON_CONTENT_TYPE
 from karapace.statsd import StatsClient
 from typing import Optional
-from watchfiles import awatch
+from watchfiles import awatch, Change
 
 import aiohttp
 import aiohttp.web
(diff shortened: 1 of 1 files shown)


karapace/auth.py
```diff
@@ -87,24 +88,31 @@ async def start_refresh_task(self, stats: StatsClient) -> None:
...
-                return
-            except Exception as ex:  # pylint: disable=broad-except
-                log.exception("Schema registry auth file could not be loaded")
-                stats.unexpected_exception(ex=ex, where="schema_registry_authfile_reloader")
+            while True:
+                try:
+                    async for changes in awatch(self._auth_filename, stop_event=self._refresh_auth_awatch_stop_event):
+                        changes_by_filename = defaultdict(set)
+                        for change, filename in changes:
+                            changes_by_filename[filename].add(change)
```
aiven-anton, line 96: It looks a bit odd to me to be dealing with multiple filenames here, when we're only ever interested in a single one.

Perhaps instead we could just flatten the data into a set for only the interesting file?

```python
changes = {
    change
    for  change, filename in changes
    if self._auth_filename in filename
}
if not changes:
    continue
```

I think we'd then be able to skip the for loop and conditional guarding `self._load_authfile()` too?

This would also allow simplifying the `raise StopIteration` into just `break`, I think?

jjaakola-aiven (approved):
LGTM

karapace/auth.py
```diff
@@ -87,24 +88,31 @@ async def start_refresh_task(self, stats: StatsClient) -> None:
 
         async def _refresh_authfile() -> None:
             """Reload authfile, but keep old auth data if loading fails"""
-            try:
-                # Poll delay for fallback mechanism if system notifications cannot be used.
-                async for changes in awatch(
-                    self._auth_filename, poll_delay_ms=5000, stop_event=self._refresh_auth_awatch_stop_event
```
aiven-anton, line 93: Do we know for sure that `poll_delay_ms` has no effect without `force_polling`? That seemed ambiguous in docs.

giuseppelillo: From the Rust code, it seems like it's not used if `force_polling` is not True

aiven-anton: ✅👍

PR #526, opened 2023-01-27
chore: Simplify pre-commit config

aiven-anton:

Change execution order of pre-commit hooks. It's a good idea to allow isort to run before black, to let black be authoritative on formatting. The very fast builtin pre-commit checks are moved to top to always execute first.

Some superfluous configuration options are removed, specifically there's no need to point out that Python tools should run on Python files, this is already configured in the hooks themselves.

By default, flake8 uses the .flake8 config file.

Black line-length config is moved from a hard-coded CLI argument to pyproject.toml which is picked up by default, and target-version=37 is added there explicitly.

#### Misc.:

- 🎉 Advertise support for Python 3.11 with a setup.py classifier.
- 🕵️ Fix a typo in README.

# Why this way

Simplify and improve project config.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,36 +1,6 @@
 ---
 minimum_pre_commit_version: 2.9.2
 repos:
-  - repo: https://github.com/psf/black
-    rev: 22.3.0
-    hooks:
-      - id: black
-        files: \.py$
-        args:
-          - --line-length=125
-
-  - repo: https://github.com/PyCQA/flake8
-    rev: 4.0.1
-    hooks:
-      - id: flake8
-        files: \.py$
-        args:
-          - --config=.flake8
-
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -1,28 +1,30 @@
 ---
 minimum_pre_commit_version: 2.9.2
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: trailing-whitespace
+        exclude: ^vendor/|^tests/.*/fixtures/.*|^tests/integration/test_data/.*
```
aiven-anton, line 8: Note for follow-up PR: `^tests/.*/fixtures/.*` and `^vendor/` doesn't seem to exist, probably something that's been removed.

I think it'd make sense to put `^tests/integration/test_data/.*` in a global exclude pattern, as we probably don't want any linting at all to happen there.

giuseppelillo (approved):

PR #531, opened 2023-01-30
fix: Bump pre-commit hooks to fix isort issue

aiven-anton:

Apply `pre-commit autoupdate` and bump CI Python to 3.11, to get isort 5.12.0.

Fix flake8 config.

### Why this way

The fix is opinionated in that it stops running pylint with project dependencies. I considered suggesting to remove usage of pylint entirely, and I still want to suggest that, but that's a discussion we can have separately from fixing CI builds on main.

CI is broken on main due to this isort issue, the root cause lies in an
underlying incompatible change in Poetry. The new version of isort has
dropped support for Python 3.7 so we bump CI to run on 3.11 instead (3.7
is EOL later this year anyway).

https://github.com/PyCQA/isort/issues/2083

With the upgrade to flake8 6, we ran into an issue with comments in the
configuration file. Those comments are moved to their own lines above
the config instead, see issue below.

https://github.com/PyCQA/flake8/issues/1756

Because of installation issues in the lint step, installation is altered
to only installed pre-commit which handles its own dependencies anyway.
Pylint is normalized to be installed just like the other pre-commit
hooks, the dependency on the environment is removed.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,7 +7,6 @@ enable=
   useless-suppression,
 
 disable=
-  bad-continuation,
   duplicate-code,
   fixme,
   import-outside-toplevel,
@@ -22,8 +21,15 @@ disable=
   too-many-nested-blocks,
   too-many-public-methods,
   too-many-statements,
-  too-public-methods,
   wrong-import-order,
+  import-error,
+  consider-using-f-string,
+  use-implicit-booleaness-not-comparison,
+  unspecified-encoding,
+  no-name-in-module,
+  use-list-literal,
+  use-dict-literal,
+
 
 [FORMAT]
 max-line-length=125
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -18,7 +18,7 @@ import aiohttp
 import aiohttp.web
 import aiohttp.web_exceptions
 import asyncio
-import cgi
+import cgi  # pylint: disable=deprecated-module
 import hashlib
 import json
 import logging
@@ -192,7 +192,7 @@ class RestApp:
             "Server": SERVER_NAME,
         }
 
-    def check_rest_headers(self, request: HTTPRequest) -> dict:  # pylint:disable=inconsistent-return-statements
+    def check_rest_headers(self, request: HTTPRequest) -> dict:
         method = request.method
         default_content = "application/vnd.kafka.json.v2+json"
         default_accept = "*/*"
--- a/karapace/serialization.py
+++ b/karapace/serialization.py
@@ -166,7 +166,7 @@ class SchemaRegistrySerializer:
         assert self.registry_client, "must not call this method after the object is closed."
         schema_id, schema = await self.registry_client.get_latest_schema(subject)
         async with self.state_lock:
-            schema_ser = schema.__str__()
+            schema_ser = str(schema)
             self.schemas_to_ids[schema_ser] = schema_id
             self.ids_to_schemas[schema_id] = schema
         return schema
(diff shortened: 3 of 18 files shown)


.pylintrc
```diff
@@ -22,8 +21,17 @@ disable=
...
   wrong-import-order,
+  import-error,
+  consider-using-f-string,
+  use-implicit-booleaness-not-comparison,
+  unspecified-encoding,
+  no-name-in-module,
+  unnecessary-dunder-call,
+  use-list-literal,
+  use-dict-literal,
+  deprecated-module,
```
aiven-anton, line 33: `import-error`: pylint is no longer installed with all project dependencies, this is covered by mypy + runtime tests.
`consider-using-f-string`, `use-list-literal`, `use-dict-literal`: Covered by [pyupgrade](https://github.com/aiven/karapace/pull/530).
`unspecified-encoding`: Covered by pyupgrade, but with different resolution.
`use-implicit-booleaness-not-comparison`: I think this is overly opinionated. `foo == []` is fine and doesn't need to be outlawed.

.pylintrc
```diff
@@ -22,8 +21,15 @@ disable=
...
   too-many-statements,
-  too-public-methods,
   wrong-import-order,
+  import-error,
+  consider-using-f-string,
+  use-implicit-booleaness-not-comparison,
+  unspecified-encoding,
+  no-name-in-module,
+  use-list-literal,
+  use-dict-literal,
```
aiven-anton, line 31: `import-error`, `no-name-in-module`: pylint is no longer installed with all project dependencies, checks like this is better left to mypy + runtime tests we already have in place.
`consider-using-f-string`, `use-list-literal`, `use-dict-literal`: Covered by [pyupgrade](https://github.com/aiven/karapace/pull/530).
`unspecified-encoding`: Covered by pyupgrade, but with different resolution.
`use-implicit-booleaness-not-comparison`: I think this is overly opinionated. `foo == []` is fine and doesn't need to be outlawed.

karapace/rapu.py
```diff
@@ -18,7 +18,7 @@
 import aiohttp.web
 import aiohttp.web_exceptions
 import asyncio
-import cgi
+import cgi  # pylint: disable=deprecated-module
```
aiven-anton, line 21: I created a separate issue for this: https://github.com/aiven/karapace/issues/533

karapace/serialization.py
```diff
@@ -166,7 +166,7 @@ async def get_schema_for_subject(self, subject: str) -> TypedSchema:
         assert self.registry_client, "must not call this method after the object is closed."
         schema_id, schema = await self.registry_client.get_latest_schema(subject)
         async with self.state_lock:
-            schema_ser = schema.__str__()
+            schema_ser = schema.__str__()  # pylint: disable=unnecessary-dunder-call
```
jjaakola-aiven, line 169: `str(schema)`

jjaakola-aiven (changes requested):

tests/integration/utils/kafka_server.py
```diff
@@ -192,5 +192,5 @@ def configure_and_start_kafka(
         ),
     )
     env: Dict[bytes, bytes] = {}
-    proc = Popen(kafka_cmd, env=env)
+    proc = Popen(kafka_cmd, env=env)  # pylint: disable=consider-using-with
```
aiven-anton, line 195: I think we should really handle these better, it's a good find by pylint, but it requires some refactoring.

Now when looking closer I noticed they only have one call site each, so perhaps it makes sense to refactor now?

karapace/serialization.py
aiven-anton: I addressed all of these, plus the one suggesting `enumerate()`. I'm a bit hesitant to addressing the two `consider-using-with` in this same PR, but if you think that makes sense I can do so.

tests/integration/utils/kafka_server.py
aiven-anton: I pushed a separate PR for this: https://github.com/aiven/karapace/pull/534

jjaakola-aiven (approved):

tvainika (approved):

PR #534, opened 2023-01-31
fix: Use contextmanager for process creation in tests

aiven-anton:

Use `@contextmanager` for handling ZooKeeper and Kafka test processes.

### Why this way

This makes sure that the responsibility for stopping running processes does not leak to call sites, but is rather handled in the same place as the processes are spawned. This was found [when upgrading pylint to latest version](https://github.com/aiven/karapace/pull/531#discussion_r1091618748), which introduces a lint error for these cases.


aiven-anton:
There have been a few consecutive runs with timeouts now. I wonder if cleaning up the processes is the cause of the increased execution time. This needs more investigation, so changed to draft.

PR #536, opened 2023-02-09
Add Backup File Overwrite Protection

Fleshgrinder:
The schema backup tool is currently overwriting the output file without asking to do so. It is also writing directly to the output file, and leaves the file intact even if the backup actually failed. This patch adds guards that fail if the output file already exists, but also an `--overwrite` flag. It also introduces a temporary file that is persisted only if the backup actually succeeded, not leaving behind any trace of the work if not.

The output needs to be a file to be overwritten, even with the `--overwrite` flag present. Any other path type is not going to be touched. Our code verifies this, and Python's `replace` function does as well.

Other smaller things worth mentioning:

- The `--location` accepts `-` as synonym for `stdout` as is conventional, the current empty string continues to work as well and stays the default. But, `-` is not going to create a file called `-` anymore.
- Renamed `Writer` to `_writer` and by doing so made it private. This could be considered a breaking change since the name of the symbol changed, and anyone relying on it fails now. However, the `Writer` was never meant to be a public symbol, and we do not have any policy regarding such changes.
- Fixed the return type hint from `Iterable[TextIO]` to `TextIO` since it is a context manager, and not something that can be iterated.

### DIFF
--- a/karapace/schema_backup.py
+++ b/karapace/schema_backup.py
@@ -15,7 +15,9 @@ from karapace.key_format import KeyFormatter
 from karapace.schema_reader import new_schema_topic_from_config
 from karapace.typing import JsonData
 from karapace.utils import json_encode, KarapaceKafkaClient, Timeout
-from typing import IO, Iterable, Optional, TextIO, Tuple, Union
+from pathlib import Path
+from tempfile import mkstemp
+from typing import IO, Optional, TextIO, Tuple, Union
 
 import argparse
 import base64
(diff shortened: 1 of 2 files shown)


karapace/schema_backup.py
```diff
@@ -44,11 +46,56 @@ class BackupError(Exception):
...
+            if dst.exists():
+                if overwrite is not True:
+                    raise FileExistsError(f"--location already exists at {dst}, use --overwrite to replace the file.")
+                if not dst.is_file():
+                    raise FileExistsError(
+                        f"--location already exists at {dst}, but is not a file and thus cannot be overwritten."
+                    )
+
+        check_dst()
+        fd, path = mkstemp(dir=dst.parent, prefix=dst.name)
```
jjaakola-aiven, line 81: This throws `FileNotFoundError` if parent directory does not exist. For completeness add to documentation.

karapace/schema_backup.py
```diff
@@ -44,11 +46,56 @@ class BackupError(Exception):
...
+        src = Path(path)
+        try:
+            fp = open(fd, "w", encoding="utf8")
+            try:
+                yield fp
+                os.fsync(fd)
```
jjaakola-aiven, line 87: If the file is buffered, flush should be called before fsync. I am not sure if the mkstemp creates buffered file?

karapace/schema_backup.py
Fleshgrinder: Not a new problem, but we could add a flag to create missing directories.

Fleshgrinder: Let us simply create them, I see no harm in that.

jjaakola-aiven (approved):

PR #537, opened 2023-02-10
Add Backup Partition Count Check

Fleshgrinder:
This adds a check that verifies that the source and target topic for a backup indeed have a single partition, instead of simply assuming that it does. This is not a breaking change, since it makes the implicit explicit, and users who are going to receive the error are going to be protected from a lot of hardship.

This also removes the consumer and producer clients from the class by moving them into dedicated top-level functions. Those top-level functions are context managers that close the clients automatically in a reliable way. This removal could be considered a breaking change, since we are removing public functions from a class. However, these functions were never meant to be public, especially not because they were directly meddling with internal state.

The new top-level functions **always** check the partition count, and client creation without a partition count check is not possible. This is on purpose, to ensure safety. Last, but not least, there are tests that verify all of the claims.

### DIFF
--- a/tests/unit/test_schema_backup.py
+++ b/tests/unit/test_schema_backup.py
@@ -2,11 +2,67 @@
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
-from karapace.schema_backup import _writer
+from kafka import KafkaConsumer, KafkaProducer
+from kafka.structs import PartitionMetadata
+from karapace import config
+from karapace.config import Config
+from karapace.schema_backup import _consumer, _producer, _writer, PartitionCountError
 from pathlib import Path
+from types import FunctionType
+from typing import AbstractSet, Callable, Type, Union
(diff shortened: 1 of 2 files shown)


tests/unit/test_schema_backup.py
```diff
@@ -2,13 +2,69 @@
...
 import sys
 
 
+class TestClients:
+    @staticmethod
+    def _partition_metadata(c: int = 1) -> AbstractSet[PartitionMetadata]:
+        return {PartitionMetadata("topic", i, 0, tuple(), tuple(), None) for i in range(0, c)}
+
+    @pytest.mark.parametrize(
+        "f,t,m",
```
jjaakola-aiven, line 25: I'd like to have longer names for variables and parameters, makes quite a bit easier to read. Applies to other places too.

jjaakola-aiven (approved):
LGTM!

PR #538, opened 2023-02-10
Add more graceful handling when deserialization fails for a record during consuming

roope-kar:
Currently, failing to deserialize a record to the requested format during consuming results in internal error. Also, json decode error was not caught.

For example: If user has produced binary records to a topic and then tried to consume them using format json. The response would be a 500 internal server error without proper message.

Should now properly catch json decode exception and change exception type from 500 to 422 including the intended message. Also adds test that should help ensure this will happen with all known formats.

### DIFF
--- a/karapace/kafka_rest_apis/consumer_manager.py
+++ b/karapace/kafka_rest_apis/consumer_manager.py
@@ -12,8 +12,8 @@ from kafka.structs import TopicPartition
 from karapace.config import Config, create_client_ssl_context
 from karapace.kafka_rest_apis.error_codes import RESTErrorCodes
 from karapace.karapace import empty_response, KarapaceBase
-from karapace.serialization import InvalidMessageHeader, InvalidPayload, SchemaRegistrySerializer
-from karapace.utils import convert_to_int, json_decode
+from karapace.serialization import DeserializationError, InvalidMessageHeader, InvalidPayload, SchemaRegistrySerializer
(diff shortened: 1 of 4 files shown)


karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -509,16 +509,30 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
...
                     try:
                         key = await self.deserialize(msg.key, request_format) if msg.key else None
-                    except (UnpackError, InvalidMessageHeader, InvalidPayload) as e:
-                        KarapaceBase.internal_error(
+                    except (
+                        UnpackError,
+                        InvalidMessageHeader,
+                        InvalidPayload,
+                        json.decoder.JSONDecodeError,
+                        UnicodeDecodeError,
```
jjaakola-aiven, line 517: Handle these in the serde function and raise e.g. `DeserializationError` from there.

jjaakola-aiven (changes requested):

karapace/kafka_rest_apis/consumer_manager.py
roope-kar: 👍🏼

roope-kar:
@jjaakola-aiven Thanks for the review, hopefully understood the change request correctly!

jjaakola-aiven (approved):

PR #539, opened 2023-02-13
feat: use UltraJSON if available

jjaakola-aiven:

UltraJson gives slight performance improvement.

# Why this way

Decided to support the vanilla `json` and `ultrajson` for now.

### DIFF
--- a/karapace/schema_models.py
+++ b/karapace/schema_models.py
@@ -19,11 +19,10 @@ from karapace.protobuf.exception import (
 )
 from karapace.protobuf.schema import ProtobufSchema
 from karapace.typing import ResolvedVersion, SchemaId, Subject
-from karapace.utils import json_encode
+from karapace.utils import json_decode, json_encode, JSONDecodeError
 from typing import Any, cast, Dict, NoReturn, Optional, Union
 
 import hashlib
-import json
 import logging
 
 LOG = logging.getLogger(__name__)
@@ -37,15 +36,8 @@ def parse_avro_schema_definition(s: str, validate_enum_symbols: bool = True, val
     parsing fails because of it, the extra data can be removed and parsed
     again.
     """
-    try:
-        json_data = json.loads(s)
-    except json.JSONDecodeError as e:
-        if e.msg != "Extra data":
-            raise
-
-        json_data = json.loads(s[: e.pos])
-
-    return avro_parse(json.dumps(json_data), validate_enum_symbols=validate_enum_symbols, validate_names=validate_names)
+    json_data = json_decode(s)
+    return avro_parse(json_encode(json_data), validate_enum_symbols=validate_enum_symbols, validate_names=validate_names)
 
 
 def parse_jsonschema_definition(schema_definition: str) -> Draft7Validator:
@@ -54,7 +46,7 @@ def parse_jsonschema_definition(schema_definition: str) -> Draft7Validator:
     Raises:
         SchemaError: If `schema_definition` is not a valid Draft7 schema.
     """
-    schema = json.loads(schema_definition)
+    schema = json_decode(schema_definition)
     Draft7Validator.check_schema(schema)
     return Draft7Validator(schema)
 
@@ -98,7 +90,7 @@ class TypedSchema:
     def to_dict(self) -> Dict[str, Any]:
         if self.schema_type is SchemaType.PROTOBUF:
             raise InvalidSchema("Protobuf do not support to_dict serialization")
-        return json.loads(self.schema_str)
+        return json_decode(self.schema_str)
 
     def fingerprint(self) -> str:
         if self._fingerprint_cached is None:
@@ -109,8 +101,8 @@ class TypedSchema:
     def normalize_schema_str(schema_str: str, schema_type: SchemaType) -> str:
         if schema_type is SchemaType.AVRO or schema_type is SchemaType.JSONSCHEMA:
             try:
-                schema_str = json_encode(json.loads(schema_str), sort_keys=True)
-            except json.JSONDecodeError as e:
+                schema_str = json_encode(json_decode(schema_str), compact=True, sort_keys=True)
+            except JSONDecodeError as e:
                 LOG.error("Schema is not valid JSON")
                 raise e
         elif schema_type == SchemaType.PROTOBUF:
@@ -150,14 +142,14 @@ def parse(
                 validate_enum_symbols=validate_avro_enum_symbols,
                 validate_names=validate_avro_names,
             )
-        except (SchemaParseException, json.JSONDecodeError, TypeError) as e:
+        except (SchemaParseException, JSONDecodeError, TypeError) as e:
             raise InvalidSchema from e
 
     elif schema_type is SchemaType.JSONSCHEMA:
         try:
             parsed_schema = parse_jsonschema_definition(schema_str)
             # TypeError - Raised when the user forgets to encode the schema as a string.
-        except (TypeError, json.JSONDecodeError, SchemaError, AssertionError) as e:
+        except (TypeError, JSONDecodeError, SchemaError, AssertionError) as e:
             raise InvalidSchema from e
 
     elif schema_type is SchemaType.PROTOBUF:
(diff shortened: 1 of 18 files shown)


tests/unit/test_avro_compatibility.py
```diff
@@ -33,8 +33,9 @@
 )
 schema8 = parse_avro_schema_definition(
     '{"type":"record","name":"myrecord","fields":[{"type":"string","name":"f1"},{"type":"string",'
-    '"name":"f2","default":"foo"}]},{"type":"string","name":"f3","default":"bar"}]}'
```
jjaakola-aiven, line 36: This data was broken so I took the trailing data out.

tests/integration/test_schema.py
```diff
@@ -2882,7 +2884,7 @@ async def test_invalid_schema_should_provide_good_error_messages(registry_async_
         f"subjects/{test_subject}/versions",
         json={"schema": schema_str[:-1]},
     )
-    assert res.json()["message"] == "Invalid AVRO schema. Error: Expecting ',' delimiter: line 1 column 18 (char 17)"
+    assert res.json()["message"].startswith("Invalid AVRO schema. Error: ")
```
jjaakola-aiven, line 2887: The actual error message is different in `ujson` than in `json`.

karapace/schema_models.py
```diff
@@ -37,15 +36,8 @@ def parse_avro_schema_definition(s: str, validate_enum_symbols: bool = True, val
...
     again.
     """
-    try:
-        json_data = json.loads(s)
-    except json.JSONDecodeError as e:
-        if e.msg != "Extra data":
-            raise
-
-        json_data = json.loads(s[: e.pos])
-
```
jjaakola-aiven, line 47: I removed this whole block as this accepts invalid JSON. If there is trailing data the data shall be considered bad.
For example `{"type": "fish"},["trailing"]` was splitted with the position from `json.JSONDecodeError` to `{"type": "fish"}`. The result is valid JSON, but should not be accepted as the input is garbage.

Fleshgrinder:
Why not [orjson](https://pypi.org/project/orjson/)? It is significantly faster than all other options, handles more types natively, and codes directly to `bytes`. Thus removing the unnecessary `str` allocation and `str` to `bytes` conversion for sending things over the wire. It also still supports Python 3.7.

jjaakola-aiven:
The `orjson` can be added later if needed. `ujson` is quite straightforward replacement for vanilla `json`.

karapace/utils.py
```diff
@@ -12,14 +12,25 @@
...
+from typing import IO, NoReturn, Optional, overload, Union
 
-import json
+import importlib
 import kafka.client_async
 import logging
 import time
 
+ultrajson = importlib.util.find_spec("ujson")
+if ultrajson:
```
Fleshgrinder, line 25: There is no need for the global variable, is there?

```suggestion
if importlib.util.find_spec("ujson"):
```

karapace/utils.py
```diff
@@ -73,17 +84,34 @@ def default_json_serialization(
...
+    if indent is not None:
+        arguments["indent"] = indent
     res = json.dumps(
         obj,
-        sort_keys=sort_keys,
-        default=default_json_serialization,
-        separators=separators,
+        **arguments,
     )
     return res.encode("utf-8") if binary else res
```
Fleshgrinder, line 100: Using overloads allows us to help editors in understanding what this function is going to return, as it either returns `bytes` or `str` in a predictable way. I also removed the global `SEPARATORS` constant, as there is no need to occupy the memory for it, as there is no reuse. Using `Optional[T] = None` for all arguments makes it possible to write wrapper functions, without copying the actual default values. JSON should also be compact by default, no need to waste precious bytes.

```python
@overload
def json_encode(
    it: Any,
    *,
    binary: Literal[False, None] = None,
    sort_keys: Optional[bool] = None,
    compact: Optional[bool] = None,
    indent: Optional[int] = None,
) -> str:
    ...


@overload
def json_encode(
    it: Any,
    *,
    binary: Literal[True],
    sort_keys: Optional[bool] = None,
    compact: Optional[bool] = None,
    indent: Optional[int] = None,
) -> bytes:
    ...


def json_encode(
    it: Any,
    *,
    binary: Optional[bool] = None,
    sort_keys: Optional[bool] = None,
    compact: Optional[bool] = None,
    indent: Optional[int] = None,
) -> Union[bytes, str]:
    kwargs = {}
    if indent is not None:
        kwargs["indent"] = indent
    if compact is not False and indent is None:  # compact by default
        kwargs["separators"] = (",", ":")
    if sort_keys is True:
        kwargs["sort_keys"] = True
    result = json.dumps(it, default=default_json_serialization, **kwargs)
    return result if binary is True else result.encode("utf8")
```

```python
string = json_encode({})  # correctly recognized as `str`
also_string = json_encode({}, binary=False)  # correctly recognized as `str`
binary = json_encode({}, binary=True)  # correctly recognized as `bytes`
```

karapace/utils.py
```diff
@@ -73,17 +84,34 @@ def default_json_serialization(
...
+def json_decode(content: [str, IO]) -> JsonData:
+    if isinstance(content, str):
+        try:
+            return json.loads(content)
+        except JSONDecodeError as ex:
+            raise ex
+    try:
+        return json.load(content)
+    except JSONDecodeError as ex:
+        raise ex
```
Fleshgrinder, line 112: There is no need to catch and rethrow the same exception, and `bytes` is a perfectly valid input for both `json` and `usjon` as well. By including `bytes` we can also get rid of **ALL** `.decode("utf8")` that are currently sprinkled across the codebase when calling this function.

```suggestion
def json_decode(it: Union[bytes, str, IO]) -> JsonData:
    if isinstance(content, bytes) or isinstance(content, str):
        return json.loads(content)
    else:
        return json.load(content)
```

karapace/schema_reader.py
```diff
@@ -326,8 +325,8 @@ def handle_messages(self) -> None:
                 value = None
                 if msg.value:
                     try:
-                        value = json.loads(msg.value.decode("utf8"))
-                    except json.JSONDecodeError:
+                        value = json_decode(msg.value.decode("utf8"))
```
Fleshgrinder, line 328: Just one of many places of unnecessary binary decoding, if the signature of `json_decode` is updated as suggested in my other comment.

```suggestion
                        value = json_decode(msg.value)
```

Fleshgrinder (changes requested):

jjaakola-aiven:
As this also impacts the formatting of the key I registered two schemas. Another with code from current main and another with this branch. Result of keys:
```
THIS BRANCH -> {"keytype":"SCHEMA","subject":"test-key","version":1,"magic":1}	
MAIN BRANCH -> {"keytype":"SCHEMA","subject":"test-diff","version":1,"magic":1}
```

Both have the same order of fields and are in the most compact form.

karapace/utils.py
jjaakola-aiven: Done.

karapace/utils.py
jjaakola-aiven: Done.

karapace/utils.py
jjaakola-aiven: Yes, rethrow was a workaround to flake8 and pylint unused import nag.

karapace/schema_reader.py
jjaakola-aiven: Removed the unwanted decoding calls.

karapace/utils.py
```diff
@@ -70,18 +80,53 @@ def default_json_serialization(
...
+    if compact is not False and indent is None:
+        kwargs["separators"] = (",", ":")
+    if sort_keys is True:
+        kwargs["sort_keys"] = True
+    result = json.dumps(obj, default=default_json_serialization, **kwargs)
+    return result.encode("utf8") if binary is True else result
+
+
+def json_decode(content: [AnyStr, IO[AnyStr]]) -> JsonData:
+    if isinstance(content, str) or isinstance(content, bytes):  # pylint: disable=consider-merging-isinstance
```
Fleshgrinder, line 127: Why would we not follow pylint's suggestion? It is more efficient to perform the work in C.

```suggestion
    if isinstance(content, (str, bytes)):
```

Fleshgrinder (changes requested):

karapace/utils.py
Fleshgrinder: Not possible in Python 3.7

Fleshgrinder (approved):

PR #542, opened 2023-02-14
Fix Create Backup Consumption Logic

Fleshgrinder:
The consumption logic is currently counting how many records it received in a single batch returned from poll, and when it is empty it concludes that the backup is successfully finished. However, there are meany reasons why a batch returned by poll is empty, especially with timeouts applied to it. A consequence of this is that a backup created at $t_1$ may contain more records than a backup created at $t_2$ (without any external changes to the topic content, e.g. compaction).

To fix this we have to use offset watermarks. With them we can determine if we are done, or not. The patch now exposes the poll timeout, so that users can increase it in case they encounter issues, and it uses a longer default poll timeout to ensure that users are not going to see errors right away (increased from 1 second to 1 minute).

----

~Requires (and includes) https://github.com/aiven/karapace/pull/540 to be merged first.~

### DIFF
--- a/karapace/backup/consumer.py
+++ b/karapace/backup/consumer.py
@@ -1,0 +1,45 @@
+"""
+Copyright (c) 2023 Aiven Ltd
+See LICENSE for details
+"""
+from __future__ import annotations
+
+from datetime import timedelta
+from isodate import duration_isoformat, parse_duration
+
+__all__ = ["PollTimeout"]
+
+
+class PollTimeout:
+    """Specifies how long a single poll attempt may take while consuming the topic.
+
+    It may be necessary to adjust this value in case the cluster is slow. The value must be given in ISO8601 duration
+    format (e.g. `PT1.5S` for 1,500 milliseconds) and must be at least on second. Defaults to one minute.
+    """
+
+    __slots__ = ("__value",)
+
+    def __init__(self, value: str | timedelta) -> None:
+        self.__value = value if isinstance(value, timedelta) else parse_duration(value)
+        if self.__value // timedelta(seconds=1) < 1:
+            raise ValueError(f"Poll timeout MUST be at least one second, got: {self}")
+
+    @classmethod
+    def default(cls) -> PollTimeout:
+        return cls(timedelta(minutes=1))
+
+    @classmethod
+    def of(cls, minutes: int = 0, seconds: int = 0, milliseconds: int = 0) -> PollTimeout:
+        """Convenience function to avoid importing ``timedelta``."""
+        return PollTimeout(timedelta(minutes=minutes, seconds=seconds, milliseconds=milliseconds))
+
+    def __str__(self) -> str:
+        """Returns the ISO8601 formatted value of this poll timeout."""
+        return duration_isoformat(self.__value)
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(value='{self}')"
+
+    def to_milliseconds(self) -> int:
+        """Returns this poll timeout in milliseconds, anything smaller than a milliseconds is ignored (no rounding)."""
+        return self.__value // timedelta(milliseconds=1)
--- a/karapace/schema_backup.py
+++ b/karapace/schema_backup.py
@@ -7,10 +7,13 @@ See LICENSE for details
 from enum import Enum
 from kafka import KafkaConsumer, KafkaProducer
 from kafka.admin import KafkaAdminClient
+from kafka.consumer.fetcher import ConsumerRecord
 from kafka.errors import TopicAlreadyExistsError
-from kafka.structs import PartitionMetadata
+from kafka.structs import PartitionMetadata, TopicPartition
 from karapace import constants
 from karapace.anonymize_schemas import anonymize_avro
+from karapace.backup.consumer import PollTimeout
+from karapace.backup.errors import BackupError, PartitionCountError, StaleConsumerError
 from karapace.config import Config, read_config
 from karapace.key_format import KeyFormatter
 from karapace.schema_reader import new_schema_topic_from_config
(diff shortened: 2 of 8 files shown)


karapace/schema_backup.py
```diff
@@ -349,25 +349,72 @@ def _restore_backup_version_2(self, producer: KafkaProducer, fp: IO) -> None:
...
+            has not been reached yet.
+        """
+        poll_timeout_ms = 1_000 if poll_timeout is None else poll_timeout // timedelta(milliseconds=1)
+        topic = self.topic_name
         with _writer(self.backup_location, overwrite=overwrite) as fp:
-            with _consumer(self.config, self.topic_name) as consumer:
-                LOG.info("Starting schema backup read for topic: %r", self.topic_name)
-
-                topic_fully_consumed = False
+            with _consumer(self.config, topic) as consumer:
```
jjaakola-aiven, line 380:
```suggestion
            with _writer(self.backup_location, overwrite=overwrite) as fp, _consumer(self.config, topic) as consumer:
```
Combining both context managers to single line would remove one indent level.

karapace/schema_backup.py
```diff
@@ -349,25 +349,72 @@ def _restore_backup_version_2(self, producer: KafkaProducer, fp: IO) -> None:
...
-
-                    for _, messages in raw_msg.items():
-                        for message in messages:
-                            ser = export_func(key_bytes=message.key, value_bytes=message.value)
-                            if ser:
-                                fp.write(ser)
-
-                LOG.info("Schema export written to %r", "stdout" if fp is sys.stdout else self.backup_location)
+                if end_offset != 0:  # non-empty topic
+                    end_offset -= 1  # high watermark to actual end offset
```
jjaakola-aiven, line 390: I would check if start offset and end offset are equal, this would be the rare case of a topic that is empty because of retention.

karapace/schema_backup.py
Fleshgrinder: I did that with `()` surrounding them so that I can put them on separate lines for clarity and it complained that this feature is not available in 3.7, didn't know that it works without the `()`.

karapace/schema_backup.py
```diff
@@ -349,25 +344,72 @@ def _restore_backup_version_2(self, producer: KafkaProducer, fp: IO) -> None:
...
+                    records: Collection[ConsumerRecord] = consumer.poll(poll_timeout_ms).get(topic_partition, [])
+                    if len(records) == 0:
+                        raise StaleConsumerError(topic_partition, start_offset, end_offset, last_offset, poll_timeout)
+                    record: ConsumerRecord
+                    for record in records:
+                        fp.write(serialize(record.key, record.value))
+                        record_count += 1
+                    last_offset = record.offset
+                    if last_offset >= end_offset:
+                        break
```
jjaakola-aiven, line 404: I'd compare the record offset to end offset in the loop and break immediately when reached. Now this could backup more than was decided as the end offset.

Fleshgrinder: Which is intentional, backing up more than the minimum is totally fine. Using a `while` loop would mean that we perform the same check twice in a row, however, using the `do`-`while` loop means that we perform only as many checks as necessary, while allowing to back up as much as possible, but at least what was in the topic when we started.

karapace/backup/consumer.py
```diff
@@ -0,0 +1,45 @@
+"""
+Copyright (c) 2023 Aiven Ltd
+See LICENSE for details
+"""
+from datetime import timedelta
+from isodate import duration_isoformat, parse_duration
```
antonagestam, line 6: question: [python-dateutil](https://github.com/dateutil/dateutil) is the usual go-to to fill the needs that stdlib datetime doesn't fulfill. Have you checked whether it can be used instead?

karapace/backup/consumer.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Union
+
+__all__ = ["PollTimeout"]
+
+
+# TODO @final (Python 3.8+)
```
antonagestam, line 12: question: Is it important somehow that this class isn't subclassed? Unless something breaks when subclassing, I'd consider not adding `@final` at all. (And if it is important, maybe add a comment explaining why?)

karapace/backup/consumer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    __slots__ = ("__value",)
+
+    def __init__(self, value: Union[str, timedelta]) -> None:
+        self.__value = value if isinstance(value, timedelta) else parse_duration(value)
+        if self.__value // timedelta(seconds=1) < 1:
+            raise ValueError(f"Poll timeout MUST be at least one second, got: {self}")
+
+    @classmethod
+    def default(cls) -> "PollTimeout":
```
antonagestam, line 28: nit (non-blocking): I think we should prefer `from __future__ import annotations` over string syntax type hints, so that we're more compatible with the future than with the past.

karapace/backup/consumer.py
Fleshgrinder: By not making it final we have to assume that someone using Karapace is subclassing this class, and using it. This has serious implications when it comes to making changes to public symbols. Everything should be final, to avoid such issues.

karapace/backup/consumer.py
Fleshgrinder: It does not support ISO8601 durations.

karapace/backup/consumer.py
antonagestam: I'm failing to see any serious implications at all. 

Karapace is an application, and not a library, so any usage in that way is obviously not supported and would have no compatibility guarantees whatsoever in the first place. Internals like this is not a public API.

Basically I disagree that we should assume this application is used as a library.

jjaakola-aiven (approved):

PR #545, opened 2023-02-15
Fix GHA Protoc Installation

Fleshgrinder:
The protobuf compiler that is being installed in GitHub Actions has a different version than the library that is used by Karapace. To avoid further drift it is necessary to ensure that they stay aligned automatically. This patch adds some code to extract the actual Protobuf version from the `requirements.txt` file.

Note that this approach is going to break with newer Protobuf versions. This is because Google decided to stop versioning all Protobuf dependencies alike. However, newer versions of Python Protobuf do not require the Protobuf Compiler to be installed anymore. Hence, if we upgrade, we can get rid of the Protobuf Compiler installation entirely.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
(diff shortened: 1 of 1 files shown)


jjaakola-aiven (approved):

libretto:
Part of the Protobuf functionality that we are implementing requires Protobuf Compiler to be installed.

PR #546, opened 2023-02-15
Async backup restoration with error handling

giuseppelillo:

Allow the backup restoration producer to send messages asynchronously.
# Why this way

If an error arises while producing a message, the exception is raised, causing the process to finish with an exit code of 1.

### DIFF
--- a/karapace/schema_backup.py
+++ b/karapace/schema_backup.py
@@ -7,7 +7,7 @@ See LICENSE for details
 from enum import Enum
 from kafka import KafkaConsumer, KafkaProducer
 from kafka.admin import KafkaAdminClient
-from kafka.errors import KafkaError, TopicAlreadyExistsError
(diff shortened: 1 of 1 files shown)


karapace/schema_backup.py
```diff
@@ -311,19 +313,23 @@ def restore_backup(self) -> None:
                     self._restore_backup_version_2(producer, fp)
                 else:
                     self._restore_backup_version_1_single_array(producer, fp)
+            producer.flush(timeout=self.timeout_kafka_producer)
+            if self.producer_exception:
```
Fleshgrinder, line 317: We have to be explicit about what we do, otherwise some weird Python coercion is going to bite us in the future. 😉

```suggestion
            if self.producer_exception is not None:
```

Fleshgrinder (changes requested):

Fleshgrinder (approved):

PR #547, opened 2023-02-20
Fix Python Executable Search in Tests

Fleshgrinder:
This removes the hardcoded `python` executable name in tests with a more dynamic approach that resolves to the Python executable used by the tests. Not all systems have Python 3 available at `python` by default, and there is no guarantee that the `PATH` in a `venv` is modified to include the `venv/bin` directory in the `PATH`. Python offers `sys.executable` to find the current executable, if that fails we fall back to the same environment variable we use in our make file, and if that fails we use `python3` (seems to be available on most systems).

### DIFF
--- a/tests/integration/test_karapace.py
+++ b/tests/integration/test_karapace.py
@@ -5,9 +5,9 @@ See LICENSE for details
 from contextlib import ExitStack
 from karapace.config import set_config_defaults
 from pathlib import Path
-from subprocess import Popen
 from tests.integration.utils.network import PortRangeInclusive
 from tests.integration.utils.process import stop_process
+from tests.utils import popen_karapace_all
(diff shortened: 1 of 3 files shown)


aiven-anton:
I like this, although I don't understand this "there is no guarantee that the PATH in a venv is modified to include the venv/bin directory in the PATH". The definition of activating a venv is it being present on PATH, no?

aiven-anton (approved):

Fleshgrinder:
No, you can also call venv binaries directly, and they then use the venv without activating it.

```shell
# no activation
./venv/bin/python -m pytest
```

aiven-anton:
@Fleshgrinder Ah, right 👍

PR #549, opened 2023-02-21
feat: Add Backup Format 3 Protobuf Schema

Fleshgrinder:
This adds the Protobuf schema for the version 3 backup format. Protobuf is going to enable us to statically verify backward and forward compatibility of the schema, and thus allow us to add fields without introducing a new version in the future. Protobuf uses a universal C extension for their dynamic languages, resulting in a massive speed-up. This is available starting with Protobuf 20, we thus will need to upgrade Protobuf to actually make use of this.

### DIFF
--- a/karapace/backup/v3/schema.proto
+++ b/karapace/backup/v3/schema.proto
@@ -1,0 +1,147 @@
+// The version 3 backup format of Karapace uses Protobuf serialization. This
+// data exchange format makes it possible to verify backward and forward
+// compatibility statically, and compared to a custom format enables us to add
+// new fields without introducing a new version.
+//
+// The backup file format itself is as follows:
+//
+// 1. The first byte is the “magic” marker, the NUL byte (`\x00`). This ensures
+//    that other tools reading the file know that this is a binary file (e.g.
+//    the `file` utility).[^1]
+// 2. The second byte is the backup format version, an unsigned 8-bit integer.
+//    This is always `\x03` for the backup format 3.[^1][^2]
+// 3. The next 4 bytes contain the byte count of the `Metadata` message that
+//    follows. The length is an unsigned 32-bit little-endian integer.[^3]
+// 4. The `Metadata` message.
+// 5. Zero or more records follow:
+//    1. 4 bytes containing the unsigned 32-bit little-endian integer byte count
+//       of the `Envelope` message that follows.[^3]
+//    2. The `Envelope` message.
+//
+// [^1]: The combination of these is referred to as the “magic marker version”.
+// [^2]: Endianness is irrelevant for a single byte integer.
+// [^3]: We are not using _varint_ encoding for the message framing, because it
+//   would require additional bound checks and logic to ensure that it actually
+//   is only a 32-bit value. Using fixed 4 bytes means that we can just read.
+//   Compression of the backups is used to reduce size, and we are not worried
+//   about the additional bytes required for encoding the framing bytes.
+syntax = "proto3";
+
+package karapace.backup.v3;
+
+// The metadata directly follows the magic marker version bytes. It contains
+// additional information regarding the backup.
+message Metadata {
+  // The version of the backup, this **MUST** be the same as the version that
+  // follows the magic marker byte. This is included here so that it can be
+  // passed around as part of the metadata to other parts of the application.
+  // The range of the allowed values for this is 1–255, to comply with the
+  // possible values that can be encoded in the “magic marker version” bytes.
+  // Restoration fails if the version found in the “magic marker version” and
+  // this field differ.
+  uint32 version = 1;
+
+  // The name of the tool that created the backup. The name is intended for
+  // humans, as such it **MUST** consist of printable NFC UTF-8 codepoints only.
+  // The length **MUST NOT** exceed 64 bytes.
+  //
+  // Always `Karapace` for backups created by Karapace.
+  optional string tool_name = 2;
+
+  // The version of the tool that created the backup. The version is intended
+  // for humans, as such it **MUST** consist of printable NFC UTF-8 codepoints
+  // only. The length **MUST NOT** exceed 64 bytes. The version **SHOULD** be
+  // [PEP440](https://peps.python.org/pep-0440/) or [SemVer](https://semver.org/)
+  // compatible.
+  optional string tool_version = 3;
+
+  // The UNIX time in seconds the backup creation started.
+  optional int64 create_time = 4;
+
+  // The estimated record count included in this backup. This is just an
+  // estimate, because it is not possible to determine the actual record count
+  // upfront when consuming from a topic based just on the start and end
+  // offsets. The estimation is still useful for getting an idea how many
+  // records there are, e.g. to create a progress bar.
+  optional uint64 estimated_record_count = 5;
+
+  // The name of the topic that was backed up.
+  optional string topic_name = 6;
+
+  // The partition count of the topic that was backed up.
+  optional uint32 partition_count = 7;
+
+  optional ChecksumAlgorithm checksum_algorithm = 8;
+}
+
+// Available checksum algorithms for the `Record.checksum` field. The checksum
+// is used to ensure integrity of the included records, it does not provide any
+// security.
+//
+// The suffix `BE` stands for big endian, and `LE` for little endian. This is
+// important for comparing the raw checksum bytes.
+enum ChecksumAlgorithm {
+  CHECKSUM_ALGORITHM_UNSPECIFIED = 0;
+  CHECKSUM_ALGORITHM_XXHASH3_64_BE = 1;
+}
+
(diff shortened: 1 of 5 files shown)


karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,140 @@
...
+  // humans, as such it **MUST** consist of printable NFC UTF-8 codepoints only.
+  // The length **MUST NOT** exceed 64 bytes.
+  //
+  // Always `Karapace` for backups created by Karapace.
+  string tool_name = 1;
+
+  // The version of the tool that created the backup. The version is intended
+  // for humans, as such it **MUST** consist of printable NFC UTF-8 codepoints
+  // only. The length **MUST NOT** exceed 64 bytes. The version **SHOULD** be
+  // [SemVer](https://semver.org/) compatible.
```
jjaakola-aiven, line 53: Note: Python projects use PEP440 which has some incompatibilities with SemVer.

karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,140 @@
...
+//   Compression of the backups is used to reduce size, and we are not worried
+//   about the additional bytes required for encoding the framing bytes.
+syntax = "proto3";
+
+package karapace.backup.v3;
+
+// Available checksum algorithms for the `Record.checksum` field. The checksum
+// is used to ensure integrity of the included records, it does not provide any
+// security.
+enum ChecksumAlgorithm {
```
jjaakola-aiven, line 35: String names instead of enumeration? Would be more flexible and adding new algorithm would not need change to protobuf IDL?

karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,140 @@
...
+  // that the algorithm that is used **MUST** produce values with a fixed
+  // endianness across all systems.
+  //
+  // The standard algorithm used by Karapace is xxHash3 64 which always produces
+  // its digest in big endian.
+  bytes checksum = 2;
+}
+
+// The record is a minimal representation of a Kafka record, just enough to
+// restore it. There is no partition index included here, because schema topics
```
jjaakola-aiven, line 93: Partition is in the data now.

karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,140 @@
...
+  // The offset the record had in the backed up topic. This is purely
+  // informational, the record is going to have a new offset after restoration.
+  uint64 offset = 5;
+
+  // The UNIX time the record had in the backed up topic.
+  int64 timestamp = 6;
+
+  // The type of timestamp the record has. At the time of writing valid values
+  // are `-1` for none, `0` for creation, and `1` for log append time.
+  int32 timestamp_type = 7;
```
jjaakola-aiven, line 123: Timestamp type could be an enumeration?

karapace/backup/v3/schema.proto
Fleshgrinder: It's a **SHOULD**, so that is fine.

karapace/backup/v3/schema.proto
Fleshgrinder: Using a string begs the question how to map from the string to an actual implementation. The enum makes this trivial. It also provides us with backward and forward compatibility guarantees, something a string cannot.

karapace/backup/v3/schema.proto
Fleshgrinder: I initially used an enum, but an enum requires us to map things (or do some math), because it is not possible to have `-1`. But a simple `timestamp_type.value - 1` would solve that… let‘s do it. 😊

karapace/backup/v3/schema.proto
Fleshgrinder: Let‘s just add PEP440 here too, since we are using Python still.

karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,152 @@
...
+//   is only a 32-bit value. Using fixed 4 bytes means that we can just read.
+//   Compression of the backups is used to reduce size, and we are not worried
+//   about the additional bytes required for encoding the framing bytes.
+syntax = "proto3";
+
+package karapace.backup.v3;
+
+// The metadata directly follows the magic marker version bytes. It contains
+// additional information regarding the backup.
+message Metadata {
```
jjaakola-aiven, line 34: I'd like to have also the backup format version in the Metadata, not only in the header of the backup file.

karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,152 @@
...
+//    the `file` utility).[^1]
+// 2. The second byte is the backup format version, an unsigned 8-bit integer.
+//    This is always `\x03` for the backup format 3.[^1][^2]
+// 3. The next 4 bytes contain the byte count of the `Metadata` message that
+//    follows. The length is an unsigned 32-bit little-endian integer.[^3]
+// 4. The `Metadata` message.
+// 5. Zero or more records follow:
+//    1. 4 bytes containing the unsigned 32-bit little-endian integer byte count
+//       of the `Envelope` message that follows.[^3]
+//    2. The `Envelope` message.
```
jjaakola-aiven, line 19: For now the magic marker is needed. But could the rest of the file have TLV-format instead of just `<length><record>`? I am proposing: `[<tag><length><record>]`. Here it would be `<metadata_tag><length><metadata record>[<envelope_tag><length><envelope record>]`

karapace/backup/v3/schema.proto
jjaakola-aiven: Asking for a duplicate. I edited the first comment as it was missing the reasoning, I am sorry as the original question was not clear.

karapace/backup/v3/schema.proto
Fleshgrinder: I think we should introduce `v4` if we want to change the file layout itself. Adding tags means that we need to manage the back-/forward compatibility for them somehow, and we have no tooling for that, unless we write it. Conventions are good, but they break apart easily. Using a fixed format means that there is no leeway, and thus nothing to manage.

I don't see us ever adding anything in between. Adding more fields to any of the messages is already possible, thanks to Protobuf, and with buf we also have the tooling to verify back-/forward compatibility of all changes.

Unless you have other reasons for proposing tags?

jjaakola-aiven: Offline discussion, tags not needed. Rely on the Protobuf to handle the back-/forward compatibility.

tvainika:
I didn't spot any issues with the backup format schema, but I think it should go in only with first implementation. Do you really want this to go in now before the implementation?

Fleshgrinder:
Yes. First, I don't want to have massive patches, but rather small ones that are easy to review. Second, having the Protobuf definition in master is what is going to allow us to add the breaking checks.

jjaakola-aiven:
I'd propose the implementation to be done in a feature branch.

Fleshgrinder:
Why the complexity? We can implement it without affecting anything else. There is not even a need for feature flags or anything.

karapace/backup/v3/schema.proto
```diff
@@ -0,0 +1,147 @@
...
+
+  // The UNIX time in seconds the backup creation started.
+  optional int64 create_time = 4;
+
+  // The estimated record count included in this backup. This is just an
+  // estimate, because it is not possible to determine the actual record count
+  // upfront when consuming from a topic based just on the start and end
+  // offsets. The estimation is still useful for getting an idea how many
+  // records there are, e.g. to create a progress bar.
+  optional uint64 estimated_record_count = 5;
```
aiven-anton, line 66: Since we're individually serializing each record, why is it not possible to count them?

aiven-anton:
I think it would be nice to some reasoning about why we are inventing our own format here over Kafka's internal RecordBatch. A `RecordBatch` already seems to contain the same data, including headers, and has checksumming. Additionally, using `RecordBatch` it seems we would be available to directly stream data from files into Kafka without intermediate processing, possibly enabling speeding up the recovery process further?

aiven-anton:
> I think it would be nice to some reasoning about why we are inventing our own format here over Kafka's internal RecordBatch. A `RecordBatch` already seems to contain the same data, including headers, and has checksumming. Additionally, using `RecordBatch` it seems we would be available to directly stream data from files into Kafka without intermediate processing, possibly enabling speeding up the recovery process further?

I raised this idea internally, we had a few discussions around it, and ultimately the idea was rejected on these grounds:

- It's probably much more feasible in terms of existing tooling to implement the custom protobuf schema of this PR.
- The proposed idea in this PR is probably fast enough.
- It's unclear whether a full `RecordBatch` (at least all fields covered by checksum) can actually be produced at backup-time. If we can't, the idea is more or less useless.
- OTOH, nothing would prevent us from revisiting this in the future, if our performance needs would surprisingly not be met by V3.

aiven-anton:
This has been superseded by #584.

PR #550, opened 2023-02-21
feat: Add `shellcheck` and `shfmt` Pre-Commit Hooks

Fleshgrinder:
This adds `shellcheck` and `shfmt` pre-commit hooks. I fixed the most pressing issues that were found, but did not go through the existing scripts otherwise. We are going to need more shell scripts to automate other things, this is why it is a good idea to add these tools to our verification chain.

~This also changes the GHA lint workflow to use the official pre-commit action, which also makes use of caching, and thus should be faster.~ The pre-commit action uses the `--color=always` option, and that option results in the action hanging forever if paired with local hooks, which we have.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,13 +1,11 @@
 # testing
+filelock==3.9.0
 pytest==6.2.5
 pytest-xdist[psutil]==2.2.1
 pytest-timeout==1.4.2
 pdbpp==0.10.2
 psutil==5.9.0
 requests==2.27.1
-
-# linting
-pre-commit>=2.2.0
 
 # performance test
 locust==2.13.0
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+root = true
+
+[*]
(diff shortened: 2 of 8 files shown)


requirements-dev.txt
```diff
@@ -6,9 +6,6 @@ pdbpp==0.10.2
 psutil==5.9.0
 requests==2.27.1
 
-# linting
-pre-commit>=2.2.0
```
Fleshgrinder, line 10: We are currently managing this version in two places – here and in the `.pre-commit-config.yaml` file – and they already diverged. There is no reason to include it here, users who wish to use it can install it on their own (in any way they like), and GitHub Actions uses the pre-commit action now which is versioned separately.

Fleshgrinder: I had to remove the action, because it is broken with local hooks, and I had to add `filelock` as direct dependency, because it was used indirectly. Still, the sentiment towards having this dependency stays the same on my side.

jjaakola-aiven (approved):

PR #551, opened 2023-02-22
chore: remove untracked files from copyright header check

jjaakola-aiven:

Remove untracked files from pre-commit copyright header check.

# Why this way

Some times, at least I, have some special purpose scripts that are not tracked and won't be commited. The pre-commit hook then fails for missing copyright headers. This corrects the behavior to select only the files tracked by Git.


aiven-anton (approved):
I think the best effort thing we could do here is to make the script accept filenames as input from pre-commit, but this seems like an improvement.

PR #555, opened 2023-03-02
Implement health status endpoint

### LINKED ISSUE #143: Implement endpoint for health check
Bugs in dependency libraries can cause the karapace APIs to become unresponsive, restarting the process is usually a valid and quick fix for the problem. To properly implement the restart a `/status` endpoint is necessary, this endpoint has to peform some request through the kafka client to make sure communication is working properly.

tvainika:

Add health status endpoint
    
Fixes #143

# Why this way

Implement health status endpoint which returns status of schema registry.  This by design avoids long running operations, and queries status from internal state variables acquiring only minimal locks.  This is designed to robustly and quickly return internal status of Karapace.

### DIFF
--- a/karapace/master_coordinator.py
+++ b/karapace/master_coordinator.py
@@ -4,6 +4,7 @@ karapace - master coordinator
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
+from dataclasses import dataclass
 from kafka import KafkaConsumer
 from kafka.coordinator.base import BaseCoordinator
 from kafka.errors import NoBrokersAvailable, NodeNotReadyError
@@ -38,6 +39,15 @@ def get_member_configuration(*, host: str, port: int, scheme: str, master_eligib
         "scheme": scheme,
         "master_eligibility": master_eligibility,
     }
+
+
+@dataclass
+class SchemaCoordinatorStatus:
+    is_primary: Optional[bool]
+    is_primary_eligible: bool
+    primary_url: Optional[str]
+    is_running: bool
+    group_generation_id: int
 
 
 class SchemaCoordinator(BaseCoordinator):
@@ -211,6 +221,16 @@ class MasterCoordinator(Thread):
             request_timeout_ms=max(session_timeout_ms, KafkaConsumer.DEFAULT_CONFIG["request_timeout_ms"]),
         )
         self.schema_coordinator_ready.set()
+
+    def get_coordinator_status(self) -> SchemaCoordinatorStatus:
+        generation = self.sc.generation() if self.sc is not None else None
+        return SchemaCoordinatorStatus(
+            is_primary=self.sc.are_we_master if self.sc is not None else None,
+            is_primary_eligible=cast(bool, self.config["master_eligibility"]),
+            primary_url=self.sc.master_url if self.sc is not None else None,
+            is_running=self.is_alive(),
+            group_generation_id=generation.generation_id if generation is not None else -1,
+        )
 
     def get_master_info(self) -> Tuple[Optional[bool], Optional[str]]:
         """Return whether we're the master, and the actual master url that can be used if we're not"""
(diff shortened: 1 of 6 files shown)


karapace/master_coordinator.py
```diff
@@ -212,6 +222,16 @@ def init_schema_coordinator(self) -> None:
         )
         self.schema_coordinator_ready.set()
 
+    def get_coordinator_status(self) -> SchemaCoordinatorStatus:
+        generation = self.sc.generation() if self.sc is not None else None
+        return SchemaCoordinatorStatus(
+            is_primary=self.sc.are_we_master if self.sc is not None else None,
+            is_primary_eligible=cast(bool, self.config["master_eligibility"]),
```
jjaakola-aiven, line 229: This is available from `SchemaCoordinator.master_eligibility`.
https://github.com/aiven/karapace/blob/2ab4e832549274442ac13a175c821a5e9f27890b/karapace/master_coordinator.py#L73

karapace/master_coordinator.py
```diff
@@ -212,6 +222,16 @@ def init_schema_coordinator(self) -> None:
...
+            primary_url=self.sc.master_url if self.sc is not None else None,
+            is_running=self.schema_coordinator_ready.is_set(),
```
jjaakola-aiven, line 231: The ready flag been set is known but not if the thread is actually still running, hanging or failing. The more correct would be `self.running`, but it does not also tell if the loop is performing.

karapace/schema_registry_apis.py
```diff
@@ -84,6 +84,22 @@ def __init__(self, config: Config) -> None:
...
+        resp["schema_registry_ready"] = self.schema_registry.schema_reader.ready
+        resp["schema_registry_reader_offset"] = self.schema_registry.schema_reader.offset
+        resp["schema_registry_reader_highest_offset"] = self.schema_registry.schema_reader.highest_offset
+        cs = self.schema_registry.mc.get_coordinator_status()
+        resp["schema_registry_is_primary"] = cs.is_primary
+        resp["schema_registry_is_primary_eligible"] = cs.is_primary_eligible
+        resp["schema_registry_primary_url"] = cs.primary_url
+        resp["schema_registry_coordinator_running"] = cs.is_running
+        resp["schema_registry_coordinator_generation_id"] = cs.group_generation_id
+        return resp
```
jjaakola-aiven, line 102: The highest offset is only read on startup so I'd make that explicit on the field name in JSON. See also other field name suggestions in the proposed format below.
And I would add also the current offset. And maybe add the highest offset from `OffsetWatcher`, the difference would be the lag and it should be small (0 or 1 in most cases). If the lag goes up then there is a problem.

Maybe format could be this?
```
{
  "schema_registry": {
    "authfile_modified_ts": self._auth.auth...,
    "ready": ..,
    "highest_offset_on_start": ..,
    "current_offset": ...,
    ....
  }
}
```

karapace/master_coordinator.py
tvainika: However `SchemaCoordinator` might not be initialized in the beginning ie. `self.sc is None`...

karapace/master_coordinator.py
tvainika: Good point. Looking more, I'm thinking if `Thread.is_alive()` would serve best way here...

karapace/schema_registry_apis.py
tvainika: I considered level structured response as there can be REST proxy related state variables too, but ended up with flat response for sake of simplicity.

Current code exposes current offset ie. the offset up to where schema reader has been reading messages. Did you meant to suggest about naming `schema_registry_reader_offset` -> `...current_offset` or did I miss something?

Incorporating `OffsetsWatcher`'s offset it is waiting into same exposed highest offset sounds like a plan to me.

jjaakola-aiven (approved):

PR #561, opened 2023-03-21
refactor: extract Kafka producer from Registry

jjaakola-aiven:

Extract Kafka producing from `karapace/schema_registry.py` to own module.
The `OffsetsWatcher` is extracted from `karapace/schema_reader.py`.


aiven-anton (approved):
LGTM! 👍

aiven-anton:
@RyanSkraba I noticed the same, i.e. instantiation of `KafkaProducer` now happens in `KarapaceSchemaRegistry.start()` instead of `KarapaceSchemaRegistry.__init__()`.

@jjaakola-aiven Do you maybe just want to confirm this is intentional, and I'm happy to merge this.

jjaakola-aiven:
@aiven-anton it is intentional, did not really though it as functional change when moved, which it clearly is.
The producer is not used until the instance becomes primary. The order of starting the schema reader or producer is not that critical, but as the startup schema reading will block the writing until state is finished I thought to put producer start as last.

RyanSkraba:
Thanks!

PR #565, opened 2023-03-22
fix: Don't dictate local Python version

aiven-anton:

Remove `.python-version` file.

### Why this way

There's no need to dictate the exact Python version developed against locally by contributors, generally it's not advisable to keep `.python-version` under version control. If we want to help in testing against multiple Python versions locally, we should set up something like tox.


Fleshgrinder:
You can have multiple Python versions with `.python-version` to, just issue `pyenv local 3.7 3.8 3.9 3.10 3.11 3.12-pre` and commit it. The file is also not _dictating_ the Python version, it uses a safe default (which in our case is the lower bound).

aiven-anton:
@Fleshgrinder That's not how `.python-version` is meant to be used, it's meant to be used. It's meant to contain the versions you want to expose, and those _all_ need to be installed to not yield user errors.

Further, to actually make use of the virtualenv created, it should not contain `3.7`, it should contain the name of the virtualenv.

> A user can always chose a different Python version.

No, not without maintaining an annoying diff with upstream. There really isn't any need for this.

aiven-anton:
But most importantly, this change doesn't in any way limit any of the usages that you mention.

aiven-anton:
Hmm, I missed pushing the changes to .gitignore and makefile, probably the cause of the confusion?

Fleshgrinder:
There is no need to change the `.python-version` file if you want to use a different Python version:

```bash
PYENV=<path to your python exe> make install
```

The difference is in user experience. Not all Python versions are available via normal package managers, whereas `pyenv` can provide almost all versions. Most notable is Python 3.7 which is not available on Mac via Homebrew. The `.python-version` sets a safe default for a great out-of-the-box experience for those who just want to get started quickly. Just as reported in https://github.com/aiven/karapace/issues/553 (the origin of this addition). Using any Python version other than 3.7 for development means that the editor/IDE thinks that is can do things that is actually cannot do. Whereas using 3.7 ensures that the editor/IDE knows what is possible, and what is not.

aiven-anton:
@Fleshgrinder I am not at all taking issue with use of pyenv or the `.python-version` file, that's what I'm using myself. What I am taking issue with is keeping `.python-version` under version control.

Fleshgrinder:
I fail to see the difference between `echo '3.7' >.python-version` and simply having the file there already. Setting a different version locally should not be done by altering files in the repository. Either use the `PYENV` environment variable or the `PYTHON_VERSION` environment variable, is intended by pyenv for exactly this use case.

RyanSkraba (approved):
LGTM

aiven-anton:
@Fleshgrinder FWIW, it's really the same reason why it's a bad practice to keep `.idea/` under version control, it restricts things that don't need restricting.

RyanSkraba (approved):
Sorry :point_up: should have been an Approve, not a comment!

aiven-anton:
@RyanSkraba I think it was an approve, but it automatically became dismissed after I pushed further changes, so all good 👍

Skärmavbild 2023-03-22 kl  16 28 58

tvainika (approved):

PR #568, opened 2023-04-03
chore: Bump pre-commit hooks

aiven-anton:


karapace/protobuf/syntax_reader.py
```diff
@@ -345,7 +345,7 @@ def skip_whitespace(self, skip_comments: bool) -> None:
             elif skip_comments and c == "/":
                 self.read_comment()
             else:
-                return None
+                return
```
juha-aiven, line 348: Interesting. `None` is no longer needed?

juha-aiven (approved):
It looks like the changes caused are mostly removing blank lines, but there are some trivial code changes too. For example, changing of `if`s to asserts. I had a look at these and they seem to be correct and make sense.

Also, in some places `Exception` has been replaced with `RuntimeError`. This should be fine.

karapace/protobuf/syntax_reader.py
aiven-anton: I knew I should have made notes of why this change was needed 😅 Backtracking now, adding the `None` back and running pylint shows this:

```
karapace/protobuf/syntax_reader.py:335:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
```

[Here's documentation for inconsistent-return-statements](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/inconsistent-return-statements.html). I think the rule kind of makes sense, but didn't think too much about it tbh :) OTOH, it doesn't feel like a rule that will make much difference in terms of maintainability/readability 🤷

aiven-anton:
> It looks like the changes caused are mostly removing blank lines, but there are some trivial code changes too. For example, changing of `if`s to asserts. I had a look at these and they seem to be correct and make sense.
> 
> Also, in some places `Exception` has been replaced with `RuntimeError`. This should be fine.

Yeah, sorry, I should have added the reasoning to the commit message/PR description. These changes were both due to another new pylint rule, to not raise bare `Exception`, which I agree with. The `RuntimeError` seemed appropriate in `version.py`. I think all other cases were within tests, where it just seemed to make more sense to use `assert` instead.

PR #571, opened 2023-04-03
Add type hints to schema_backup

aiven-anton:

- Add type hints to schema_backup.
- Add TypedDict to describe configuration.
- Fix non-functioning overload of json_encode.
- anonymize_avro_schema_message() is updated to adhere to the interface stipulated by SchemaBackup.create(), i.e. must accept None.

### DIFF
--- a/karapace/anonymize_schemas/anonymize_avro.py
+++ b/karapace/anonymize_schemas/anonymize_avro.py
@@ -5,6 +5,7 @@ Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
 from typing import Any, Dict, List, Union
+from typing_extensions import TypeAlias
 
 import hashlib
 import re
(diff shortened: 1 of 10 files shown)


aiven-anton:
> What comes to overloading `json_encode` using `binary` flag, my opinion is that having two functions, one for binary and one for non-binary json encoding would be better than overloading same named function.

I don't have a strong opinion either way, but the overload was currently not setup in a way that properly distinguishes the return type on the `binary` argument (`Literal` is required to make that work). I think that's why I'm able to remove two type-ignore comments where it's used. 

An unrelated, but tangential thing is I've looked at some of the backup code, and I suspect we make a few unnecessary conversions between str/bytes there. In that case at least, I think we should be able to deal almost exclusively with bytes.

I personally think the overload in itself is fine, but wouldn't be against a refactor that splits it up.

tvainika (approved):

PR #573, opened 2023-04-03
Use pip-tools for dependency management

aiven-anton:

Introduce pip-tools for dependency management.

References: #566 

### Why this way

See linked discussion. In summary:

- Dependency resolution is tedious, and computers do it better than humans.
- Much higher chance of having reproducible builds, as this way makes sure all actual dependencies are pinned.
- Transitive dependencies clean themselves up, if we no longer transitively depend on a package, `make requirements` will remove it from requirements.txt.


requirements/requirements.txt
```diff
@@ -0,0 +1,90 @@
...
+    # via -r requirements/requirements.in
+watchfiles==0.19.0
+    # via -r requirements/requirements.in
+yarl==1.8.2
+    # via aiohttp
+zipp==3.15.0
+    # via importlib-metadata
+
+# The following packages are considered to be unsafe in a requirements file:
+# setuptools
```
aiven-anton, line 90: This warning happens because `jsonschema` specifies setuptools as a dependency. It looks like this is fixed in latest major version (4.*).

tvainika (changes requested):
conflicts

tvainika (approved):

PR #576, opened 2023-04-05
chore: Start adding types to karapace.protobuf

aiven-anton:


karapace/protobuf/reserved_element.py
```diff
@@ -4,21 +4,23 @@
...
 from karapace.protobuf.location import Location
 from karapace.protobuf.utils import append_documentation
+from karapace.utils import assert_never
 
 
 @dataclass
 class ReservedElement:
     location: Location
+    values: list[str | int | KotlinRange]
     documentation: str = ""
```
aiven-anton, line 20: Note: Changed order of fields here, because `values` doesn't have a sensible default (inferred by `.to_schema()` lacking the ability to handle `value: None`), and since we need to support Python 3.8, we cannot add `kw_only`.

PR #578, opened 2023-04-09
Introduce multi-arch builds with support for ARM

### LINKED ISSUE #521: Provide images for the arm64 architecture
# What is currently missing?

First world problem, but it would be awesome to have arm64 images. The main use (for us) would be for development and integration testing on Apple silicone. 

# How could this be improved?

Add or modify the container build step to also produce arm64 images. 

# Is this a feature you would work on yourself?
If you support the idea, then yes. The change itself should be straight forward. 
[x] I plan to open a pull request for this feature

aiven-anton:

With this commit we start building multi-arch container images to enable running containers on ARM systems.

Closes #521.

References #585.

### Why this way

I did experimentation/debugging in this repository: https://github.com/aiven-anton/test-build-arm, see results of build [12](https://github.com/aiven-anton/test-build-arm/actions/runs/4650710382) and [13](https://github.com/aiven-anton/test-build-arm/actions/runs/4650712617).

Using the Docker actions makes for very low amount of boilerplate code for building containers.


tvainika (changes requested):
I suspect this does not work.

Sadly our container builds from GitHub Actions are triggered only on main branch. Therefore no container CI action was triggered by this pull request.  Therefore I pushed this branch into my own fork's main branch, and this resulted build failure https://github.com/tvainika/karapace/actions/runs/4675463034

I think build failure is because `Gather context` step does not have all tags with fetched history. Actions/checkout needs `fetch-depth` flag to fetch all tags.

aiven-anton:
> Sadly our container builds from GitHub Actions are triggered only on main branch.

Perhaps I should start in this end, building an image for PRs and adding some very basic smoke test. That would probably make it easier to gain confidence for a PR like this one later on. Very good idea that you tried this out on a fork! 🙏

The failure is a bit odd though, the `$(git describe --always)` from the "Gather context" step was taken as-is from the current "Build and push container image" step, and checkout hasn't changed so I think if git history/tags is too shallow now, it must have been before too. I think what's triggering this is the explicit passing of `KARAPACE_VERSION`, which means Karapace is no longer attempting to gather the version from git itself. Inspecting [the line](https://github.com/aiven/karapace/blob/41fe18bd0ef8eaed4bbd2cd6e9962e00768f8a64/version.py#L33) where it used to, it's clear that's a different invocation than in "Gather context": it has `--tags`. 

I'll put this PR in draft for now, and will try to come back to it once we have smoke tests 🙂

aiven-anton:
@tvainika I managed to get the build working on my fork now, I did these things:

- Make sure there are actually tags in the repository (did not get cloned by default).
- Set `fetch-depth: 0` as you suggested.
- Add `--tags` to `git describe` invocation.

Passing build: https://github.com/aiven-anton/karapace/actions/runs/4872183301/jobs/8690049653

tvainika (approved):

PR #579, opened 2023-04-09
Optimize container build time

aiven-anton:

- Use cache mounts for pip and apt, heavily reducing network requests when cache is warm.
- Remove git as a build-time dependency, other than switching to archive endpoints as was done in https://github.com/aiven/karapace/commit/39b4f0850c3eaa6ef0d41a7b4e152cc4291ab0f3, this also requires either setting `KARAPACE_VERSION` as container build arg or building karapace/version.py independently.
- Add hadolint pre-commit check for linting Dockerfile.
- Move to using Python base images for builder and final stage. This allows omitting installation of some build tools. It also allows moving to a more recent Python version, no longer being bound by what's in distro repositories. Wheel availability of some of our Python dependencies prevents us from moving to 3.11 for now.
- Change installation approach to construct a virtualenv in the builder step, and copying it unaltered to the final stage, with dependencies and Karapace itself installed in it. This allows having even fewer layers in the final stage, and is simpler.
- Introduces a _much_ stricter .dockerignore, ignoring files by default and explicitly including what's required. This makes sure changes in unrelated files does not evict layer cache. For example, a few files that previously erroneously evicted caches, because everything was included:
  - `.git/*`
  - `.mypy_cache/*`
  - `container/Dockerfile` itself
  - `__pycache__/*`
  - `.idea/*`

Image size remains the same as previously (the `develop` tag is the new version):

```sh
$ docker image ls
REPOSITORY                           TAG               IMAGE ID       CREATED          SIZE
ghcr.io/aiven/karapace               develop           30efca01fe54   6 seconds ago    203MB
ghcr.io/aiven/karapace               latest            ef8485cb1181   3 weeks ago      203MB
```

References #585.

### Why this way

- Removing git as a build-time dependency is a simplification.
- Caching optimizes, makes building images faster. 
- Hadolint increases quality of Dockerfile, helps with maintainability.
- Pinning apt requirements to adhere to a Hadolint rules. This ensures reproducible builds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,11 +1,19 @@
-# Ignoring files that are specific to a given checkout, these change based on
-# the user commands and not on the repository history. They are not important
-# to determine the state of the repository and would invalidate the cache
-# layer.
-#
-# - .git/logs/HEAD - command history
-# - .git/index - binary file for the current index, very important for a
-#    working repository, not interesting for our image
-#
-.git/logs/HEAD
-.git/index
+# Ignore everything by default. Making as few files as possible part of default context
+# ensures only relevant changes will evict layer cache.
+*
+
+# Include source directories and files required for building.
+!karapace
+!requirements/*.txt
+!setup.py
+!version.py
+!README.rst
+!container/start.sh
+!container/healthcheck.py
+
+# Ignore some files in source directories.
+**/.DS_Store
+**/Thumbs.db
+**/*.pyc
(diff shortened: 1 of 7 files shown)


aiven-anton:
Just like #578, this would PR would gain from a smoke test in place, marking this as draft until we have it too.

tvainika (approved):

PR #580, opened 2023-04-10
Simplify Dockerfile and narrow context 

aiven-anton:

- Move to using Python base images for builder and final stage. This allows omitting installation of some build tools. It also allows moving to a more recent Python version, no longer being bound by what's in distro repositories. Wheel availability of some of our Python dependencies prevents us from moving to 3.11 for now.
- Change installation approach to construct a virtualenv in the builder step, and copying it unaltered to the final stage, with dependencies and Karapace itself installed in it. This allows having even fewer layers in the final stage, and is simpler.
- Introduces a _much_ stricter .dockerignore, ignoring files by default and explicitly including what's required. This makes sure changes in unrelated files does not evict layer cache. For example, a few files that previously erroneously evicted caches, because everything was included:
  - `.git/*`
  - `.mypy_cache/*`
  - `container/Dockerfile` itself
  - `__pycache__/*`
  - `.idea/*`

Image size remains the about the same as previously, but slightly reduced (the `karapace-dev` tag is the new version):

```sh
$ docker image ls
REPOSITORY                           TAG               IMAGE ID       CREATED          SIZE
karapace-dev                         latest            b31c7d811294   10 seconds ago   194MB
ghcr.io/aiven/karapace               develop           b8366c494bec   5 days ago       204MB
ghcr.io/aiven/karapace               latest            ef8485cb1181   3 weeks ago      203MB
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,11 +1,18 @@
-# Ignoring files that are specific to a given checkout, these change based on
-# the user commands and not on the repository history. They are not important
-# to determine the state of the repository and would invalidate the cache
-# layer.
-#
-# - .git/logs/HEAD - command history
-# - .git/index - binary file for the current index, very important for a
-#    working repository, not interesting for our image
-#
-.git/logs/HEAD
-.git/index
+# Ignore everything by default. Making as few files as possible part of default context
+# ensures only relevant changes will evict layer cache.
+*
+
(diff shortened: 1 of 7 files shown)


aiven-anton:
Closed in favor of #579.

PR #581, opened 2023-04-12
Implement backup version 3

aiven-anton:

Continuation of #549.

TBD

### Why this way

TBD

### Todo

- [ ] Verify buf-breaking-action works as intended when protocol files exist on `main`.
- [ ] Rebase on top of #583, make sure branch doesn't have the `Refactor schema backups to separate API and backends` commit.
- [ ] Version protoc must be in sync between runtime (python dep) and the version used to build schemas, or the generated Python code raises import-time errors.
- [ ] Re-evaluate whether protobuf actually gives us any useful guarantees, given it's "everything is optional design". The infrastructure put in place around validating protobuf's compatibility across PR merges, likely does not prevent us from introducing changes that breaks "processability".


.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,30 @@
...
+    - run: buf format -w
+    - run: |
+        diff=$(git --no-pager diff)
+        if [[ ! -z "$diff" ]]; then
+          echo "💥 Non-empty output from git diff, please run buf generate and format."
+          echo
+          echo "$diff"
+          exit 1
+        else
+          echo "✅ No diff to current currency module."
```
aiven-anton, line 25: Bad copy

PR #583, opened 2023-04-13
Refactor schema backups to separate API and backends

aiven-anton:

To accomodate for a third version of the backup on-disk format, the existing formats are re-structured into separate modules under `karapace.backup.vN`. The entirety of backups is moved from `karapace.schema_backups` into the `karapace.backups` package.

### Why this way

The rough idea is that format-specific code will live under the version-specific modules, and act as backends that is being dispatched to from `karapace.schema_backup`.

`karapace.schema_backup` itself had two separate responsibilities: internal API, and exposed CLI. To reflect this in code organization, the file is split into `karapace.backup.api` and `karapace.backup.cli`.


karapace/schema_backup.py
```diff
@@ -274,15 +303,17 @@ def check_dst() -> None:
...
+    finally:
+        # Seek back to start.
         fp.seek(0)
-        fp.readline()
-        return BackupVersion.V2
-    fp.seek(0)
-    return BackupVersion.V1
+    # Consume until linefeed.
+    fp.readline()
+    return version
```
aiven-anton, line 316: These changes are to accommodate for parsing a V3 marker.

karapace/schema_backup.py
```diff
@@ -310,10 +345,16 @@ def restore_backup(self) -> None:
             LOG.info("Starting backup restore for topic: %r", self.topic_name)
 
             with open(self.backup_location, encoding="utf8") as fp:
-                if _check_backup_file_version(fp) == BackupVersion.V2:
-                    self._restore_backup_version_2(producer, fp)
+                version = _read_backup_file_version(fp)
```
jjaakola-aiven, line 348: This could return the correct backend, no need to have `if-elif-else`:
```suggestion

class BackupBackend:
    def items_from_file(self, fp) ->Generator[tuple[str, str], None, None]:
        ...
        
    def serialize_record(self, key_bytes: bytes | None, value_bytes: bytes | None) -> str:
        ...

class V2BackupBackend(BackupBackend):
    ...
---------------

                backend = _select_backup_backend(fp)
                items = backend.items_from_file(fp)
```

The API would also be formalized.

aiven-anton: Hmm, I wrote a long response here on Friday, it seems to have been lost, maybe I forgot to post it.

---

The reason I did not opt to introduce an interface here is because the three versions will be sufficiently different for us to have any valuable gains from a shared interface.

- V1 is only readable, so the only thing it has in common with V2 is the signature of the reader: `(fp: IO) -> Iterator[tuple[str, str]]`.
- V1 and V2 are dealing with `str`, while I'm very much hoping that V3 won't have to make any UTF-8 interpretations at all, i.e. exclusively deal with `bytes`.
- V1 and V2 takes `key` + `value` for serialization. In V3 we need a much richer object to be able to pass all the required metadata, so a shared interface doesn't fit well here either. 

Since the current state of this PR mostly moves things around, I'm hoping we can find the value in its current scope sufficient to move forward, and then consider whether we additionally need further refactoring.

aiven-anton: I can take another round at formalizing the interface.

.pylintrc
```diff
@@ -29,6 +29,7 @@ disable=
   no-name-in-module,
   use-list-literal,
   use-dict-literal,
+  no-value-for-parameter,
```
aiven-anton, line 32: This gives false positive, and is something mypy handles much better.

karapace/backup/backend.py
```diff
@@ -0,0 +1,104 @@
...
+            key, value = item
+            yield ProducerSend(
+                topic_name=topic_name,
+                key=self.key_encoder(key),
+                value=self.value_encoder(value),
+                partition=PARTITION_ZERO,
+            )
+
+
+class BaseBackupProducer(abc.ABC, Generic[B]):
```
aiven-anton, line 78:
```suggestion
class BaseBackupWriter(abc.ABC, Generic[B]):
```

`s/Producer/Writer/` probably makes more sense than "producer" here, it's distinct from the other kind of "producer" that's involved (`KafkaProducer`), and it mirrors "BackupReader".

jjaakola-aiven (dismissed):

PR #587, opened 2023-04-18
chore: formalize public API surface and versioning

### LINKED ISSUE #245: Proper releases
# What is currently missing?

- [x] tagged releases
- [x] docker image realeases
- [ ] PyPI package releases

aiven-anton:

Document what is to be considered public API, and what's not. Document use of semantic versioning.

Closes #245.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -553,10 +553,24 @@ Example of complete authorization file
         ]
     }
 
+Public API surface and versioning
+=================================
+
+Karapace is mainly an application and decidedly not a library. It's exposing a
+REST API, as well as a few commands documented here. It's important to note that
+only those documented entrypoints are considered "public" APIs, everything else
+is considered internal implementation details and should be expected to change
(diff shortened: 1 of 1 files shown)


tvainika (changes requested):
There is much earlier prominent note about REST API being compatible with Confluent Schema Registry and Kafka REST Proxy 6.1.1. I think this note is out of place, and does not also reflect our aim for compatibility.

In addition to public API endspoints (REST), we have some data formats (e.g. `_schemas` topic for schema registry), which has compatibility requirements and cannot be considered only internal implementation details.

aiven-anton:
> In addition to public API endspoints (REST), we have some data formats (e.g. _schemas topic for schema registry), which has compatibility requirements and cannot be considered only internal implementation details.

True, it's probably also worth mentioning. It's splitting hairs, but I consider it slightly different from defining our API surface thiugh, it's more like a "supported upgrade paths" definition, right?

PR #599, opened 2023-05-03
Fix references validation from request

### LINKED ISSUE #591: Creating a schema from `provectus/kafka-ui` results in a `TypeError: 'NoneType' object is not iterable`
# What happened?

When creating a schema from the `provectus/kafka-ui` project, it sends a post like that:

`POST with subject: 'aa', request: {'schema': '{}', 'schemaType': 'JSON', 'references': None}`

Note the `references` set to None. This results in this traceback:

```
  File "/usr/local/lib/python3.9/dist-packages/karapace/rapu.py", line 326, in _handle_request
    data = await callback(**callback_kwargs)
  File "/usr/local/lib/python3.9/dist-packages/karapace/schema_registry_apis.py", line 1143, in subject_post
    references = self._validate_references(content_type, schema_type, body)
  File "/usr/local/lib/python3.9/dist-packages/karapace/schema_registry_apis.py", line 1010, in _validate_references
    for reference in references:
TypeError: 'NoneType' object is not iterable
```

This should not traceback. If None is an invalid value here then it should at least return a message saying so and an issue should be opened on the kafka-ui project.

The support for this key was added by 7e1de2a75bfba2f7ef2087a37907e25f21dc5e28. I'm running the latest `develop` image at the time of writing this (`sha256:2cf7d6202aea913d696f25b20705d84672d55b6bd461d83385f05791c1245fbe`).

tvainika:

Fix references validation from request

# Why this way

Handle null as empty list of `references`, and reject wrongly typed values.  Fixes #591

### DIFF
--- a/tests/integration/test_dependencies_compatibility_protobuf.py
+++ b/tests/integration/test_dependencies_compatibility_protobuf.py
@@ -356,3 +356,56 @@ async def test_protobuf_schema_compatibility_dependencies2(registry_async_client
     )
     assert res.status_code == 200
     assert res.json() == {"is_compatible": False}
+
+
+SIMPLE_SCHEMA = """\
+syntax = "proto3";
+
+message Msg {
+  string name = 1;
+}
+"""
+
+
+async def test_protobuf_schema_references_rejected_values(registry_async_client: Client) -> None:
+    subject = create_subject_name_factory("test_protobuf_schema_references_values")()
+    res = await registry_async_client.put(f"config/{subject}", json={"compatibility": "BACKWARD"})
+    assert res.status_code == 200
+
+    res = await registry_async_client.post(
+        f"subjects/{subject}/versions", json={"schemaType": "PROTOBUF", "schema": SIMPLE_SCHEMA, "references": 1}
+    )
+    assert res.status_code == 400
+
+    res = await registry_async_client.post(
+        f"subjects/{subject}/versions", json={"schemaType": "PROTOBUF", "schema": SIMPLE_SCHEMA, "references": "foo"}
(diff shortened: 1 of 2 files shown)


tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -356,3 +356,52 @@ async def test_protobuf_schema_compatibility_dependencies2(registry_async_client
     )
     assert res.status_code == 200
     assert res.json() == {"is_compatible": False}
+
+
+async def test_protobuf_schema_references_values(registry_async_client: Client) -> None:
```
aiven-anton, line 361: suggestion: This test seem to describe a few different expectations, baked into one test case. Can we break this apart into test with descriptive names, such as "returns bad request when {x,y}" and "returns http OK when z"? I find that greatly helps understanding original intent when debugging test failures.

aiven-anton (approved):

tests/integration/test_dependencies_compatibility_protobuf.py
tvainika: I separated rejected and valid values, ie. two test cases.

aiven-anton (approved):

tests/integration/test_dependencies_compatibility_protobuf.py
aiven-anton: I think this is OK, it's kind of a pet peeve of mine anyway. In my ideal world each case would have it's own descriptive name, so that one could deduce _"ah, we're expecting a rejection here because [some good reason]"_.

PR #560, opened 2023-03-14
Add dependencies/references support

libretto:
Protobuf references support:
Add references to Protobuf schemas,
Support compare protobuf schemas with references,
Support serialization/deserialization of protobuf schemas with dependencies.

PR #610, opened 2023-05-10
Improve references support

tvainika:

* Include references in responses
* Add support to latest version with references

Both are to better support `references` feature implemented in #560 but missing parts of full references support.

# Why this way

Schema fetch and list operations should contain references if any in the response.

Registering a schema with a reference to version -1 is a reference to latest version of referenced schema.

Also minor cleanups, including eliminating of cached string representation of non-protobuf schemas, which should not have been changed with #560.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -524,9 +524,16 @@ class KafkaSchemaReader(Thread):
         return self.database.get_referenced_by(subject, version)
 
     def _resolve_reference(self, reference: Reference) -> Dependency:
-        subject_data = self.database.find_subject_schemas(subject=reference.subject, include_deleted=False)
+        subject_data: Dict[ResolvedVersion, SchemaVersion] = self.database.find_subject_schemas(
+            subject=reference.subject, include_deleted=False
+        )
         if not subject_data:
             raise InvalidReferences(f"Subject not found {reference.subject}.")
+
+        # -1 is alias to latest version
+        if reference.version == -1:
+            reference.version = max(subject_data)
+
         schema_version: SchemaVersion = subject_data.get(reference.version, None)
         if schema_version is None:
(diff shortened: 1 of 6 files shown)


karapace/schema_registry_apis.py
```diff
@@ -446,15 +446,16 @@ async def schemas_list(self, content_type: str, *, request: HTTPRequest, user: U
...
-                )
+                response_schema = {
+                    "subject": schema_version.subject,
+                    "version": schema_version.version,
+                    "id": schema_version.schema_id,
+                    "schemaType": schema_version.schema.schema_type,
+                }
+                if schema_version.references:
+                    response_schema["references"] = [r.to_dict() for r in schema_version.references]
+                response_schema["schema"] = schema_version.schema.schema_str
```
aiven-anton, line 457: question: Why not set this in the dict literal above?

karapace/schema_reader.py
```diff
@@ -524,9 +524,16 @@ def get_referenced_by(self, subject: Subject, version: ResolvedVersion) -> Optio
         return self.database.get_referenced_by(subject, version)
 
     def _resolve_reference(self, reference: Reference) -> Dependency:
-        subject_data = self.database.find_subject_schemas(subject=reference.subject, include_deleted=False)
+        subject_data: Dict[ResolvedVersion, SchemaVersion] = self.database.find_subject_schemas(
```
aiven-anton, line 527: thought: This explicit annotation shouldn't be needed, since it's the exact return type of the called method.

karapace/schema_registry_apis.py
tvainika: I tried to keep ordering of keys. Actually elsewhere our json responses are sorted by key, not sure if it was intentional or not.

PR #611, opened 2023-05-10
chore: Add type hints karapace.schema_reader

aiven-anton:

- Add type hints to `karapace.schema_reader`. While doing this, discovered a few inconsistencies needing addressing to pass static type checking.
- Replace type aliases with `NewType`s.

### Why this way

Increase type coverage.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -4,6 +4,8 @@ karapace - Kafka schema reader
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
+from __future__ import annotations
+
 from avro.schema import Schema as AvroSchema
 from contextlib import closing, ExitStack
 from jsonschema.validators import Draft7Validator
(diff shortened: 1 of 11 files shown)


karapace/schema_reader.py
```diff
@@ -517,42 +532,53 @@ def handle_msg(self, key: dict, value: Optional[dict]) -> None:
...
         )
+
         if not subject_data:
             raise InvalidReferences(f"Subject not found {reference.subject}.")
 
         # -1 is alias to latest version
         if reference.version == -1:
             reference.version = max(subject_data)
 
-        schema_version: SchemaVersion = subject_data.get(reference.version, None)
```
aiven-anton, line 537: This was obviously wrong, see `None` default in `.get()` call.

karapace/schema_reader.py
```diff
@@ -308,10 +319,17 @@ def handle_messages(self) -> None:
...
+                    if not isinstance(value, dict):
+                        assert value == ""
+                        return None
+                    return value
+
                 value = None
                 if msg.value:
                     try:
-                        value = json_decode(msg.value)
+                        value = parse_value(msg.value)
```
aiven-anton, line 332: This is a balancing act between `test_regression_config_for_inexisting_object_should_not_throw` on the one hand that expects to be able to pass `value=""`, and `self.handle_msg()` that only seem to be accidentally resilient to string values in some special cases.

Perhaps the test should actually be passing `None`? It's hard to tell.

karapace/schema_reader.py
```diff
@@ -244,6 +250,10 @@ def _get_beginning_offset(self) -> int:
     def _is_ready(self) -> bool:
         if self.ready:
             return True
+
+        if self.consumer is None:
+            raise RuntimeError("Ready check was called on reader that isn't running")
```
tvainika, line 255: I think ready check should not raise an exception but instead return `False` and maybe before that log error.

karapace/schema_reader.py
tvainika: While reading the changes I can see that e.g. pushing string or number can be decoded as valid json but same time cause probably odd exceptions from karapace code elsewhere. This change makes it a bit clearer if such unexpected situations happen.

tvainika (changes requested):
ready status raising should be changed IMHO, otherwise good to go. BTW you removed only schema_reader from mypy override list. Can any of other modified files be removed from there?

karapace/schema_reader.py
aiven-anton: This is only called from a single place where it's guarded by `self.consumer is not None`: https://github.com/aiven/karapace/blob/85d8cc73f538d8af11aad0d2bc8681258c4aa1f0/karapace/schema_reader.py#L295-L299.

I would have tended to agree if this was a pubic method, but since it's internal, and due to how things work I'd argue it makes more sense to just fall over if this isn't true. But perhaps in that case we should unify so that we use assert here as well, that properly communicates that it's a pre-condition.

tvainika (approved):

aiven-anton:
> BTW you removed only schema_reader from mypy override list. Can any of other modified files be removed from there?

Unfortunately no, but most of the files touched are already covered, I think it's only `karapace/statsd.py` that's not, and it had a few more errors to deal with.

PR #613, opened 2023-05-11
feature: Implement backup verification command

aiven-anton:

Adds a CLI command to allow verifying checksum integrity of data files in a V3 backup. The command exposes a `--level` argument to allow two different granularity levels of verification.

Running with `--level=file` will only inspect the bytes of data files, making sure they sum up to the expected checksums found in metadata. No parsing of records will happen when running in this mode.

Running with `--level=record` is closer to a dry-run of restoration, and will parse each record and run all the integrity checks we do at restore-time: verifying checksum checkpoints, metadata checksum, as well as record count checks.

Having these separate levels of verification allows running the deeper checks right after producing a backup, and building process around that to dismiss a backup should it fail the verification. This in turn allows greater confidence in running only the shallow file-level check before initiating a restoration.

The suggested operations are:

#### At backup-time

Run backup + deep check.

```sh
$ karapace_schema_backup get [...]
$ karapace_schema_backup verify --level=record [...]  # process should be setup to discard the backup if this fails
```

#### At restoration-time

Run checksum check, then initiate restoration.

```sh
$ karapace_schema_backup verify --level=file [...]  # refuse to restore if this fails.
$ karapace_schema_backup restore [...]

### CODE REVIEW

**Comment by aiven-anton** [comment]
For discussion: perhaps we should also add an option to deeper verification, like verifying the record counts + min and max offsets. The idea would be then that during backup you'd run the deeper verification right after backup process finishes, basically to assert that backup is correctly written to disk, and give confidence that the checksum is valid.

For the restoration process, it then brings enough confidence to check only the checksum matches bytes in data files before initiating the restoration. Because the deeper check were made at backup-time, the checksum verification now brings even stronger guarantees, by proxy.

To clarify, the proposed flow of operations:

#### At backup-time

Run backup + deep check.

```sh
$ karapace_schema_backup get [...]
$ karapace_schema_backup verify --record-integrity [...]  # process should be setup to discard the backup if this fails
```

#### At restoration-time

Run checksum check, then initiate restoration.

```sh
$ karapace_schema_backup verify [...]  # refuse to restore if this fails.
$ karapace_schema_backup restore [...]
```

karapace/backup/backends/v3/backend.py
```diff
@@ -88,6 +89,40 @@ def read(self, path: Path, topic_name: str) -> Iterator[Instruction]:
...
+        for data_file in metadata.data_files:
+            try:
+                for _ in self._read_data_file(
+                    path=(path.parent / data_file.filename),
+                    metadata=metadata,
+                    data_file=data_file,
+                ):
+                    pass
+            except DecodeError:
+                yield False, data_file
```
jjaakola-aiven, line 122: The exception has some extra information, offset, record num etc. This information now gets omitted, would it be any value to print out?

tests/integration/backup/test_v3_backup.py
```diff
@@ -426,3 +425,223 @@ def test_can_inspect_v1(self) -> None:
...
+            check=False,
+            env=no_color_env(),
+        )
+
+        assert cp.returncode == 1
+        assert cp.stderr == b""
+        assert b"Integrity of 2db42756:0.data is not intact!" in cp.stdout
+        assert b"Found checksum mismatches." in cp.stdout
+
+    def test_can_refute_record_integrity(self) -> None:
```
jjaakola-aiven, line 593: The backup file can be  invalid also when it has too few or too many records.
Also the checksum can mismatch on the record batch or the last full checksum.

Maybe the test files could be named to reflect the how the file is broken?

karapace/backup/backends/v3/backend.py
aiven-anton: Fair point, it probably makes sense to be ambitious and have this from the get-go. 👍

aiven-anton: The API layer now yields results as `VerifyFailure | VerifySuccess`, where `VerifyFailure` has an optional exception that is formatted in the command.

karapace/backup/backends/v3/backend.py
```diff
@@ -14,16 +14,33 @@
...
+    data_file: DataFile
+
+
+@default_dataclass
+class VerifyFailure:
+    data_file: DataFile
+    exception: DecodeError | None
+
+
+VerifyResult: TypeAlias = "VerifySuccess | VerifyFailure"
```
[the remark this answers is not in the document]
aiven-anton: In my opinion it's an improvement to encode this on the type level: the successful case cannot have an exception, and so trying to access `result.data_file` on it is a static type error.

tests/integration/backup/test_v3_backup.py
aiven-anton: The filename is now descriptive of its corruption:

```
renamed:    tests/integration/test_data/backup_v3_corrupt/2db42756.metadata -> tests/integration/test_data/backup_v3_corrupt_last_record_bit_flipped_no_checkpoints/2db42756.metadata
renamed:    tests/integration/test_data/backup_v3_corrupt/2db42756:0.data -> tests/integration/test_data/backup_v3_corrupt_last_record_bit_flipped_no_checkpoints/2db42756:0.data
```

jjaakola-aiven (approved):

jjaakola-aiven (approved):

PR #614, opened 2023-05-12
feature: Make schema backup handle empty topics

aiven-anton:

- Adds `start_offset` and `end_offset` to data files in metadata.
- Adds verification of start and end offset in backup reader.
- Adds `record_count` to `Metadata`.
- Start sourcing `partition_count` directly from what's read from Kafka, instead of via `len(data_files)`.
- Test can restore empty topics.


karapace/backup/backends/v3/backend.py
```diff
@@ -168,10 +203,16 @@ def get_checkpoint(
...
+    def update(
+        self,
+        bytes_offset: int,
+        record_offset: int,
+    ) -> None:
         self.records_written_since_checkpoint += 1
         self.records_written += 1
         self.bytes_written_since_checkpoint += bytes_offset
+        self.min_offset = record_offset if self.min_offset is None else min(record_offset, self.min_offset)
+        self.max_offset = record_offset if self.max_offset is None else max(record_offset, self.max_offset)
```
giuseppelillo, line 215: Is it not guaranteed that `record_offset` > `self.min_offset` ?

giuseppelillo (dismissed):

karapace/backup/backends/v3/backend.py
aiven-anton: You're right, if these aren't monotonic it means that something is broken. It probably makes sense to be rigid about that, and raise errors should these expectations not hold.

giuseppelillo (approved):

PR #616, opened 2023-05-15
feature: Implement backup metadata inspect command

aiven-anton:

Implements a command for dumping metadata of a backup file: `karapace_schema_backup inspect --location=<backup location>`.
    
Refactors the internal API into separate functions instead of the shared `SchemaBackup` class. While this was done for the purpose of fitting the inspect command, which does not make sense to require a config file for, the existing methods did not gain much from having a shared class.

Output from invoking the new command on a test backup file:

```json
$ karapace_schema_backup inspect --location=tests/integration/test_data/backup_v3_single_partition/2db42756.metadata
{
  "version": 3,
  "tool_name": "karapace",
  "tool_version": "3.4.6-65-g9259060",
  "started_at": "2023-05-08T09:31:56.238000+00:00",
  "finished_at": "2023-05-08T09:31:56.571000+00:00",
  "topic_name": "2db42756",
  "topic_id": null,
  "partition_count": 1,
  "checksum_algorithm": "xxhash3_64_be",
  "data_files": [
    {
      "filename": "2db42756:0.data",
      "partition": 0,
      "checksum_hex": "f414f504a8e49313",
      "record_count": 2
    }
  ]
}
```

### Why this way?

I first implemented this using the rich library, doing some fancy output formatting, but I scratched that when I realized it's probably much more useful to output this as JSON. Piping through `jq` makes it readable enough for humans.


aiven-anton:
Should test that we can inspect a backup with unknown (future) checksum implementation.

giuseppelillo (approved):

PR #620, opened 2023-05-16
fix: do not override SASL mechanism for user REST proxy

Acuion:
If `sasl_mechanism` already defined in the config - do not override it with `PLAIN`.
Our Kafka installation allows SCRAM authentication only, so current code leads to failed authentication. Preferred mechanism already stated in the config, so let's use it for the user proxy.

# Why this way
`config["sasl_mechanism"]` default value is `None`, so let's check for it. I'm not sure if we can remove this line completely and rely on that undefined SASL mechanism often means `PLAIN`. This is the most safe patch - keeping existing behavior.


tvainika (approved):
Looks good. This is also safe approach. Karapace configuration uses shared `sasl_mechanism` and some other configuration values for both REST proxy and schema registry, so it is better to fallback to `PLAIN` if not configured.

Issue #593, opened 2023-05-01
Got "Servname not supported for ai_socktype" when deploy Karapace Schema Registry


Hongbo-Miao:
Originally asked in [Stack Overflow](https://stackoverflow.com/q/76144191/2000548). Below is a copy. Any guide would be appreciate, thanks! 😃

---

I am trying to deploy [Karapace Schema Registry](https://github.com/aiven/karapace) in a local Kubernetes based on Karapace's [docker-compose.yml](https://github.com/aiven/karapace/blob/1fc2ce73484037599f35d5d5a963a8beb7064c33/container/docker-compose.yml). Here is a copy:

```yaml
version: '3'
services:
  zookeeper:
    image: confluentinc/cp-zookeeper:latest
    ports:
      - "2181:2181"
    environment:
      ZOOKEEPER_CLIENT_PORT: 2181
      ZOOKEEPER_TICK_TIME: 2000

  kafka:
    image: confluentinc/cp-kafka:latest
    depends_on:
      - zookeeper
    ports:
      - "9101:9101"  # JMX
      - "9092:9092"  # Kafka
    environment:
      # Listeners:
      # PLAINTEXT_HOST -> Expose kafka to the host network
      # PLAINTEXT -> Used by kafka for inter broker communication / containers
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:29092,PLAINTEXT_HOST://localhost:9092
      KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: 0
      KAFKA_CONFLUENT_SCHEMA_REGISTRY_URL: http://karapace-registry:8081
      # Metrics:
      KAFKA_JMX_PORT: 9101
      KAFKA_JMX_HOSTNAME: localhost
      # Keep in sync with tests/integration/conftest.py::configure_and_start_kafka
      KAFKA_BROKER_ID: 1
      KAFKA_BROKER_RACK: "local"
      KAFKA_AUTO_CREATE_TOPICS_ENABLE: "false"
      KAFKA_DEFAULT_REPLICATION_FACTOR: 1
      KAFKA_DELETE_TOPIC_ENABLE: "true"
      KAFKA_INTER_BROKER_LISTENER_NAME: "PLAINTEXT"
[...]

PR #625, opened 2023-05-24
Metrics addon

libretto:

    connections_active
    request_size_avg
    request_size_max
    response_size_avg
    response_size_max


requirements/requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.9
```
aiven-anton, line 2: issue: Requirements need to be generated with Python 3.8, we should ideally have a CI check that validates this.

tvainika:
Need to look in more detail, but just initial comments:
1. New features need some documentation, maybe a chapter in README to explain this feature
2. Some configuration to enable/disable this could be useful if not everyone wants this? Maybe could be related to fixing #593 ie. disabling statsd in full.
3. Can you explain rationale for calculating avg and max in karapace side instead of sending values as is, and letting metrics handled calculate e.g. average over given time period instead of something fixed here. Also in my quick look this implementation causes quite lot inaccuracies w.r.t. how time is handled. Sameple windows are resetted based on new data, instead per time frequency, or did I get this wrong?

libretto:
@tvainika I am currently working on expanding the functionality of karapace metrics, which is already in existence. My objective is to incorporate metrics that are similar to what can be found in competitor products. To achieve this, I am utilizing the calculation methods employed by concurrent projects. These projects utilize the org.apache.kafka.common.metrics Java library, which includes static functions for calculating metrics such as Avg, Max, Min, and more. However, I haven't been able to find a Python library that offers similar functionality. Therefore, the metrics library I propose is a straightforward Python implementation based on the methods used in the aforementioned Java library.

While I have not personally explored the reasoning behind calculating average (avg) and minimum (min) metrics instead of sending them to a statsd server, it appears that both the Kafka team and competitors of karapace have found this approach to be rational. Despite the possibility of this method seeming irrational, it is crucial to provide a compatible solution for current customers who are migrating from competitors. Even if certain aspects may appear irrational, it is important to accommodate the preferences and expectations of customers to ensure a seamless transition to our platform.

About config: I will add config there.

libretto:
Created the other branch #631

PR #628, opened 2023-05-26
feat: Add REST producer max request size config

jjaakola-aiven:

The REST producer uses default maximum request size of 1048576 bytes. This is now defined explicitly and configuration file allows overriding the value.


tvainika (dismissed):
Looks good, but please add new configuration to top level README listing all configuration options.

tvainika (approved):

PR #630, opened 2023-05-26
Handle message field compare with known types

### LINKED ISSUE #626: Protobuf does not check imported types

tvainika:

Hard-coded known types do not have message element linked similar to user defined messages.  Handle comparing of those fields without internal errors from exceptions.

Fixes #626

### DIFF
--- a/karapace/protobuf/field_element.py
+++ b/karapace/protobuf/field_element.py
@@ -4,14 +4,17 @@ See LICENSE for details
 """
 # Ported from square/wire:
 # wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/FieldElement.kt
+
+from __future__ import annotations
+
 from karapace.protobuf.compare_result import CompareResult, Modification
 from karapace.protobuf.compare_type_storage import TypeRecordMap
 from karapace.protobuf.field import Field
 from karapace.protobuf.location import Location
 from karapace.protobuf.option_element import OptionElement
 from karapace.protobuf.proto_type import ProtoType
+from karapace.protobuf.type_element import TypeElement
 from karapace.protobuf.utils import append_documentation, append_options
-from typing import List
 
 
 class FieldElement:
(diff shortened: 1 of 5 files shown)


karapace/protobuf/field_element.py
```diff
@@ -142,21 +151,26 @@ def compare_type(
     def compare_message(
         cls, self_type: ProtoType, other_type: ProtoType, result: CompareResult, types: CompareTypes
     ) -> None:
-        from karapace.protobuf.message_element import MessageElement
+        from karapace.protobuf.type_element import TypeElement
```
jjaakola-aiven, line 154: Why imported at this point?

tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -457,3 +593,79 @@ async def test_protobuf_references_latest(registry_async_client: Client) -> None
     )
     assert res.status_code == 200
     assert "id" in res.json()
+
+
+async def test_protobuf_customer_update_references(registry_async_client: Client) -> None:
```
jjaakola-aiven, line 598: `test_protobuf_customer_update_when_having_references` or similar. The test does not update the references, but tests a update case with references.

karapace/protobuf/field_element.py
tvainika: Not sure, fixed now. Maybe there was some thing about circular dependencies...

PR #631, opened 2023-05-28
Extended metrics with config

libretto:

    connections_active
    request_size_avg
    request_size_max
    response_size_avg
    response_size_max
    master_slave_role
    request_error_rate
    response_rate
    response_byte_rate
    latency_max
    latency_avg

My objective is to incorporate metrics that are similar to what can be found in competitor products. To achieve this, I am utilizing the calculation methods from kafka.metrics.Metrics class. 

Karapace competitors use sampled metrics and perform calculations for avg, min,  max on the application side in order to facilitate Pull-based statistics services. This particular model is commonly utilized in platforms like Prometheus, and it is also the preferred choice for our client. However, at present, karapace only supports StatsD. Nevertheless, I anticipate that customers who actively rely on Pull-based statistics will likely require support for such services, and therefore, integrating them into our metrics class would be necessary to meet their needs.


libretto:
@tvainika could You review current version of PR?

PR #632, opened 2023-05-29
feature: Add topic configurations to v3 format

giuseppelillo:

- added `topic_configurations` and `replication_factor` to  Metadata schema
- backed up configurations include all the custom topic configs and a set of default configurations
- topic creation when restoring a backup with format v3 will fail if the topic already exists


karapace/backup/api.py
```diff
@@ -303,15 +293,34 @@ def _write_partition(
...
+    config: Config,
+) -> None:
+    if config["topic_name"] != instruction.name:
+        LOG.warning(
+            "Not creating topic, because the name %r from the config and the name %r from the CLI differ.",
+            config["topic_name"],
+            instruction.name,
+        )
+        return
+    _maybe_create_topic(
```
jjaakola-aiven, line 307: Should we have similar option as for the backup files that overwriting is not allowed by default? The logic for creating the topic will just log `Topic already exist` and allows to proceed.

giuseppelillo: Yes, I'm still finishing this part but that's the idea. For v1 and v2 the behaviour should not be changed (just logging), but for V3 the topic creation will not be allowed if the topic already exists

aiven-anton (commented):
Some early comments, feel free to ignore if they're not relevant because of in-draft :)

karapace/backup/api.py
```diff
@@ -422,6 +436,32 @@ def create_backup(
...
+                        name, val, _, _, _, _ = cv
+                    topic_config[name] = val
+                return topic_config
+
+            topic_config = get_topic_config(admin, topic_name)
+            if replication_factor is None:
+                description: list = admin.describe_topics(topics=[topic_name])
+                if len(description) != 1:
+                    raise RuntimeError(f"Wrong number of topic description retrieved: {description}")
+                replication_factor = len(description[0]["partitions"][0]["replicas"])
```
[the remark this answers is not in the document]
giuseppelillo: This also was moved into a separate function inside `topic_configurations.py`

karapace/backup/api.py
```diff
@@ -174,24 +175,13 @@ def _admin(config: Config) -> KafkaAdminClient:
...
-            )
-            return
-    else:
-        topic = NewTopic(
-            name=name,
-            num_partitions=1,
-            replication_factor=config["replication_factor"],
-            topic_configs={"cleanup.policy": "compact"},
-        )
+def _maybe_create_topic(name: str, *, config: Config, replication_factor: int, topic_configs: Mapping[str, str]) -> None:
```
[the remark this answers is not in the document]
giuseppelillo: Now it's returning a bool to indicate if the creation was successful or not, I would still leave the `maybe` to make it clear that you have to check whether the topic was really created or not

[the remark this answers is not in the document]
giuseppelillo: Although you could infer it from the return type and also from the docstring, so maybe it's not necessary anymore, I'm not sure :)

karapace/backup/api.py
```diff
@@ -303,15 +293,34 @@ def _write_partition(
...
+    instruction: RestoreTopicLegacy,
+    config: Config,
+) -> None:
+    if config["topic_name"] != instruction.name:
+        LOG.warning(
+            "Not creating topic, because the name %r from the config and the name %r from the CLI differ.",
+            config["topic_name"],
+            instruction.name,
+        )
+        return
```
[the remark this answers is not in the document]
giuseppelillo: This is slightly more complex than that because:
- `replication_factor` for V1/V2 is taken from configuration and it's not hard coded, so in order to correctly create a `RestoreTopic` we would need to edit the `BaseBackupReader.read` interface to accept also the config as parameter
- `_handle_restore_topic` also needs to distinguish between V1/V2 and V3 behaviour when dealing with a topic already existing.

To be able to have a single `RestoreTopic` type of instruction probably the cleanest way would be to insert the version inside the `RestoreTopic` dataclass, but again we currently cannot distinguish between V1 and V2 (even though it's not really meaningful in this case).

Overall I think that having different instructions and different handlers is more clean than having a single instruction with more attributes and a single handler that handles different cases with conditional logic. All of this is IMHO of course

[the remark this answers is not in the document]
aiven-anton: Ah, this makes sense 👍

karapace/backup/topic_configurations.py
```diff
@@ -0,0 +1,73 @@
...
+            if name in DEFAULT_CONFIGS or (config_source_filter is not None and config_source in config_source_filter):
+                topic_config[name] = val
+    return topic_config
+
+
+def get_replication_factor(admin: KafkaAdminClient, topic_name: str) -> int:
+    """Get the replication factor for the specified topic by looking for how many replicas are there for one partition."""
+    description: list = admin.describe_topics(topics=[topic_name])
+    assert len(description) == 1
+    return len(description[0]["partitions"][0]["replicas"])
```
aiven-anton, line 73: suggestion: Most of the new code is prepared for multi-partition topics. Would it make sense to just assert `len(description[0]["partitions"]) == 1` here, just to make sure that this function is thought of should we introduce the support later on?

karapace/backup/topic_configurations.py
```diff
@@ -0,0 +1,73 @@
...
+    for cv in config_values:
+        if config_version == 0:
+            name, val, _, _, _ = cv
+            # DescribeConfigs Response v0 does not expose config source: retrieve just a subset of basic configs
+            if name in DEFAULT_CONFIGS:
+                topic_config[name] = val
+        else:
+            name, val, _, config_source, _, _ = cv
+            if name in DEFAULT_CONFIGS or (config_source_filter is not None and config_source in config_source_filter):
+                topic_config[name] = val
```
aiven-anton, line 65: question: Do we have test coverage of both these versions of `DescribeConfig`?

karapace/backup/topic_configurations.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ConfigSource(int, Enum):
+    UNKNOWN = 0
+    TOPIC_CONFIG = 1
+    DYNAMIC_BROKER_CONFIG = 2
+    DYNAMIC_DEFAULT_BROKER_CONFIG = 3
+    STATIC_BROKER_CONFIG = 4
+    DEFAULT_CONFIG = 5
+    DYNAMIC_BROKER_LOGGER_CONFIG = 6
```
aiven-anton, line 20: question: Do we want to record these sources in the backups?

karapace/backup/topic_configurations.py
```diff
@@ -0,0 +1,73 @@
...
+    "min.insync.replicas",
+    "retention.bytes",
+    "retention.ms",
+]
+
+
+def get_topic_configurations(
+    admin: KafkaAdminClient,
+    topic_name: str,
+    config_source_filter: Optional[Iterable[ConfigSource]] = None,
```
aiven-anton, line 34:
```suggestion
    config_source_filter: Container[ConfigSource] = (),
```

nitpick (non-blocking): I think `Container` is slightly more appropriate here, as what we need an object that supports `in`.

nitpick (non-blocking): I think we can simplify by letting this default to the empty tuple, that way we don't need to handle `None`.

karapace/backup/topic_configurations.py
```diff
@@ -0,0 +1,73 @@
...
+    assert len(cfgs) == 1
+    assert len(cfgs[0].resources) == 1
+    err, _, _, _, config_values = cfgs[0].resources[0]
+    if err != 0:
+        raise for_code(err)
+    topic_config = {}
+    for cv in config_values:
+        if config_version == 0:
+            name, val, _, _, _ = cv
+            # DescribeConfigs Response v0 does not expose config source: retrieve just a subset of basic configs
```
aiven-anton, line 59: question: Would it not be a safer default behavior to capture everything?

karapace/backup/topic_configurations.py
```diff
@@ -0,0 +1,73 @@
...
+    DEFAULT_CONFIG = 5
+    DYNAMIC_BROKER_LOGGER_CONFIG = 6
+
+
+DEFAULT_CONFIGS = [
+    "cleanup.policy",
+    "min.insync.replicas",
+    "retention.bytes",
+    "retention.ms",
+]
```
aiven-anton, line 28:
```suggestion
DEFAULT_CONFIGS: Final = (
    "cleanup.policy",
    "min.insync.replicas",
    "retention.bytes",
    "retention.ms",
)
```

nitpick (non-blocking): We can make this final and immutable.

karapace/backup/cli.py
```diff
@@ -40,6 +40,11 @@ def parse_args() -> argparse.Namespace:
         p.add_argument("--poll-timeout", help=PollTimeout.__doc__, type=PollTimeout, default=PollTimeout.default())
 
     parser_get.add_argument("--use-format-v3", action="store_true", help="Use experimental V3 backup format.")
+    parser_get.add_argument(
+        "--replication-factor",
+        help="Use a specified replication factor instead of retrieving it from the current topic state",
+        type=int,
+    )
```
aiven-anton, line 47: issue: We discussed making this a required argument with a special `--replication-factor=auto` mode, because the behavior when not providing a value isn't reliable. It still seems like a good idea to do that, and to document in the help-text that it must be used with caution.

karapace/backup/topic_configurations.py
giuseppelillo: What also Kafka itself does is check [how many replicas there are for the first partition](https://github.com/apache/kafka/blob/3.4.0/core/src/main/scala/kafka/admin/TopicCommand.scala#L344), that number then becomes the replication factor. So this part would remain the same even if we will backup multi-partition topic

aiven-anton: Ah thanks for pointing that out 👍

karapace/backup/cli.py
giuseppelillo: Resolving this because we agreed to have the replication factor mandatory for v3 format.

karapace/backup/topic_configurations.py
giuseppelillo: I don't believe it's really necessary, because all the configs that are backed up will be restored anyway. I would avoid polluting the metadata record with too much information.

aiven-anton: ✅

karapace/backup/topic_configurations.py
giuseppelillo: I added some integration tests for `get_topic_configurations`. Unfortunately they don't cover DescribConfig response v0, because it's only returned by very old brokers

karapace/backup/topic_configurations.py
giuseppelillo: I've changed it, returning all configs in case of DescribeConfigs v0

karapace/backup/topic_configurations.py
aiven-anton: @giuseppelillo Do we need to support it? If all versions of it are already EOL, I'd argue for not including that support.

aiven-anton (approved):
Nice! 👌

karapace/backup/api.py
```diff
@@ -404,6 +430,9 @@ def create_backup(
         if not records are received within that time and the target offset has not
         been reached an exception is raised. Defaults to one minute.
     :param overwrite: the output file if it exists.
+    :param replication_factor: if specified, metadata will contain this value for
```
jjaakola-aiven, line 433: Could remove the note of `instead of the one...` as the code does not attempt to read the replication factor for the topic.
This was discussed in other media and decision was to not to implement replication factor reading from Kafka. During rebalancing or upgrading the number of partitions can be more than configured. Also there is no good API to retrieve the value from Kafka.

jjaakola-aiven, line 433:
```suggestion
    :param replication_factor: if specified, metadata for Version 3 backup will contain this value for
```

karapace/backup/cli.py
```diff
@@ -40,6 +40,12 @@ def parse_args() -> argparse.Namespace:
         p.add_argument("--poll-timeout", help=PollTimeout.__doc__, type=PollTimeout, default=PollTimeout.default())
 
     parser_get.add_argument("--use-format-v3", action="store_true", help="Use experimental V3 backup format.")
+    parser_get.add_argument(
+        "--replication-factor",
+        help="Use a specified replication factor instead of retrieving it from the current topic state. "
```
jjaakola-aiven, line 45: Same here, replication factor is not read from topic state.

jjaakola-aiven (approved):

PR #636, opened 2023-05-31
chore: Add partial stubs for the isodate library

aiven-anton:

Add partial stubs for the isodate library.

This revealed a surprising bug in `PollTimeout`, where we don't handle `parse_duration()` returning `isodate.duration.Duration`.

### Why this way

Increase type coverage.

This is potentially controversial as it only adds partial stubs for the isodate library. This means that the interactions we have currently will now be typed, which is an improvement. But it also means that if someone tries to introduce other valid interactions with the library, that aren't covered by the stubs, it will yield static type checker errors.

The "right" fix when that happens is to expand the partial stubs so that they also cover the new usage, but it's quite hard to make this discoverable, and teach that this is how to fix the issue.

### DIFF
--- a/karapace/backup/poll_timeout.py
+++ b/karapace/backup/poll_timeout.py
@@ -7,8 +7,9 @@ from __future__ import annotations
 from datetime import timedelta
 from functools import cached_property
 from isodate import duration_isoformat, parse_duration
+from typing import Final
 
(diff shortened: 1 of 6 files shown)


tvainika (approved):
Interesting API design that returns either `timedelta` or `Duration`...

PR #643, opened 2023-06-01
fix: Reduce mutability in schema references

### LINKED ISSUE #640: Hashable object is mutated
### What happened?

Discovered while working on static type checking coverage.

`Reference` implements `__hash__()` as a tuple of `(self.name, self.subject, self.version)`:

https://github.com/aiven/karapace/blob/82ab84af387c444c088fd1bbc32e1f1104bb1e03/karapace/schema_references.py#L61

Yet, this object is treated as mutable: 

https://github.com/aiven/karapace/blob/82ab84af387c444c088fd1bbc32e1f1104bb1e03/karapace/schema_reader.py#L558

### What did you expect to happen?

Object should either:

- Not implement `__hash__()`.
- Be immutable. Adding `Final` all of its attributes should not reveal static type checking errors.

aiven-anton:

First off, this fixes the issue that `Reference` objects were mutated while acting as immutable by implementing `__hash__()`. To do that, it breaks up the `Reference` class into two, so that instead of it having two overloaded meanings, the unresolved state is instead represented as its own entity, `LatestVersionReference`. A `LatestVersionReference` can be converted to a `Reference` by providing a version to its `.resolve()` method. This is a typical "make illegal states unrepresentable" kind of refactoring.
    
Related to this, a few `list` and `dict`s are replaced with their immutable abstract counterparts, `Sequence` and `Mapping`. This greatly helps in reasoning about where mutation can happen, and stops me from constantly wondering "will this list be altered when I pass this down this chain of methods".

Fixes #640.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -32,7 +32,7 @@ from karapace.karapace import KarapaceBase
 from karapace.protobuf.exception import ProtobufUnresolvedDependencyException
 from karapace.rapu import HTTPRequest, JSON_CONTENT_TYPE, SERVER_NAME
 from karapace.schema_models import ParsedTypedSchema, SchemaType, SchemaVersion, TypedSchema, ValidatedTypedSchema
-from karapace.schema_references import Reference
+from karapace.schema_references import LatestVersionReference, Reference, reference_from_mapping
 from karapace.schema_registry import KarapaceSchemaRegistry, validate_version
 from karapace.typing import JsonData, JsonObject, ResolvedVersion, SchemaId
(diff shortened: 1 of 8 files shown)


karapace/schema_registry_apis.py
```diff
@@ -79,13 +79,6 @@ class SchemaErrorMessages(Enum):
     REFERENCES_SUPPORT_NOT_IMPLEMENTED = "Schema references are not supported for '{schema_type}' schema type"
 
 
-def references_list(references: dict | None) -> list[Reference] | None:
```
aiven-anton, line 82: unused

tvainika (approved):

PR #645, opened 2023-06-07
feature: Allow backup creation to specify target directory

giuseppelillo:
Instead of using a predefined directory for storing the backup files, `karapace_schema_backup get` should allow to specify the name of the directory where the files will be stored.


aiven-anton:
I'm in the midst of looking at this, I just need to convince myself there's no foot-gun here, I think we should be safe.

Just for posterity, linking the discussion that lead to current behavior: https://github.com/aiven/karapace/pull/584#discussion_r1198894936

aiven-anton (approved):

PR #647, opened 2023-06-08
fix: Broken backup handling of empty topics

aiven-anton:

- Remove an incorrect assertion on number of data files in `Metadata.__post_init__`. Empty partitions will not have data files.
- Add integration test verifying a full backup-restoration cycle for an empty partition.


tests/integration/backup/test_v3_backup.py
```diff
@@ -181,7 +190,10 @@ def test_roundtrip_from_kafka_state(
         check=True,
     )
 
-    # Verify restored topic.
+    # Verify configuration is identical.
+    assert topic_config == get_topic_configurations(admin_client, new_topic.name, {ConfigSource.TOPIC_CONFIG})
```
aiven-anton, line 194: I moved this so that it doesn't happen in-between reading and verifying of records.

PR #648, opened 2023-06-08
feature: Allow passing backup directory to backup commands

aiven-anton:

With this change, we allow passing the same path of the directory created when producing a backup, into restoration and inspection commands. Together with the fix in 5becdb4b, this makes V3 more similar to V2, in the respect of handling given paths.

Also improves test coverage of some of the helper functions in the internal API.

### DIFF
--- a/tests/integration/backup/test_v3_backup.py
+++ b/tests/integration/backup/test_v3_backup.py
@@ -161,6 +161,7 @@ def test_roundtrip_from_kafka_state(
         f"{new_topic.name}:0.data",
     ]
     (metadata_path,) = backup_directory.glob("*.metadata")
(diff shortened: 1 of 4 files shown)


tests/integration/backup/test_v3_backup.py
```diff
@@ -176,7 +177,7 @@ def test_roundtrip_from_kafka_state(
             "--topic",
             new_topic.name,
             "--location",
-            str(metadata_path),
+            str(backup_directory),
```
aiven-anton, line 180: I think it's overkill to introduce a specific integration test for this feature, so I changed only this test to verify passing a parent directory works. The other ones passes the metadata file path.

PR #652, opened 2023-06-09
Karapace metrics

libretto:

    connections_active
    request_size_avg
    request_size_max
    response_size_avg
    response_size_max
    master_slave_role
    request_error_rate
    response_rate
    response_byte_rate
    latency_max
    latency_avg

My objective is to incorporate metrics that are similar to what can be found in competitor products. To achieve this, I am utilizing the calculation methods from kafka.metrics.Metrics class. 

Karapace competitors use sampled metrics and perform calculations for avg, min,  max on the application side in order to facilitate Pull-based statistics services. This particular model is commonly utilized in platforms like Prometheus, and it is also the preferred choice for our client. However, at present, karapace only supports StatsD. Nevertheless, I anticipate that customers who actively rely on Pull-based statistics will likely require support for such services, and therefore, integrating them into our metrics class would be necessary to meet their needs.


libretto:
@tvainika just fixed more issues and added as many as possible metrics similar to competitive products. Also now I utilize native kafka metrics classes.

aiven-anton:
Continuing discussion from [original PR](https://github.com/aiven/karapace/pull/625#issuecomment-1565252743). If possible, in the future I'd ask that you please avoid closing and re-opening separate PRs like this, as it makes it quite hard to follow progress and review. If you need to "start over", please feel free to force push to your feature branch instead.

---

> Karapace competitors use sampled metrics and perform calculations for avg, min, max on the application side in order to facilitate Pull-based statistics services. This particular model is commonly utilized in platforms like Prometheus [...]

This does not sound accurate to me. In my experience when using Prometheus, it's much more common to avoid calculations on the application side, and expose metrics in their raw forms. Averages, and other aggregations are then produced by posing PromQL queries to the Prometheus service.

It seems reasonable to take that route here as well, and for instance rather expose `sent_response_bytes`, `received_request_bytes`. Together with a request count, that should be everything required to reliable calculate aggregations in Prometheus or similar.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,169 @@
...
+connections-active - The number of active HTTP(S) connections to server.
+                     Data collected inside aiohttp request handler.
+
+Copyright (c) 2023 Aiven Ltd
+See LICENSE for details
+"""
+from datetime import datetime
+from kafka.metrics import MetricName, Metrics
+from kafka.metrics.measurable_stat import AbstractMeasurableStat
+from kafka.metrics.stats import Avg, Max, Rate, Total
```
aiven-anton, line 14: issue: I don't think this is meant to be a public API, as there is no trace of this in documentation. These should probably be seen as internal implementation details within kafka-python, and I don't think we should use them.

We are having internal discussions about replacing kafka-python, which is another reason not to tie us closer to it.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,169 @@
...
+    # pylint: disable=unused-argument
+    def measure(self, config: object, now: int) -> float:
+        return self.value
+
+    def record(self, config: object, value: float, time_ms: int) -> None:
+        self.value = value
+
+
+class Singleton(type):
+    _instances: Dict["Singleton", "Singleton"] = {}
```
aiven-anton, line 42:
```suggestion
    _instances: Dict["type[Singleton]", "Singleton"] = {}
```

issue: I don't believe we're mapping instance to instance here, but rather class to instance.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,169 @@
...
+        return cls._instances[cls]
+
+
+class KarapaceMetrics(metaclass=Singleton):
+    def __init__(self) -> None:
+        self.active: Optional[object] = None
+        self.stats_client: Optional[StatsClient] = None
+        self.is_ready = False
+        self.metrics = Metrics()
+        self.event = threading.Event()
```
aiven-anton, line 57:
```suggestion
        self.stop_event = threading.Event()
```

suggestion: This can have more descriptive name.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,169 @@
...
+    def report(self) -> None:
+        if not self.active:
+            return
+        if isinstance(self.stats_client, StatsClient):
+            for metric_name in self.metrics.metrics:
+                value = self.metrics.metrics[metric_name].value()
+                self.stats_client.gauge(metric_name.name, value)
+
+    def schedule(self) -> None:
+        self.report()
```
aiven-anton, line 155:
```suggestion
    def schedule(self) -> None:
        self.report()
```

issue: What's the point of this method? Let's replace it with invocations to `.report()` instead?

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,169 @@
...
+    def error(self) -> None:
+        if not self.active:
+            return
+        timestamp = int(datetime.utcnow().timestamp() * 1e3)
+        self.metrics.get_sensor("request-error-rate").record(1, timestamp)
+
+    def report(self) -> None:
+        if not self.active:
+            return
+        if isinstance(self.stats_client, StatsClient):
```
aiven-anton, line 149: issue: I think this should be a `RuntimeError` as it shouldn't be possible for `self.stats_client` not be a `StatsClient` when `self.active` is `True`.

karapace/karapacemetrics.py
libretto: In my previous pull request, I had implemented most of the metrics using my own approach similar to kafka java classes. However, I discovered that these classes in kafka-python were essentially ports of Kafka Java classes, which are also utilized by a competing product. Considering the importance of compatibility, I made the decision to create the metrics using the same approach.

karapace/karapacemetrics.py
libretto: > Continuing discussion from [original PR](https://github.com/aiven/karapace/pull/625#issuecomment-1565252743). If possible, in the future I'd ask that you please avoid closing and re-opening separate PRs like this, as it makes it quite hard to follow progress and review. If you need to "start over", please feel free to force push to your feature branch instead.
> 
> > Karapace competitors use sampled metrics and perform calculations for avg, min, max on the application side in order to facilitate Pull-based statistics services. This particular model is commonly utilized in platforms like Prometheus [...]
> 
> This does not sound accurate to me. In my experience when using Prometheus, it's much more common to avoid calculations on the application side, and expose metrics in their raw forms. Averages, and other aggregations are then produced by posing PromQL queries to the Prometheus service.
> 
> It seems reasonable to take that route here as well, and for instance rather expose `sent_response_bytes`, `received_request_bytes`. Together with a request count, that should be everything required to reliable calculate aggregations in Prometheus or similar.

My client requires support for metrics that are already supported in competitor projects. Additionally, they anticipate the option to integrate with Prometheus in the future. Consequently, I have implemented the required support accordingly. The test results using Graphite, which I used for testing purposes, show the expected data output. For Prometheus integration, we have two options: utilizing the statsd-prometheus module or developing our own converter to send aggregated data to Prometheus. I believe Prometheus has the capability to handle aggregated data, allowing us to aggregate the metrics as required.

libretto:
> Continuing discussion from [original PR](https://github.com/aiven/karapace/pull/625#issuecomment-1565252743). If possible, in the future I'd ask that you please avoid closing and re-opening separate PRs like this, as it makes it quite hard to follow progress and review. If you need to "start over", please feel free to force push to your feature branch instead.
> 
> > Karapace competitors use sampled metrics and perform calculations for avg, min, max on the application side in order to facilitate Pull-based statistics services. This particular model is commonly utilized in platforms like Prometheus [...]
> 
> This does not sound accurate to me. In my experience when using Prometheus, it's much more common to avoid calculations on the application side, and expose metrics in their raw forms. Averages, and other aggregations are then produced by posing PromQL queries to the Prometheus service.
> 
> It seems reasonable to take that route here as well, and for instance rather expose `sent_response_bytes`, `received_request_bytes`. Together with a request count, that should be everything required to reliable calculate aggregations in Prometheus or similar.

My client requires support for metrics that are already supported in competitor projects. Additionally, they anticipate the option to integrate with Prometheus in the future. Consequently, I have implemented the required support accordingly. The test results using Graphite, which I used for testing purposes, show the expected data output. For Prometheus integration, we have two options: utilizing the statsd-prometheus module or developing our own converter to send aggregated data to Prometheus. I believe Prometheus has the capability to handle aggregated data, allowing us to aggregate the metrics as required.

karapace/karapacemetrics.py
libretto: done

karapace/karapacemetrics.py
libretto: done

karapace/karapacemetrics.py
libretto: done

karapace/karapacemetrics.py
libretto: fixed

karapace/karapacemetrics.py
libretto: but my way worked and passed all pypy checks

karapace/karapacemetrics.py
aiven-anton: Thanks for explaining your reasoning, however, this in my opinion is a blocking issue. We should not add these new dependencies on private APIs.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,168 @@
...
+            for metric_name in self.metrics.metrics:
+                value = self.metrics.metrics[metric_name].value()
+                self.stats_client.gauge(metric_name.name, value)
+        else:
+            raise RuntimeError
```
aiven-anton, line 154: issue: Add a message to this exception.

karapace/karapacemetrics.py
aiven-anton: Hmm right, that's another issue because the file does not have `from __future__ import annotations`, but I was also off in my comment.

`cls` is equal to `KarapaceMetrics` here, which is an instance of `Singleton` because that is its metaclass (`type(KarapaceMetrics) is Singleton`). So the key type is actually correct, however the value type isn't. You can verify this by checking `assert isinstance(instance, Singleton)` in `__call__`.

I'd suggest updating the return type of `__call__`, as well as the value type of `_instances` to `object`.

karapace/karapacemetrics.py
libretto: Additionally, since kafka-python is also released under the Apache License, the licensing aspect should not pose a blocking issue either.

karapace/karapacemetrics.py
libretto: For me also wors following construction:
    _instances: Dict[type("Singleton"), "Singleton"] = {}

libretto: Construction     _instances: dict[type[object], Singleton] = {} works

aiven-anton: @libretto That doesn't make sense though, the key is overly broad and the value is incorrect? I was proposing `dict[Singleton, object]`.

karapace/karapacemetrics.py
aiven-anton: This is the API reference of kafka-python: https://kafka-python.readthedocs.io/en/master/apidoc/modules.html

Regardless of what's exposed in Java, the APIs used here are not documented in kafka-python, and hence, are to be considered private implementation details.

aiven-anton:
> My client requires support for metrics that are already supported in competitor projects. Additionally, they anticipate the option to integrate with Prometheus in the future. Consequently, I have implemented the required support accordingly. The test results using Graphite, which I used for testing purposes, show the expected data output. For Prometheus integration, we have two options: utilizing the statsd-prometheus module or developing our own converter to send aggregated data to Prometheus. I believe Prometheus has the capability to handle aggregated data, allowing us to aggregate the metrics as required.

@libretto Aggregated metrics are out of scope for Karapace, we are not going to accept PRs that implement that since we do not want to carry that maintenance burden. If you need 1-to-1 metrics compatibility, I'd suggest implementing this in a separate application where you can implement the exact aggregations you need.

That said, we would consider a PR that adds a Prometheus endpoint for exposing metrics to Karapace. That should be probably be implemented in a separate PR though.

aiven-anton (changes requested):
See prior comments.

karapace/karapacemetrics.py
libretto: I need a dict where keys are exactly represented, classes... But seems Your offer also looks like dict[instance,instance]. So let me try to get rid of dict at all.

libretto:
> > My client requires support for metrics that are already supported in competitor projects. Additionally, they anticipate the option to integrate with Prometheus in the future. Consequently, I have implemented the required support accordingly. The test results using Graphite, which I used for testing purposes, show the expected data output. For Prometheus integration, we have two options: utilizing the statsd-prometheus module or developing our own converter to send aggregated data to Prometheus. I believe Prometheus has the capability to handle aggregated data, allowing us to aggregate the metrics as required.
> 
> @libretto Aggregated metrics are out of scope for Karapace, we are not going to accept PRs that implement that since we do not want to carry that maintenance burden. If you need 1-to-1 metrics compatibility, I'd suggest implementing this in a separate application where you can implement the exact aggregations you need.
> 
> That said, we would consider a PR that adds to Karapace a Prometheus endpoint for exposing metrics. That should probably be implemented in a separate PR though.

@aiven-anton Are You in favor of allowing record calls to the KarapaceMetrics() class, but with the requirement of removing aggregation from the class? Does this mean you are not concerned about the impact on performance caused by the 10-11 stats request calls per karapace request to StatsD or Prometheus?

karapace/karapacemetrics.py
aiven-anton: > So let me try to get rid of dict at all.

No need to do this.

> I need a dict where keys are exactly represented, classes... But seems Your offer also looks like dict[instance,instance].

No, as stated earlier, I was off in my first comment. You can verify that `assert isinstance(cls, Singleton)` passes inside the method. (This is because `type(KarapaceMetrics) == Singleton`).

aiven-anton:
> Are You in favor of allowing record calls to the KarapaceMetrics() class, but with the requirement of removing aggregation from the class? Does this mean You are not concerned about the impact on performance caused by the 10-11 stats request calls per karapace request to StatsD or Prometheus?

What I am saying is that we should not calculate averages, maximums, and rates, and instead only expose raw metrics. This means I think the proposed metrics should be changed like this:

- Replace `request_size_avg` and `request_size_max` with something like `request_bytes_received`. 
- Replace `response_byte_rate`, `response_size_avg` and `response_size_max` with something like `response_bytes_sent`.
- Replace `request_error_rate` with a counter.
- Replace `latency_max` and `latency_avg` with something like `latency_total`.
- Introduce a request counter.

libretto:
@aiven-anton 
Hello, I've made changes to the code to only send raw statistics, if possible, to the Stats Server. Currently, the data is being sent to the statsd server. Could you please review the code and let me know if this is what you were referring to?

Additionally, I'm planning to add support for Prometheus. The Prometheus client acts as an HTTP server that provides an endpoint for obtaining statistics periodically. Is this acceptable to you?

Furthermore, I'm considering a better approach. One option is to inherit from the StatsClient or create an AbstractStatsClient class with interfaces. This way, both the StatsClient and PrometheusStatsClient can inherit from the same class, and my KarapaceMetrics class can utilize it as an interface. What are your thoughts on this?

libretto (commented):
Please check

libretto:
@aiven-anton I kindly request that you review my previous comment and provide an answer regarding the proposed Prometheus solution. Specifically, I would like to know if I can utilize the pull model, where the server is inside karapace, or if I need to use the push model and have clients request the use of Pushgateway between karapace and Prometheus.

aiven-anton:
Sorry for the delay, I appreciate the update here 👍

> Additionally, I'm planning to add support for Prometheus. The Prometheus client acts as an HTTP server that provides an endpoint for obtaining statistics periodically. Is this acceptable to You?

Yes, but I strongly encourage doing this in a follow-up PR to keep the scope manageable and reviewable. And you're exactly right, with the Prometheus model we would not use a client, but expose a special metrics endpoint (probably on a separate port). But definitely I think that this is right approach, the Prometheus implementation should be a server that Prometheus "scrapes" from rather than the Pushgateway you mention as alternative.

> Furthermore, I'm considering a better approach. One option is to inherit from the StatsClient or create an AbstractStatsClient class with interfaces. This way, both the StatsClient and PrometheusStatsClient can inherit from the same class, and my KarapaceMetrics class can utilize it as an interface. What are Your thoughts on this?

Yes, it sounds reasonable to have some common interface, and define separate pluggable backends for Statsd and later Prometheus. I'd also encourage deferring the introduction of this until you work on the Prometheus implementation, as doing it beforehand is likely to end up with an abstraction that doesn't meet all requirements.

libretto:
@aiven-anton can You check the current PR and approve it yet?

libretto:
@aiven-anton Hi, could You review this code and merge it with the main branch?

libretto:
> See prior comments.

could You review?

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,133 @@
...
+
+        schedule.every(10).seconds.do(self.connections)
+        self.worker_thread.start()
+
+    def request(self, size: int) -> None:
+        if not self.active:
+            return
+        if not isinstance(self.stats_client, StatsClient):
+            raise RuntimeError("no StatsClient available")
+        self.stats_client.gauge("request-size", size)
```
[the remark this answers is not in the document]
libretto: I'm unclear about the necessity of using "request_size_total" when our aim is to create a graph depicting request size over time. As far as I can see, there isn't a metric that requires a counter. As I know both Prometheus and StatsD have a similar gauge metric type.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def response(self, size: int) -> None:
+        if not self.active:
+            return
+        if not isinstance(self.stats_client, StatsClient):
+            raise RuntimeError("no StatsClient available")
+        self.stats_client.gauge("response-size", size)
```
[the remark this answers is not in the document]
libretto: why "Total"?

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,133 @@
...
+            raise RuntimeError("no StatsClient available")
+        self.stats_client.gauge("master-slave-role", latency_ms)
+
+    def error(self) -> None:
+        if not self.active:
+            return
+        if not isinstance(self.stats_client, StatsClient):
+            raise RuntimeError("no StatsClient available")
+        self.error_count += 1
+        self.stats_client.gauge("error", self.error_count)
```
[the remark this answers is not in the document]
libretto: Our goal is to create a graph showing the request error rate, which represents the number of error requests per second. However, considering your previous advice that we shouldn't perform calculations or aggregation on the karapace side, I'm currently uncertain about whether to use a gauge or counter to achieve the desired rate graph as the output.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,133 @@
...
+        if not isinstance(self.stats_client, StatsClient):
+            raise RuntimeError("no StatsClient available")
+        psutil.Process(os.getpid()).connections()
+        connections = 0
+        for conn in psutil.net_connections(kind="tcp"):
+            if not conn.laddr:
+                continue
+            if conn.laddr[0] == self.app_host and conn.laddr[1] == self.app_port and conn.status == "ESTABLISHED":
+                connections += 1
+        self.stats_client.gauge("connections-active", connections)
```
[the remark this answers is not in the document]
libretto: ok

README.rst
```diff
@@ -461,6 +461,15 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
    * - ``master_election_strategy``
      - ``lowest``
      - Decides on what basis the Karapace cluster master is chosen (only relevant in a multi node setup)
+   * - ``metrics_mode``
+     - ``statsd``
+     - Statistics server mode. For karapace supports ststsd server
+   * - ``statsd_uri``
+     - ``127.0.0.1:8125``
+     - Host:Port of statsd server
+   * - ``metrics_extended``
```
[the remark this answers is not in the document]
libretto: It provides the option to deactivate metrics routines when extended metrics are unnecessary.

README.rst
```diff
@@ -461,6 +461,15 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
...
+     - ``true``
+     - Enable extended metrics. Extended metrics: connections_active, request_size_avg, request_size_max, response_size_avg, response_size_max
```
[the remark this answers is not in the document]
libretto: ok

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,133 @@
...
+        app_host = config.get("host")
+        app_port = config.get("port")
+        if app_host and app_port:
+            self.app_host = app_host
+            self.app_port = app_port
+        else:
+            raise RuntimeError("No application host or port defined in application")
+
+        schedule.every(10).seconds.do(self.connections)
+        self.worker_thread.start()
```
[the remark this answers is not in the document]
libretto: The thread job operates at a frequency of every 10 seconds, which is a relatively resource-efficient approach.

karapace/karapacemetrics.py
```diff
@@ -0,0 +1,119 @@
+"""
```
aiven-anton, line 1: issue: On the naming of this module, we're already in the karapace namespace, let's simply name the module `metrics`? (Resulting name would be `karapace.metrics` instead of repeating `karapace.karapacemetrics`).

karapace/karapacemetrics.py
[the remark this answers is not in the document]
aiven-anton: @libretto Could you shed some further light on this question?

> Is the threading needed only for the reading connection count?

karapace/karapacemetrics.py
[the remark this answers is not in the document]
aiven-anton: @libretto I think gauge simply doesn't make sense here?

> our aim is to create a graph depicting request size over time

The way to do that with prometheus is to graph `request_size_total / request_count`, I'm unsure what makes sense with StatsD.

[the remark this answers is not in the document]
libretto: In order to calculate "request_size_total / request_count," we would need to store "request_count" and perform the calculation for "request_size_total" on the karapace side. However, I noticed your comment here: [https://github.com/Aiven-Open/karapace/pull/652#issuecomment-1584361484] where you mentioned that this approach might not be advisable. Consequently, I modified the code in this manner because both Statsd and Prometheus can visualize this data without the need for internal calculations. Perhaps this approach may be less efficient than mine, but at least it avoids any calculations within karapace.

karapace/karapacemetrics.py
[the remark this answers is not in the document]
libretto: For instance, if Karapace is handling 100 requests per second, the function that retrieves the connection count would be called with every request. In my solution, we acquire this data every 10 seconds, which means significantly fewer calls (about 1000 times fewer). While this approach might not be as precise, it's effective enough for our purposes.

karapace/karapacemetrics.py
libretto: fixed

karapace/karapacemetrics.py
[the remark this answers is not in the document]
libretto: @aiven-anton @jjaakola-aiven could you check?

aiven-anton:
This is superseded by https://github.com/Aiven-Open/karapace/pull/711.

PR #654, opened 2023-06-09
feature: Allow skipping topic creation for backup restoration

giuseppelillo:
Introduce `--skip-topic-creation` flag for backup restoration command. With this flag it's possible to restore the backed up data into an already existing topic.


karapace/backup/api.py
```diff
@@ -424,10 +425,12 @@ def restore_backup(
...
         for instruction in backend.read(backup_location, topic_name):
             if isinstance(instruction, RestoreTopicLegacy):
-                _handle_restore_topic_legacy(instruction, config)
+                if not skip_topic_creation:
+                    _handle_restore_topic_legacy(instruction, config)
                 producer = stack.enter_context(_producer(config, instruction.topic_name))
             elif isinstance(instruction, RestoreTopic):
-                _handle_restore_topic(instruction, config)
+                if not skip_topic_creation:
+                    _handle_restore_topic(instruction, config)
```
aiven-anton, line 433: suggestion (non-blocking): This increases cyclic complexity, and is sort of orthogonal to what this function should do (delegate to the correct handler), and might be preferable to do as early return checks within the handler functions instead.

karapace/backup/cli.py
```diff
@@ -34,6 +34,11 @@ def parse_args() -> argparse.Namespace:
     for p in (parser_get, parser_restore, parser_export_anonymized_avro_schemas):
         p.add_argument("--config", help="Configuration file path", required=True)
         p.add_argument("--topic", help="Kafka topic name to be used", required=False)
+        p.add_argument(
+            "--skip-topic-creation",
+            action="store_true",
+            help="Skip topic creation, restoring only the data into an already existing topic.",
```
aiven-anton, line 40:
```suggestion
            help="Skip topic creation, restoring only the data into an already existing topic.",
            default=False,
```

issue: Without this I believe the annotation in `restore_backup` is not correct (`bool` vs `bool | None`).

tests/integration/backup/test_v3_backup.py
```diff
@@ -424,6 +424,105 @@ def test_roundtrip_from_file(
     assert data_file.read_bytes() == new_data_file.read_bytes()
 
 
+def test_roundtrip_from_file_skipping_topic_creation(
```
aiven-anton, line 427: question: Should we also test behavior when the topic doesn't exist? It seems likely we'll run into that.

karapace/backup/cli.py
giuseppelillo: On the docs it says that when the action is `store_true`, the default is False. Anyway i've added it an explicit default=False so it's more clear.

tests/integration/backup/test_v3_backup.py
```diff
@@ -424,6 +427,160 @@ def test_roundtrip_from_file(
...
+
+    # Verify data files are identical.
+    (new_data_file,) = new_metadata_path.parent.glob("*.data")
+    assert data_file.read_bytes() == new_data_file.read_bytes()
+
+
+def test_backup_restoration_fails_when_topic_does_not_exist_and_skip_creation_is_true(
+    admin_client: KafkaAdminClient,
+    kafka_servers: KafkaServers,
+    # pp,
```
aiven-anton, line 532:
```suggestion
```

tests/integration/backup/test_v3_backup.py
aiven-anton: Nice, I'm happy with the test as it locks the behavior in 👍 It would be nice to have a more solid way to test this in the future, distinguishing between other reasons for timing out.

PR #655, opened 2023-06-12
Improve typing and enable mypy checks for many modules

tvainika:

Add type hints and some minor refactorings for better static checking with mypy.

# Why this way

This contains only easy wins to enable more mypy check coverage.  See individual commits for more details if needed.

### DIFF
--- a/karapace/protobuf/rpc_element.py
+++ b/karapace/protobuf/rpc_element.py
@@ -5,6 +5,8 @@ See LICENSE for details
 # Ported from square/wire:
 # wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/RpcElement.kt
 
+from __future__ import annotations
+
 from karapace.protobuf.location import Location
 from karapace.protobuf.utils import append_documentation, append_indented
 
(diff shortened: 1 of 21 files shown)


karapace/protobuf/rpc_element.py
```diff
@@ -19,7 +21,7 @@ def __init__(
         response_type: str = "",
         request_streaming: bool = False,
         response_streaming: bool = False,
-        options: list = None,
+        options: list | None = None,
```
aiven-anton, line 24: question (non-blocking): Can we eliminate the implicit `Any` here, i.e provide a list item type? (I usually go about that by changing to `list[object]` to discover expectations)

tests/unit/protobuf/test_proto_file_element.py
```diff
@@ -18,7 +18,7 @@
 
 import copy
 
-location: Location = Location.get("some/folder", "file.proto")
+location: Location = Location("some/folder", "file.proto")
```
aiven-anton, line 21: nitpick (non-blocking): Probably unnecessary type hint? (same in some other modules in same package).

aiven-anton (approved):
Nice! 👍

PR #656, opened 2023-06-13
fix: set aiohttp Application client max size according with producer config

jjaakola-aiven:

When setting the producer max request size larger than default the aiohttp Application client max size must be set also larger.

# Why this way

When REST Proxy is enabled the aiohttp client max size selection is adaptive if `aiohttp_client_max_size` configuration is not set. In other cases the default or explicitly set configuration value is used.


README.rst
```diff
@@ -308,6 +308,9 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
    * - Parameter
      - Default Value
      - Description
+   * - ``aiohttp_client_max_size``
```
aiven-anton, line 311: question: Shouldn't the usage of aiohttp be kept as an internal implementation detail, that we can change without changing names of configuration parameters?

jjaakola-aiven: `http_request_max_size`.

karapace/config.py
```diff
@@ -186,6 +187,23 @@ def set_config_defaults(config: ConfigDefaults) -> Config:
...
+        # Always add the aiohttp default client max size as the headroom above the producer max request size.
+        # The input JSON size for REST Proxy is not easy to estimate, lot of small records in single request has
+        # a lot of overhead due to JSON structure.
+        new_config["http_request_max_size"] = new_config["producer_max_request_size"] + DEFAULT_AIOHTTP_CLIENT_MAX_SIZE
+    elif new_config["http_request_max_size"] is None:
+        # Set the default aiohttp client max size
+        new_config["http_request_max_size"] = DEFAULT_AIOHTTP_CLIENT_MAX_SIZE
+    else:
+        # Use the aiohttp client max size from the configuration without altering.
+        pass
```
aiven-anton, line 205:
```suggestion
```

suggestion (non-blocking): Skip unnecessary branch.

karapace/constants.py
```diff
@@ -10,6 +10,8 @@
 API_VERSION_AUTO_TIMEOUT_MS: Final = 30000
 TOPIC_CREATION_TIMEOUT_MS: Final = 20000
 DEFAULT_SCHEMA_TOPIC: Final = "_schemas"
+DEFAULT_PRODUCER_MAX_REQUEST: Final = 1048576
+DEFAULT_AIOHTTP_CLIENT_MAX_SIZE = 1024**2
```
aiven-anton, line 14:
```suggestion
DEFAULT_AIOHTTP_CLIENT_MAX_SIZE: Final = 1024**2
```

issue: I assume this one should've also received `Final`.

aiven-anton (commented):
Change looks good, left a nitpick and a minor comment.

tvainika (changes requested):
It looks like after iterating with config name you got things broken as `aiohttp.web.Application` is now reading `aiohttp_xxx` config value which was renamed to `http_xxx`.

However I'd like to question whole approach of adding new configuration parameter. What if simply add `aiohttp.web.Application` constructor parameter that is always large enough (based on `producer_max_request_size` setting)? No new config parameter, no need for more complex configuration but simply works?

karapace/constants.py
tvainika: It is a bit confusing that those numerical values are equal to each other, but simply represented differently, one as simple value and other as exponential notation.

karapace/config.py
jjaakola-aiven: I can remove, the reason why I added `pass` is to be explicit with the use of the value in the configuration.

PR #657, opened 2023-06-14
feature: Return specific exit code for failed data restoration

giuseppelillo:
`karapace_schema_backup restore` returns exit code 3 if an error was raised while restoring data. This is helpful to be able to distinguish between a generic error and a data restoration error, allowing clients to correctly retry the restoration command.

### DIFF
--- a/karapace/backup/api.py
+++ b/karapace/backup/api.py
@@ -11,7 +11,14 @@ from .backends.v3.constants import V3_MARKER
 from .backends.v3.schema import ChecksumAlgorithm
 from .backends.writer import BackupWriter, StdOut
 from .encoders import encode_key, encode_value
-from .errors import BackupError, BackupTopicAlreadyExists, EmptyPartition, PartitionCountError, StaleConsumerError
+from .errors import (
+    BackupDataRestorationError,
+    BackupError,
+    BackupTopicAlreadyExists,
+    EmptyPartition,
+    PartitionCountError,
+    StaleConsumerError,
+)
 from .poll_timeout import PollTimeout
 from .topic_configurations import ConfigSource, get_topic_configurations
 from enum import Enum
@@ -20,6 +27,7 @@ from kafka import KafkaConsumer, KafkaProducer
 from kafka.admin import KafkaAdminClient, NewTopic
 from kafka.consumer.fetcher import ConsumerRecord
 from kafka.errors import KafkaError, TopicAlreadyExistsError
+from kafka.future import Future
 from kafka.structs import PartitionMetadata, TopicPartition
 from karapace import constants
 from karapace.backup.backends.v1 import SchemaBackupV1Reader
@@ -256,6 +264,16 @@ def _consumer(config: Config, topic: str) -> Iterator[KafkaConsumer]:
 
 
 @contextlib.contextmanager
+def _enable_producer_callback_errors() -> Iterator[None]:
+    global_value = Future.error_on_callbacks
+    Future.error_on_callbacks = True
+    try:
+        yield None
+    finally:
+        Future.error_on_callbacks = global_value
+
+
+@contextlib.contextmanager
 def _producer(config: Config, topic: str) -> Iterator[KafkaProducer]:
     """Creates an automatically closing Kafka producer client.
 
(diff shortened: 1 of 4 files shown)


karapace/backup/api.py
```diff
@@ -384,14 +391,17 @@ def _handle_producer_send(
...
-    ).add_errback(_raise_backup_error)
+    try:
+        producer.send(
+            instruction.topic_name,
+            key=instruction.key,
+            value=instruction.value,
+            partition=instruction.partition_index,
+            headers=[(key.decode() if key is not None else None, value) for key, value in instruction.headers],
+            timestamp_ms=instruction.timestamp,
+        ).add_errback(_raise_backup_error)
```
jjaakola-aiven, line 402: The error raised in the callback is generic backup error. Should it be also treated as data restoration error?

tests/integration/backup/test_v3_backup.py
```diff
@@ -334,6 +334,48 @@ def test_errors_when_omitting_replication_factor(config_file: Path) -> None:
     assert "the following arguments are required: --replication-factor" in exc_info.value.stderr.decode()
 
 
+def test_error_data(
```
aiven-anton, line 337:
```suggestion
def test_exits_with_return_code_3_for_data_restoration_error(
```

karapace/backup/api.py
```diff
@@ -384,14 +391,17 @@ def _handle_producer_send(
...
+    except Exception as ex:
+        raise BackupDataRestorationError() from ex
```
aiven-anton, line 404:
```suggestion
    except Exception as ex:
        raise BackupDataRestorationError() from ex
```

issue: I'm pretty adamantly against catching wild `Exception` without doing anything than a re-raise (catching a `MemoryError` and directly trying to instantiate an object is probably good way to hide the original exception). Do we really need to do that? I think the only thing we should catch here is `BackupError`, i.e. whatever is raised by `_raise_backup_error`.

karapace/backup/api.py
aiven-anton: Instead of this catch and re-raise, perhaps we could just change the exception in `_raise_backup_error`?

tests/integration/backup/test_v3_backup.py
```diff
@@ -334,6 +334,48 @@ def test_errors_when_omitting_replication_factor(config_file: Path) -> None:
...
+                    "--topic",
+                    topic_name,
+                    "--location",
+                    str(metadata_path),
+                ],
+                capture_output=True,
+                check=True,
+            )
+            print(er.value.stderr)
+            assert er.value.returncode == 3
```
aiven-anton, line 376:
```suggestion
        print(er.value.stderr)
        assert er.value.returncode == 3
```

issue: This lines will never be reached, they must be outside the managed block.

tests/integration/backup/test_v3_backup.py
```diff
@@ -334,6 +334,48 @@ def test_errors_when_omitting_replication_factor(config_file: Path) -> None:
...
+                    str(config_file),
+                    "--topic",
+                    topic_name,
+                    "--location",
+                    str(metadata_path),
+                ],
+                capture_output=True,
+                check=True,
+            )
+            print(er.value.stderr)
```
aiven-anton, line 375:
```suggestion
```

suggestion: I would drop this as it will mostly add noise to logs.

aiven-anton (changes requested):

karapace/backup/api.py
giuseppelillo: `_raise_backup_error` now raises `BackupDataRestorationError` directly.

karapace/backup/api.py
```diff
@@ -384,16 +392,30 @@ def _handle_producer_send(
...
+            partition=instruction.partition_index,
+            headers=[(key.decode() if key is not None else None, value) for key, value in instruction.headers],
+            timestamp_ms=instruction.timestamp,
+        ).add_errback(_raise_backup_error)
+    except (KafkaError, AssertionError) as ex:
+        raise BackupDataRestorationError("Error while calling send on restoring messages") from ex
+
+
+@contextlib.contextmanager
+def _enable_producer_callback_errors():
```
aiven-anton, line 409: issue: Missing type hints.

aiven-anton, line 409: suggestion: I'd also consider making this part of `_producer()` in this module, I think it's safe to enable it for all of backups.

karapace/backup/api.py
```diff
@@ -384,16 +392,30 @@ def _handle_producer_send(
...
+            partition=instruction.partition_index,
+            headers=[(key.decode() if key is not None else None, value) for key, value in instruction.headers],
+            timestamp_ms=instruction.timestamp,
+        ).add_errback(_raise_backup_error)
+    except (KafkaError, AssertionError) as ex:
+        raise BackupDataRestorationError("Error while calling send on restoring messages") from ex
+
+
+@contextlib.contextmanager
+def _enable_producer_callback_errors() -> Iterator[Callable[..., None]]:
```
aiven-anton, line 409:
```suggestion
def _enable_producer_callback_errors() -> Iterator[None]:
```

We're not yielding anything from this function, just `None` :)

aiven-anton (approved):
LGTM, I'll leave it to @jjaakola-aiven to merge.

aiven-anton (approved):

PR #660, opened 2023-06-15
fix: raise producer exception

giuseppelillo:


karapace/backup/api.py
```diff
@@ -390,14 +378,20 @@ def _handle_restore_topic(
...
+    global producer_exception  # pylint: disable=global-statement
+    producer_exception = exception
 
 
 def _handle_producer_send(
     instruction: ProducerSend,
     producer: KafkaProducer,
 ) -> None:
+    if producer_exception is not None:
+        raise BackupDataRestorationError("Error while producing restored messages") from producer_exception
```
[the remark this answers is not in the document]
aiven-anton: The flush happens at the exit of this block: 

https://github.com/aiven/karapace/blob/2d5ab720f388482e8bd1ea80ef15bb878fbf5232/karapace/backup/api.py#LL445C42-L445C42

karapace/backup/api.py
```diff
@@ -390,14 +378,24 @@ def _handle_restore_topic(
...
+def check_producer_exception() -> None:
+    if producer_exception is not None:
+        raise BackupDataRestorationError("Error while producing restored messages") from producer_exception
 
 
 def _handle_producer_send(
     instruction: ProducerSend,
     producer: KafkaProducer,
 ) -> None:
+    check_producer_exception()
```
aiven-anton, line 398: issue: Let's document (with comment) the reason why we're also doing this here. Is it to reach a stop quicker than waiting for the call in `restore_backup()`?

karapace/backup/api.py
```diff
@@ -390,14 +378,24 @@ def _handle_restore_topic(
         raise BackupTopicAlreadyExists(f"Topic to restore '{instruction.topic_name}' already exists")
 
 
+producer_exception: Exception | None = None
+
+
 def _raise_backup_error(exception: Exception) -> NoReturn:
-    raise BackupDataRestorationError("Error while producing restored messages") from exception
+    global producer_exception  # pylint: disable=global-statement
+    producer_exception = exception
```
aiven-anton, line 386: issue: Let's collect all exceptions that occur instead of just the last? I believe the `check_producer_exception()` call in `_handle_producer_send()` doesn't make any difference for the possibility of multiple underlying exceptions. In theory all `_handle_producer_send`s can be executed before the first actual send succeeds.

karapace/backup/api.py
```diff
@@ -390,14 +378,24 @@ def _handle_restore_topic(
         raise BackupTopicAlreadyExists(f"Topic to restore '{instruction.topic_name}' already exists")
 
 
-def _raise_backup_error(exception: Exception) -> NoReturn:
-    raise BackupDataRestorationError("Error while producing restored messages") from exception
+producer_exception: Exception | None = None
+
+
+def _raise_backup_error(exception: Exception) -> None:
```
aiven-anton, line 384:
```suggestion
def _store_backup_error(exception: Exception) -> None:
```

issue: This is no longer raising.

karapace/backup/api.py
```diff
@@ -449,7 +468,7 @@ def restore_backup(
     # We set up an ExitStack context, so that we can enter the producer context only
     # after processing a RestoreTopic instruction.
     with contextlib.ExitStack() as stack:
-        producer = None
+        producer: BackupRestorationProducer | None = None
```
aiven-anton, line 471: thought (non-blocking): It's surprising that this needs an annotation now, but didn't before.

karapace/backup/api.py
```diff
@@ -263,29 +262,53 @@ def _consumer(config: Config, topic: str) -> Iterator[KafkaConsumer]:
...
-        yield None
-    finally:
-        Future.error_on_callbacks = global_value
+class BackupRestorationProducer:
+    def __init__(self, kafka_producer: KafkaProducer) -> None:
+        self.producer = kafka_producer
+        self.producer_exception: Exception | None = None
+
+    def _producer_error_callback(self, exception: Exception) -> None:
+        LOG.error("Producer error: %r", exception)
```
aiven-anton, line 271:
```suggestion
        LOG.exception(exception)
```

issue: I think we might want to log the full exception with traceback and all here?

karapace/backup/api.py
```diff
@@ -263,29 +262,53 @@ def _consumer(config: Config, topic: str) -> Iterator[KafkaConsumer]:
...
 
-@contextlib.contextmanager
-def _enable_producer_callback_errors() -> Iterator[None]:
-    global_value = Future.error_on_callbacks
-    Future.error_on_callbacks = True
-    try:
-        yield None
-    finally:
-        Future.error_on_callbacks = global_value
+class BackupRestorationProducer:
```
aiven-anton, line 265: thought (non-blocking): I'm not a very big fan of this wrapper, it seems like we could have easily just bound parameters to a callback function inline inside `restore_backup` to achieve the same thing. That would have avoided having to echo the pretty big signature of the `.send()` method. No need to change though, this is somewhat opinionated.

karapace/backup/api.py
giuseppelillo: it's not necessary, i just sometimes use type annotations to be more explicit about what it's being created, in this case i wanted to be explicit that here we're using a `BackupRestorationProducer` and not a normal producer

aiven-anton: If it's not necessary, I'd skip it and just rely on inference. (opinionated, so feel free to ignore! 🙂)

karapace/backup/api.py
```diff
@@ -450,6 +435,16 @@ def restore_backup(
     # after processing a RestoreTopic instruction.
     with contextlib.ExitStack() as stack:
         producer = None
+        _producer_exception = None
+
+        def _producer_error_callback(exception: Exception) -> None:
+            LOG.error("Producer error: %r", exception)
```
jjaakola-aiven, line 441:
```suggestion
            LOG.error("Producer error callback", exc_info=exception)
```

karapace/backup/api.py
aiven-anton: It seems this one regressed.

karapace/backup/api.py
```diff
@@ -461,10 +458,16 @@ def restore_backup(
             elif isinstance(instruction, ProducerSend):
                 if producer is None:
                     raise RuntimeError("Backend has not yet sent RestoreTopic.")
-                _handle_producer_send(instruction, producer)
+                _handle_producer_send(instruction, producer, _producer_error_callback)
+                # Immediately check if producer send generated an exception
```
aiven-anton, line 462: issue: Please, let's document the _why_ here, not only the _what_. The comments still don't communicate that this call is an optimization that in some cases might reduce time-to-error, and they don't document that the other call is _essential_ for correct behavior.

jjaakola-aiven (dismissed):

karapace/backup/api.py
```diff
@@ -461,10 +458,16 @@ def restore_backup(
             elif isinstance(instruction, ProducerSend):
                 if producer is None:
                     raise RuntimeError("Backend has not yet sent RestoreTopic.")
-                _handle_producer_send(instruction, producer)
+                _handle_producer_send(instruction, producer, _producer_error_callback)
+                # Immediately check if producer.send() generated an exception
```
aiven-anton, line 462:
```suggestion
                # Immediately check if producer.send() generated an exception. This call is
                # only an optimization, as producing as asynchronous and no sends might
                # have been executed once we reach this line.
```

karapace/backup/api.py
```diff
@@ -461,10 +458,16 @@ def restore_backup(
...
+                _check_producer_exception()
             else:
                 assert_never(instruction)
 
+    # Check if an exception was raised after the producer was flushed and closed
+    # by `kafka_producer_from_config` context manager
```
aiven-anton, line 468:
```suggestion
    # Check if an exception was raised after the producer was flushed and closed
    # by `kafka_producer_from_config` context manager. As opposed to the previous
    # call, this one is essential for correct behavior, as when we reach this point the
    # producer can no longer be sending messages (it has been flushed and closed).
```

karapace/backup/api.py
aiven-anton: Yes ☝️

jjaakola-aiven (approved):

PR #661, opened 2023-06-20
Fix exception handler fetching raw schema after delete

tvainika:

Fix exception handler after fetching raw schema after delete

# Why this way

Fix exception handling to avoid throwing internal server errors.

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -2674,6 +2674,12 @@ async def test_schema_hard_delete_whole_schema(registry_async_client: Client) ->
     assert res.json()["error_code"] == 40401
     assert res.json()["message"] == f"Subject '{subject}' not found."
 
+    # Check that fetching unescaped schema gives valid error message
+    res = await registry_async_client.get(f"subjects/{subject}/versions/latest/schema")
+    assert res.status_code == 404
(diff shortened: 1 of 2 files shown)


tests/integration/test_schema.py
```diff
@@ -2674,6 +2674,12 @@ async def test_schema_hard_delete_whole_schema(registry_async_client: Client) ->
     assert res.json()["error_code"] == 40401
     assert res.json()["message"] == f"Subject '{subject}' not found."
 
+    # Check that fetching unescaped schema gives valid error message
```
aiven-anton, line 2677: question: Just looking at this test, it's quite hard to see why escaping is relevant here. Isn't the reason for 404 that the schema was deleted?

tvainika: There are two endpoints to fetch schema, `/subjects/{{subject}}/versions/{{version}}` returning JSON object and `/subjects/{{subject}}/versions/{{version}}/schema` returning only "unescaped" or "raw" schema. The first endpoint was handling error already fine & tests, the second one was somehow missing this exception handling

aiven-anton: Ah, that makes sense, thanks for explaining 🙏

aiven-anton (approved):

PR #662, opened 2023-06-22
Fix partitioning if no key given

tvainika:

Default partitioner uses random partition if no key is given.  Add test that verifies random partition assignment with time-out.

# Why this way

Internally no key was converted to zero length byte array, which caused kafka library partitioner to handle it differently than no key.


aiven-anton (approved):
Fix looks good to me 👍

PR #664, opened 2023-06-22
fix: inconsistent schema during message produce

eliax1996:
This pr is ment to fix [this issue](https://github.com/aiven/karapace/issues/658) the test where we try this problem is `test_publish_with_schema_id_of_another_subject`.

The list of subjects bound to a topic has been added to the `/schemas/ids/{id}` if the `{"includeSubjects": "True"}` param are included. This helps the rest proxy in understand if a certain subject when it's required to produce a message providing only the `SchemaID`, once the REST proxy get the Schema and the list of allowed Subjects it calculate the subject and checks if the required subject it's allowed or not.

### DIFF
--- a/karapace/client.py
+++ b/karapace/client.py
@@ -90,6 +90,7 @@ class Client:
         json: JsonData = None,
         headers: Optional[Headers] = None,
         auth: Optional[BasicAuth] = None,
+        params: Optional[Mapping[str, str]] = None,
     ) -> Result:
         path = self.path_for(path)
         if not headers:
@@ -101,6 +102,7 @@ class Client:
             headers=headers,
             auth=auth,
             ssl=self.ssl_mode,
+            params=params,
         ) as res:
             # required for forcing the response body conversion to json despite missing valid Accept headers
             json_result = await res.json(content_type=None)
--- a/karapace/in_memory_database.py
+++ b/karapace/in_memory_database.py
@@ -11,7 +11,7 @@ from karapace.schema_models import SchemaVersion, TypedSchema
 from karapace.schema_references import Reference, Referents
 from karapace.typing import ResolvedVersion, SchemaId, Subject
 from threading import Lock, RLock
-from typing import Iterable
+from typing import Iterable, Sequence
 
 import logging
 
@@ -111,7 +111,7 @@ class InMemoryDatabase:
         version: ResolvedVersion,
         deleted: bool,
         schema: TypedSchema,
-        references: list[Reference] | None,
+        references: Sequence[Reference] | None,
     ) -> None:
         with self.schema_lock_thread:
             self.global_schema_id = max(self.global_schema_id, schema_id)
@@ -183,6 +183,17 @@ class InMemoryDatabase:
                     selected_schemas = [schema for schema in selected_schemas if schema.deleted is False]
                 res_schemas[subject] = selected_schemas
         return res_schemas
+
+    def subjects_for_schema(self, schema_id: SchemaId) -> list[Subject]:
+        subjects = []
+        with self.schema_lock_thread:
+            for subject, subject_data in self.subjects.items():
+                for version in subject_data.schemas.values():
+                    if version.deleted is False and version.schema_id == schema_id:
+                        subjects.append(subject)
+                        break
+
+        return subjects
 
     def find_schema_versions_by_schema_id(self, *, schema_id: SchemaId, include_deleted: bool) -> list[SchemaVersion]:
         schema_versions: list[SchemaVersion] = []
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -17,12 +17,15 @@ from karapace.errors import InvalidSchema
 from karapace.kafka_rest_apis.admin import KafkaRestAdminClient
 from karapace.kafka_rest_apis.consumer_manager import ConsumerManager
 from karapace.kafka_rest_apis.error_codes import RESTErrorCodes
+from karapace.kafka_rest_apis.schema_cache import TopicSchemaCache
 from karapace.karapace import KarapaceBase
 from karapace.rapu import HTTPRequest, HTTPResponse, JSON_CONTENT_TYPE
-from karapace.schema_reader import SchemaType
+from karapace.schema_models import TypedSchema, ValidatedTypedSchema
+from karapace.schema_type import SchemaType
 from karapace.serialization import InvalidMessageSchema, InvalidPayload, SchemaRegistrySerializer, SchemaRetrievalError
+from karapace.typing import SchemaId, Subject
 from karapace.utils import convert_to_int, json_encode, KarapaceKafkaClient
-from typing import Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import aiohttp.web
 import asyncio
@@ -416,7 +419,7 @@ class UserRestProxy:
         self.admin_client = None
         self.admin_lock = asyncio.Lock()
         self.metadata_cache = None
-        self.schemas_cache = {}
+        self.topic_schema_cache = TopicSchemaCache()
         self.consumer_manager = ConsumerManager(config=config, deserializer=self.serializer)
         self.init_admin_client()
         self._last_used = time.monotonic()
@@ -739,18 +742,83 @@ class UserRestProxy:
                 return False
         return isinstance(schema, str)
 
-    async def get_schema_id(self, data: dict, topic: str, prefix: str, schema_type: SchemaType) -> int:
-        log.debug("Retrieving schema id for %r", data)
-        if f"{prefix}_schema_id" in data and data[f"{prefix}_schema_id"] is not None:
-            log.debug("Will use schema id %d for serializing %s on topic %s", data[f"{prefix}_schema_id"], prefix, topic)
-            return int(data[f"{prefix}_schema_id"])
-        schema_str = data[f"{prefix}_schema"]
-        log.debug("Registering / Retrieving ID for schema %s", schema_str)
-        if schema_str not in self.schemas_cache:
-            subject_name = self.serializer.get_subject_name(topic, data[f"{prefix}_schema"], prefix, schema_type)
-            schema_id = await self.serializer.get_id_for_schema(data[f"{prefix}_schema"], subject_name, schema_type)
-            self.schemas_cache[schema_str] = schema_id
-        return self.schemas_cache[schema_str]
+    async def get_schema_id(
+        self,
+        data: dict,
+        topic: str,
+        prefix: str,
+        schema_type: SchemaType,
+    ) -> SchemaId:
+        """
+        This method search and validate the SchemaId for a request, it acts as a guard (In case of something wrong
+        throws an error).
+
+        :raises InvalidSchema:
+        """
+        log.debug("[resolve schema id] Retrieving schema id for %r", data)
+        schema_id: Union[SchemaId, None] = (
+            SchemaId(int(data[f"{prefix}_schema_id"])) if f"{prefix}_schema_id" in data else None
+        )
+        schema_str = data.get(f"{prefix}_schema")
+
+        if schema_id is None and schema_str is None:
+            raise InvalidSchema()
+
+        if schema_id is None:
+            parsed_schema = ValidatedTypedSchema.parse(schema_type, schema_str)
+            subject_name = self.serializer.get_subject_name(topic, parsed_schema, prefix, schema_type)
+            schema_id = await self._query_schema_id_from_cache_or_registry(parsed_schema, schema_str, subject_name)
+        else:
+
+            def subject_not_included(schema: TypedSchema, subjects: List[Subject]) -> bool:
+                subject = self.serializer.get_subject_name(topic, schema, prefix, schema_type)
+                return subject not in subjects
+
+            parsed_schema, valid_subjects = await self._query_schema_and_subjects(
+                schema_id,
+                need_new_call=subject_not_included,
+            )
+
+            if subject_not_included(parsed_schema, valid_subjects):
+                raise InvalidSchema()
+
+        return schema_id
+
+    async def _query_schema_and_subjects(
+        self, schema_id: SchemaId, *, need_new_call: Optional[Callable[[TypedSchema, List[Subject]], bool]]
+    ) -> Tuple[TypedSchema, List[Subject]]:
+        try:
+            return await self.serializer.get_schema_for_id(schema_id, need_new_call=need_new_call)
+        except SchemaRetrievalError as schema_error:
+            # if the schema doesn't exist we treated as if the error was due to an invalid schema
+            raise InvalidSchema() from schema_error
+
+    async def _query_schema_id_from_cache_or_registry(
+        self,
+        parsed_schema: ValidatedTypedSchema,
+        schema_str: str,
+        subject_name: Subject,
+    ) -> SchemaId:
+        """
+        Checks if the schema registered with a certain id match with the schema provided (you can provide
+        a valid id but place in the body a totally unrelated schema).
+        Also, here if we don't have a match we query the registry  since the cache could be evicted in the meanwhile
+        or the schema could be registered without passing though the http proxy.
+        """
+        schema_id = self.topic_schema_cache.get_schema_id(subject_name, parsed_schema)
+        if schema_id is None:
+            log.debug("[resolve schema id] Registering / Retrieving ID for %s and schema %s", subject_name, schema_str)
+            schema_id = await self.serializer.upsert_id_for_schema(parsed_schema, subject_name)
+            log.debug("[resolve schema id] Found schema id %s from registry for subject %s", schema_id, subject_name)
+            self.topic_schema_cache.set_schema(subject_name, schema_id, parsed_schema)
+        else:
+            log.debug(
+                "[resolve schema id] schema ID %s found from cache for %s and schema %s",
+                schema_id,
+                subject_name,
+                schema_str,
+            )
+        return schema_id
 
     async def validate_schema_info(self, data: dict, prefix: str, content_type: str, topic: str, schema_type: str):
         try:
@@ -788,10 +856,14 @@ class UserRestProxy:
                 status=HTTPStatus.REQUEST_TIMEOUT,
             )
         except InvalidSchema:
+            if f"{prefix}_schema" in data:
+                err = f'schema = {data[f"{prefix}_schema"]}'
+            else:
+                err = f'schema_id = {data[f"{prefix}_schema_id"]}'
             KafkaRest.r(
                 body={
                     "error_code": RESTErrorCodes.INVALID_DATA.value,
-                    "message": f'Invalid schema. format = {schema_type.value}, schema = {data[f"{prefix}_schema"]}',
+                    "message": f"Invalid schema. format = {schema_type.value}, {err}",
                 },
                 content_type=content_type,
                 status=HTTPStatus.UNPROCESSABLE_ENTITY,
@@ -882,7 +954,7 @@ class UserRestProxy:
         raise FormatError(f"Unknown format: {ser_format}")
 
     async def schema_serialize(self, obj: dict, schema_id: Optional[int]) -> bytes:
-        schema = await self.serializer.get_schema_for_id(schema_id)
+        schema, _ = await self.serializer.get_schema_for_id(schema_id)
         bytes_ = await self.serializer.serialize(schema, obj)
         return bytes_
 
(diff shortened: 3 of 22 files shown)


karapace/serialization.py
```diff
@@ -219,8 +219,13 @@ async def get_id_for_schema(self, schema: str, subject: str, schema_type: Schema
         except InvalidSchema as e:
             raise InvalidPayload(f"Schema string {schema} is invalid") from e
         schema_ser = str(schema_typed)
+
+        # if it's None shouldn't we firstly query to the original source? Maybe the schema has been added
```
eliax1996, line 223: @giuseppelillo wdyt?

requirements/requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.11
```
aiven-anton, line 2: This must be generated with Python 3.8.

karapace/typing.py
```diff
@@ -18,4 +18,7 @@
 Subject = NewType("Subject", str)
 Version = Union[int, str]
 ResolvedVersion = NewType("ResolvedVersion", int)
+# note: the SchemaID is a unique id among all the schemas (and each version should be assigned to a different id)
+# basically the mapping SchemaId -> TypedSchema is an ImmutableMapping.
```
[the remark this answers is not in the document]
eliax1996: during the development together with @giuseppelillo we wondered if the SchemaId was a "real" global id or an id valid only on the same user. I though that this info should be available during the definition of the type, I'm not referring to any specific mapping, here I'm just saying that it's a mapping in the mathematical term, it's a bidirectional function (an isomorphism) aka a mapping that do not vary since those objects represent values

karapace/in_memory_database.py
```diff
@@ -218,6 +218,15 @@ def find_subject_schemas(self, *, subject: Subject, include_deleted: bool) -> di
                 if schema_version.deleted is False
             }
 
+    def schema_id_registered_to_subject(self, *, subject: Subject, schema_id: SchemaId) -> bool:
+        if subject not in self.subjects:
+            return False
+        subject_schemas = self.subjects[subject]
```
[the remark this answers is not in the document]
eliax1996: going to do that but to me it's really fishy/antipattern to use the exception as a control flow mechanism 😄

karapace/serialization.py
```diff
@@ -138,6 +139,12 @@ async def get_schema_for_id(self, schema_id: int) -> ParsedTypedSchema:
         except InvalidSchema as e:
             raise SchemaRetrievalError(f"Failed to parse schema string from response: {json_result}") from e
 
+    async def subject_has_schema_id_registered(self, subject: str, schema_id: SchemaId) -> bool:
+        result = await self.client.get(f"subjects/{subject}/registered_id/{schema_id}")
+        if not result.ok:
+            return False
+        return True
```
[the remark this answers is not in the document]
eliax1996: thanks :D

karapace/serialization.py
```diff
@@ -172,52 +179,69 @@ async def close(self) -> None:
...
-
         schema_id, schema = await self.registry_client.get_latest_schema(subject)
         async with self.state_lock:
             schema_ser = str(schema)
             self.schemas_to_ids[schema_ser] = schema_id
             self.ids_to_schemas[schema_id] = schema
         return schema
 
-    async def get_id_for_schema(self, schema: str, subject: str, schema_type: SchemaType) -> int:
+    async def get_schema_id_on_subject(self, subject: str, schema_id: SchemaId) -> bool:
```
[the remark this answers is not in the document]
eliax1996: good catch!

karapace/serialization.py
```diff
@@ -172,52 +179,69 @@ async def close(self) -> None:
...
             return self.schemas_to_ids[schema_ser]
+
+        # should we wrap the SchemaRetrievalError erro type?
         schema_id = await self.registry_client.post_new_schema(subject, schema_typed)
 
         async with self.state_lock:
             self.schemas_to_ids[schema_ser] = schema_id
             self.ids_to_schemas[schema_id] = schema_typed
-        return schema_id
+        return SchemaId(schema_id)
```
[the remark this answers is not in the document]
eliax1996: yeah, it was a quick hack not finished :D thanks also for this catch!

karapace/in_memory_database.py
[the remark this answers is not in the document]
aiven-anton: In Python, exceptions represent both error and panic :)

karapace/typing.py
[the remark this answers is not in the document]
eliax1996: resolved OOG (out of github) :D

karapace/serialization.py
giuseppelillo: Could this be tested?

tests/integration/test_rest.py
```diff
@@ -143,13 +147,20 @@ async def test_avro_publish_primitive_schema(rest_async_client, admin_client):
...
 async def test_avro_publish(rest_async_client, registry_async_client, admin_client):
     tn = new_topic(admin_client)
-    other_tn = new_topic(admin_client)
-    await wait_for_topics(rest_async_client, topic_names=[tn, other_tn], timeout=NEW_TOPIC_TIMEOUT, sleep=1)
+    await wait_for_topics(rest_async_client, topic_names=[tn], timeout=NEW_TOPIC_TIMEOUT, sleep=1)
     header = REST_HEADERS["avro"]
     # check succeeds with 1 record and brand new schema
-    res = await registry_async_client.post(f"subjects/{other_tn}/versions", json={"schema": second_schema_json})
-    assert res.ok
-    new_schema_id = res.json()["id"]
```
giuseppelillo, line 152: Is this is not covered anymore, or the behaviour tested previously has been replaced?

karapace/serialization.py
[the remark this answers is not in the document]
aiven-anton: This still doesn't look resolved, please don't mark it as such until it's actually fixed :)

karapace/kafka_rest_apis/__init__.py
```diff
@@ -739,18 +836,60 @@ def is_valid_schema_request(data: dict, prefix: str) -> bool:
...
-        log.debug("Registering / Retrieving ID for schema %s", schema_str)
-        if schema_str not in self.schemas_cache:
-            subject_name = self.serializer.get_subject_name(topic, data[f"{prefix}_schema"], prefix, schema_type)
-            schema_id = await self.serializer.get_id_for_schema(data[f"{prefix}_schema"], subject_name, schema_type)
-            self.schemas_cache[schema_str] = schema_id
-        return self.schemas_cache[schema_str]
+    async def get_schema_id(self, data: dict, topic: str, prefix: str, schema_type: SchemaType) -> SchemaId:
+        log.debug("[resolve schema id] Retrieving schema id for %r", data)
+        # Schema id or schema str if schema id is none.
+        schema_id: SchemaId | None = SchemaId(data[f"{prefix}_schema_id"]) if f"{prefix}_schema_id" in data else None
```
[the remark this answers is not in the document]
aiven-anton: Yeah, I think that's more in line with the previous implementation that also had `int(...)` 👍

karapace/kafka_rest_apis/__init__.py
```diff
@@ -739,18 +836,60 @@ def is_valid_schema_request(data: dict, prefix: str) -> bool:
...
+            registered = await self.serializer.has_schema_id_on_subject(subject_name, schema_id)
+            if registered:
+                log.debug("[resolve schema id] schema ID %s is registed for %s.", schema_id, subject_name)
+                schema: TypedSchema = await self.serializer.get_schema_for_id(schema_id)
+                self.topic_schema_cache.set_schema(subject_name, schema_id, schema)
+            else:
+                raise InvalidSchema()
+        else:
+            log.debug("[resolve schema id] schema ID %s found from cache directly for %s.", schema_id, subject_name)
+        return schema_id
```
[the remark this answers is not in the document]
eliax1996: changed the doc and refactored the method, can you take a look and tell me if now it's more clear?

tests/integration/test_rest.py
eliax1996: Now it is, thanks!

karapace/kafka_rest_apis/__init__.py
```diff
@@ -739,18 +742,88 @@ def is_valid_schema_request(data: dict, prefix: str) -> bool:
...
+
+        if schema_id is None:
+            schema_id = await self._query_schema_id_from_cache_or_registry(parsed_schema, schema_str, subject_name)
+        else:
+            await self._check_subject_has_schema(schema_id, subject_name, parsed_schema)
+
+        return schema_id
+
+    async def _check_subject_has_schema(
+        self, schema_id: SchemaId, subject_name: Subject, parsed_schema: Union[TypedSchema, None]
```
giuseppelillo, line 766:
```suggestion
        self, schema_id: SchemaId, subject_name: Subject, parsed_schema: TypedSchema | None
```

giuseppelillo: And also all the other instances

eliax1996: I cannot do that, we are targeting python 3.8

karapace/schema_registry_apis.py
```diff
@@ -921,6 +930,32 @@ async def subject_versions_list(
                 status=HTTPStatus.NOT_FOUND,
             )
 
+    async def subject_schema_id_registered(
+        self, content_type: str, *, subject: str, schema_id: str, user: User | None = None
+    ) -> None:
```
aiven-anton, line 935: question: I don't know rapu well enough to know this: what happens when we don't enter one of the two branches in the method. What will we return if the method returns `None`?

karapace/serialization.py
```diff
@@ -165,59 +170,70 @@ def __init__(
...
+        assert self.registry_client, "must not call this method after the object is closed."
+
         schema_ser = str(schema_typed)
+
+        # if it's None shouldn't we firstly query to the original source? Maybe the schema has been added
+        # without the rest proxy in between and so the map isn't updated!
         if schema_ser in self.schemas_to_ids:
             return self.schemas_to_ids[schema_ser]
+
+        # should we wrap the SchemaRetrievalError erro type?
```
aiven-anton: suggestion: I think either address these comments, or make them todos, or remove them.

karapace/serialization.py
```diff
@@ -165,59 +170,70 @@ def __init__(
...
+            namespace = "dummy"
+            if schema_type is SchemaType.AVRO:
+                if isinstance(schema.schema, avro.schema.NamedSchema):
+                    namespace = schema.schema.namespace
+            if schema_type is SchemaType.JSONSCHEMA:
+                namespace = schema.to_dict().get("namespace", "dummy")
+            #  Protobuf does not use namespaces in terms of AVRO
+            if schema_type is SchemaType.PROTOBUF:
+                namespace = ""
+        return Subject(f"{self.subject_name_strategy(topic_name, namespace)}-{subject_type}")
```
aiven-anton, line 195: issue: It looks like `namespace` can now be `None` when we reach the `return`, but it doesn't look like the (implicit) strategy interface supports `record_name: None`. So this doesn't look right to me.

karapace/schema_registry_apis.py
aiven-anton: To clarify, I mean: what will the resulting HTTP response be?

eliax1996: the result would be a 200 with an empty response body. It's tested in a couple of end to end rest test (I agree isn't clear and isn't beautiful)

karapace/serialization.py
eliax1996: I've already fixed this in my last commit, I resolved my doubt about this 😄 and I've removed a branch from the get_schema algorithm

karapace/schema_registry_apis.py
aiven-anton: Ok, I think it's fine, more an issue with the framework. I would consider responding with HTTP 204, but that's not a blocker IMO.

karapace/serialization.py
eliax1996: It's ugly but as the software is now this is working [issue here](https://github.com/aiven/karapace/issues/667).
We are always only using the `topic_name` strategy, isn't possible for the user to use another strategy. So if you look the specific implementation:
```python
NAME_STRATEGIES = {
    "topic_name": topic_name_strategy,
    "record_name": record_name_strategy,
    "topic_record_name": topic_record_name_strategy,
}
```

--->

```python
def topic_name_strategy(topic_name: str, record_name: str) -> str:  # pylint: disable=unused-argument
    return topic_name
```

you know why it works 😞

tests/integration/test_rest.py
```diff
@@ -479,6 +496,252 @@ async def test_publish_with_incompatible_data(rest_async_client, registry_async_
...
+    subject = "fake_subject-value"
+    fake_id = "12123912"
+
+    resp = await registry_async_client.get(f"/subjects/{subject}/registered_id/{fake_id}")
+    resp_json = resp.json()
+    assert resp_json == {
+        "error_code": SchemaErrorCodes.SUBJECT_NOT_FOUND.value,
+        "message": f"Subject '{subject}' not found.",
+    }, "query an unexisting subject should return the right status code"
+    assert resp.status_code == 404, "unexisting subject is a not found"
```
giuseppelillo, line 554: This should go inside schema registry test file

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
-addopts = -ra -q --tb=short --numprocesses auto --import-mode=importlib
+addopts = -ra -q --tb=short --import-mode=importlib
```
aiven-anton, line 2: issue: This changes how tests are run in CI, I don't think we should make this change, and if we should, let's do it in a separate PR where it can be discussed.

FWIW, I personally use direnv with this line:

```
export PYTEST_ADDOPTS="-vvvs -n0 --kafka-bootstrap-servers=localhost:9092 --timeout=300"
```

eliax1996: It's something that I forgot to remove 😄 because otherwise the debugger of pycharm wasn't working. Good I'm going to try with the Env!

karapace/in_memory_database.py
```diff
@@ -184,6 +184,17 @@ def find_schemas(self, *, include_deleted: bool, latest_only: bool) -> dict[Subj
...
         return res_schemas
 
+    def subjects_for_schema(self, schema_id: SchemaId) -> list[Subject]:
+        subjects = []
+        with self.schema_lock_thread:
+            for subject, subject_data in self.subjects.items():
+                for version in subject_data.schemas.values():
+                    if version.deleted is False and version.schema_id == schema_id:
+                        subjects.append(subject)
+                        break
```
aiven-anton, line 194: question: Why this break? Is a schema guaranteed to have only one subject, if so, why return a list?

eliax1996: The break it's for the iteration over the values of the map for the subject, not for the iteration over all the subjects. It's like: if you found this schema for this subject there's no reason to iterate over the other schemas, just add it to the list and start for the next subject

karapace/schema_registry_apis.py
```diff
@@ -511,6 +512,9 @@ def _has_subject_with_id() -> bool:
             response_body["references"] = [r.to_dict() for r in schema.references]
         if fetch_max_id:
             response_body["maxId"] = schema.max_id
+        if include_subjects:
+            response_body["subjects"] = self.schema_registry.database.subjects_for_schema(parsed_schema_id)
```
aiven-anton, line 516: I suspect this is quite cheap to include, I would just always include it.

Including fields conditionally makes it very hard to expose a strict schema once we start introducing FastAPI (and OpenAPI).

eliax1996: Do we have a valid load test to check the difference?

karapace/in_memory_database.py
aiven-anton: Right, I shouldn't be reading code before having coffee 🙈

karapace/client.py
```diff
@@ -90,18 +90,13 @@ async def get(
...
             headers = {}
         client = await self.get_client()
-        async with client.get(
-            path,
-            json=json,
-            headers=headers,
-            auth=auth,
-            ssl=self.ssl_mode,
-        ) as res:
+        async with client.get(path, json=json, headers=headers, auth=auth, ssl=self.ssl_mode, params=params) as res:
```
aiven-anton, line 99: nitpick: Consider magic trailing comma :)

karapace/kafka_rest_apis/__init__.py
```diff
@@ -739,18 +742,82 @@ def is_valid_schema_request(data: dict, prefix: str) -> bool:
...
+            schema_id = await self._query_schema_id_from_cache_or_registry(parsed_schema, schema_str, subject_name)
+        else:
+            parsed_schema, valid_subjects = await self._query_schema_and_subjects(schema_id, skip_cache=False)
+            subject_name = self.serializer.get_subject_name(topic, parsed_schema, prefix, schema_type)
+
+            if subject_name not in valid_subjects:
+                # first checking again the result by contacting directly the schema registry
+                _, updated_subjects = await self._query_schema_and_subjects(schema_id, skip_cache=True)
+                if subject_name not in updated_subjects:
+                    raise InvalidSchema()
```
aiven-anton, line 779: thought: It looks like this strategy means that the worst case will produce two over-the-network calls. If we instead encapsulated this logic inside the serializer, it seems like we should be able to determine whether the first try used the cache or not, and that way eliminate the need for one extra call?

karapace/serialization.py
```diff
@@ -165,62 +177,70 @@ def __init__(
...
+        # note: the post is idempotent, so it is like a get or insert (aka upsert)
         schema_id = await self.registry_client.post_new_schema(subject, schema_typed)
 
         async with self.state_lock:
             self.schemas_to_ids[schema_ser] = schema_id
             self.ids_to_schemas[schema_id] = schema_typed
         return schema_id
 
-    async def get_schema_for_id(self, schema_id: int) -> TypedSchema:
+    async def get_schema_for_id(self, schema_id: SchemaId, skip_cache: bool = False) -> [TypedSchema, List[Subject]]:
```
aiven-anton, line 232: issue: This return type is not valid, missing `tuple`.

karapace/serialization.py
```diff
@@ -108,17 +110,20 @@ async def get_latest_schema(self, subject: str) -> Tuple[int, ParsedTypedSchema]
             raise SchemaRetrievalError(f"Invalid result format: {json_result}")
         try:
             schema_type = SchemaType(json_result.get("schemaType", "AVRO"))
-            return json_result["id"], ParsedTypedSchema.parse(schema_type, json_result["schema"])
+            return SchemaId(json_result["id"]), ParsedTypedSchema.parse(schema_type, json_result["schema"])
         except InvalidSchema as e:
             raise SchemaRetrievalError(f"Failed to parse schema string from response: {json_result}") from e
 
-    async def get_schema_for_id(self, schema_id: int) -> ParsedTypedSchema:
+    async def get_schema_for_id(self, schema_id: SchemaId) -> [TypedSchema, List[Subject]]:
```
aiven-anton, line 117: issue: Also invalid return type.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -739,18 +742,82 @@ def is_valid_schema_request(data: dict, prefix: str) -> bool:
...
+
+        return schema_id
+
+    async def _query_schema_and_subjects(
+        self,
+        schema_id: SchemaId,
+        *,
+        skip_cache: bool,
+    ) -> [TypedSchema, List[Subject]]:
```
aiven-anton, line 788: issue: Another invalid return type.

karapace/kafka_rest_apis/__init__.py
eliax1996: not so trivially since we need the typed schema to compute the subject, but I found a way to always do only 1 network call :) if we pass down a function that having the schema and the list of allowed subjects can tell if the subject it's present or not we can always do the call if the result isn't cached or we can do the call only if the subject isn't present in the allowed subjects from the cache.

tests/integration/test_rest.py
```diff
@@ -479,6 +493,167 @@ async def test_publish_with_incompatible_data(rest_async_client, registry_async_
...
+        headers=REST_HEADERS["avro"],
+    )
+    assert res.status_code == 422
+    res_json = res.json()
+    assert res_json["error_code"] == 42205
+    assert "message" in res_json
+    assert "Invalid schema. format = AVRO, schema_id = 2" in res_json["message"]
+
+
+async def test_publish_write_schema_on_karapace_read_from_rest(rest_async_client, registry_async_client, admin_client):
```
giuseppelillo, line 599: Duplicates `test_publish_with_schema_id_of_another_subject`

giuseppelillo (changes requested):

giuseppelillo (approved):

PR #666, opened 2023-06-26
feat(rest:consumer): add timestamp to consumer fetch response

jeqo:

Adds `timestamp` to the consumer records response.

# Why this way

Timestamp is already available on the internal response. This implementation is following the same approach as e.g offset.


aiven-anton (approved):
LGTM

Issue #668, opened 2023-06-30
Using '-1' as a version does not return the latest version


mattieserver:
# What happened?
We trying to get the latest version of a schema with '-1' as version you get the following error:
```
URL: /subjects/XXXXXX/versions/-1
{"error_code":42202,"message":"The specified version '-1' is not a valid version id. Allowed values are between [1, 2^31-1] and the string \"latest\""}
```

# What did you expect to happen?

'-1' should work the same as using 'latest' as the version.

# What else do we need to know?

Docker image version: 3.6.1

mattieserver:
If you look at https://github.com/aiven/karapace/blob/ec75a340d33a36c99eb5d107876fa08cb6ca38f9/karapace/schema_references.py#L79 it seems like it should have worked.

nosahama:
Fixed in #885.

PR #669, opened 2023-06-30
Add support for '-1' as version

mattieserver:
This allows adding '-1' as the version just like 'latest'.

References: #668

# Why this way

This seems like the quickest way to add support for the '-1'

### DIFF
--- a/karapace/schema_registry.py
+++ b/karapace/schema_registry.py
@@ -54,6 +54,8 @@ def _resolve_version(
 def validate_version(version: Version) -> Version:
     try:
         version_number = int(version)
+        if version_number == -1:
+            return "latest"
         if version_number > 0:
             return version
         raise InvalidVersion(f"Invalid version {version_number}")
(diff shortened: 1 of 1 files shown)


jjaakola-aiven (commented):
This is the easiest way to add the `-1` support. I'd prefer to move the responsibility of resolving to a `Version` class and the validation of the input string closer to the REST API. That would mean a bit more of refactoring:
 * Move `karapace.typing.Version` to `karapace.schema_models.Version`
 * Change to a class and implement the resolving and validation.
 * Change the `_resolve_version` calls `version.resolve(schema_versions)`.
 * Create the version objects from the input string in `karapace.schema_registry_api`.
 * Add tests

mattieserver:
I expected something like this 😄 

I try to implement the changes like you requested.

aiven-anton:
This should possibly be kept in sync with the respective handling here: https://github.com/Aiven-Open/karapace/blob/4ab7e9e3dfe64544117fdce594c742175e15b8e6/karapace/schema_reader.py#L572

nosahama:
#668 will be fixed with #885, closing this PR.

PR #675, opened 2023-07-07
Remove Subjects info from the GET API response.

sebinsunny:
The subjects info in the API response is throwing an error in the customer client library. The fix is to remove it from the API response.

```
 curl https://avnadmin:AVNS_IwJRh4AyJf5kUh4@kafka-sebin-dev-sandbox.aivencloud.com:12696/schemas/ids/1 
{"schema":"{\"doc\":\"example\",\"fields\":[{\"default\":5,\"doc\":\"my test number\",\"name\":\"test\",\"namespace\":\"test\",\"type\":\"int\"}],\"name\":\"example\",\"namespace\":\"example\",\"type\":\"record\"}","subjects":["test"]}
```
```

Exception in thread "test_subject" org.apache.kafka.streams.errors.StreamsException: Deserialization exception handler is set to fail upon a deserialization error. If you would rather have the streaming pipeline continue after a deserialization error, please set the default.deserialization.exception.handler appropriately.
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "subjects" (class io.confluent.kafka.schemaregistry.client.rest.entities.SchemaString), not marked as ignorable (4 known properties: "references", "schema", "schemaType", "maxId"])

```

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -505,9 +505,7 @@ class KarapaceSchemaRegistryController(KarapaceBase):
                 status=HTTPStatus.NOT_FOUND,
             )
 
-        subjects = self.schema_registry.database.subjects_for_schema(parsed_schema_id)
-
-        response_body = {"schema": schema.schema_str, "subjects": subjects}
+        response_body = {"schema": schema.schema_str}
         if schema.schema_type is not SchemaType.AVRO:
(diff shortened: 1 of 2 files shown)


eliax1996:
@sebinsunny thanks for the contribution, the issue it's addressed [here](https://github.com/aiven/karapace/pull/676)

PR #681, opened 2023-07-20
Resolve schema reference recursively during schema parsing

### LINKED ISSUE #679: Rest Proxy does not support protobuf schemas with references
# What happened?

When publishing a new message using a schema_id associated to a protobuf schema which uses references, Rest Proxy responds with a 422 even though the message is correct. This happens because the protobuf parser raises a `ProtobufUnresolvedDependencyException`. The same happens when consuming messages.

## Example

POST `/subjects/ref/versions`
```json
{
    "schemaType": "PROTOBUF",
    "schema": "syntax = \"proto3\"; message Reference { string name = 1; }"
}
```

POST `/subjects/topic-value/versions`
```json
{
    "schemaType": "PROTOBUF",
    "schema": "syntax = \"proto3\"; import \"Reference.proto\"; message Example { Reference example = 1; }",
    "references": [{"name": "Reference.proto", "subject": "ref", "version": 1}]
}
```

POST `/topics/topic`
`Content-Type: application/vnd.kafka.protobuf.v2+json`
```json
{
    "value_schema_id": 2,  # schema id from prev response
    "records": [
        {
            "value": {
                "example": {
                    "name": "myname"
                }
            }
        }
    ]
}
```

Response:
```
{
    "error_code": 42205,
    "message": "Invalid schema. format = PROTOBUF, schema_id = 2"
}
```

eliax1996:
The previous implementation of the schema resolver wasn't resolving recursively the entities of a schema that was containing references. With that implementation the schema are recursively resolved and the references are sent to the parser.

Closes #679

### DIFF
--- a/karapace/serialization.py
+++ b/karapace/serialization.py
@@ -5,17 +5,19 @@ See LICENSE for details
 from aiohttp import BasicAuth
 from avro.io import BinaryDecoder, BinaryEncoder, DatumReader, DatumWriter
 from cachetools import TTLCache
+from functools import lru_cache
 from google.protobuf.message import DecodeError
 from jsonschema import ValidationError
 from karapace.client import Client
+from karapace.dependency import Dependency
 from karapace.errors import InvalidReferences
 from karapace.protobuf.exception import ProtobufTypeException
 from karapace.protobuf.io import ProtobufDatumReader, ProtobufDatumWriter
 from karapace.schema_models import InvalidSchema, ParsedTypedSchema, SchemaType, TypedSchema, ValidatedTypedSchema
-from karapace.schema_references import Reference, reference_from_mapping
-from karapace.typing import SchemaId, Subject
+from karapace.schema_references import LatestVersionReference, Reference, reference_from_mapping
+from karapace.typing import ResolvedVersion, SchemaId, Subject
 from karapace.utils import json_decode, json_encode
-from typing import Any, Callable, Dict, List, MutableMapping, Optional, Tuple
+from typing import Any, Callable, Dict, List, MutableMapping, Optional, Set, Tuple
 from urllib.parse import quote
 
 import asyncio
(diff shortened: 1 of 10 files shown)


tvainika:
Commit message could be improved to explain the content, not only reference to an issue...

karapace/serialization.py
```diff
@@ -101,16 +102,43 @@ async def post_new_schema(
             raise SchemaRetrievalError(result.json())
         return SchemaId(result.json()["id"])
 
-    async def get_latest_schema(self, subject: str) -> Tuple[SchemaId, ParsedTypedSchema]:
-        result = await self.client.get(f"subjects/{quote(subject)}/versions/latest")
+    async def get_schema(
+        self,
+        subject: Subject,
+        version: Optional[ResolvedVersion] = None,
```
giuseppelillo, line 108: This probably needs a docstring explaining that `version == None` means latest version.

eliax1996:
@giuseppelillo I've added an `lru_cache` annotation to the schema resolver to decrease the cost of resolving recently used schemas.
I've also created the `_get_schema_r` method. In that way, if for e.g. a schema has 2 dependencies (like the dependency A that is depending on a dependency B that is on its own dependent again on A) the last call stack would be something like:
```python3
_get_schema_r(subject(A), id(A), set())
_get_schema_r(subject(B), id(B), set((subject(A), id(A)))
_get_schema_r(subject(A), id(A), set((subject(A), id(A), (subject(B), id(B))))
~~~~ Raises  InvalidSchema(
                f"The schema has at least a cycle in dependencies, "
                f"one path of the cycle is given by the following nodes: {set((subject(A), id(A), (subject(B), id(B))))}"
            )
```

Consider that I've just inserted here this implementation since as you mentioned we couldn't prevent the user from producing an invalid schema that is parsed from our processing.
But, as I've said offline I think we should reject this types of registrations and we should discard invalid messages from the kafka processing queue by preventing such malformed schemas from being considered at all in our registry.
PPS schemas with cycles in dependencies [aren't valid Avro schemas](https://stackoverflow.com/questions/45378889/how-to-include-proto-files-having-mutual-dependency) 😸

karapace/serialization.py
```diff
@@ -101,19 +103,79 @@ async def post_new_schema(
...
+                    json_result["schema"],
+                    references=references,
+                    dependencies=dependencies,
+                ),
+                ResolvedVersion(json_result["version"]),
+            )
         except InvalidSchema as e:
             raise SchemaRetrievalError(f"Failed to parse schema string from response: {json_result}") from e
 
+    @lru_cache(maxsize=1000)
```
jjaakola-aiven, line 157: I run some rough estimation of a `ValidatedTypedSchema` size. A fairly simple schema can cause the object size to be 10 KB. So having worst case happening, the LRU cache with 1000 entries would be 10 MB. There are some optimizations to reduce memory footprint with larger schema registries and I would not want to have much more here either. Would some smaller number be ok here?

eliax1996: Sure, what about 100? It would be ok to have 1MB occupied, theoretically we can even get rid of the memorisation completely, if you have a good enough number in mind we can opt for it

giuseppelillo (approved):

PR #689, opened 2023-07-27
fix: fix nested one of field handling for protobuf schema dependency verification

jjaakola-aiven:

On nested field with one of restriction the parent name did not include the element name.
This causes the dependency verification to fail as the actual full name of the nested one of field type
was not in the known used types.


eliax1996 (approved):
LGTM

PR #690, opened 2023-07-31
fix: respect nested runtime/.gitignore

aiven-anton:

#681 added an additional `.gitignore` line for the `runtime/` directory, however, this directory already has a [nested .gitignore](https://github.com/Aiven-Open/karapace/blob/main/runtime/.gitignore).

### Why this way

I think it might be preferable to keep the current nested .gitignore, as otherwise there's nothing stopping removing the `runtime/` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,3 @@ venv
 .python-version
 .hypothesis/
 .DS_Store
-# ignoring protobuf generated files.
-runtime/


eliax1996:
I'm ok in rollbacking :) but why do we need the `runtime` folder at all?

eliax1996:
I am trying to see if protobuf has problems without the folder being present or if it creates it.

.gitignore
```diff
@@ -20,5 +20,3 @@ venv
 .python-version
 .hypothesis/
 .DS_Store
-# ignoring protobuf generated files.
-runtime/
```
eliax1996, line 24: I've checked and the runtime folder needs to exist for the tests to work, but instead of having 2 different `.gitignore` files, what about adding here something like:
`runtime/**`?

eliax1996 (approved):

.gitignore
aiven-anton: @eliax1996 Git cannot represent empty directories, so I don't think that would work properly? I'm not a big fan of how this works currently though, but it doesn't seem like a high priority to change it either ... 🤷

PR #691, opened 2023-08-01
fix(rest:offsets): prevent `not_leader_for_partition` error [HH-2464]

jclarysse:
API requests to `/topics/<topic>/partitions/<partition_id>/offsets` often lead to `not_leader_for_partition` error as the request goes to the least loaded node. It needs to go to the leader instead.

[HH-2464]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ venv
 .python-version
 .hypothesis/
 .DS_Store
+# ignoring protobuf generated files.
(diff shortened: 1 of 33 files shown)


jclarysse:
after switching dev env my log got messy and rebase didn't work anymore, so closed this pr and created a new one https://github.com/Aiven-Open/karapace/pull/692

PR #696, opened 2023-08-08
Use KARAPACE_ADVERTISED_PORT environment value

### LINKED ISSUE #688: KARAPACE_ADVERTISED_PORT does not work
# What happened?

Sometime between when #400 was merged and now, the `advertised_port` was removed from `DEFAULTS` in `karapace/config.py`. This causes `set_settings_from_environment` not to work because it only considers settings available in `DEFAULTS`. `advertised_port` is not in `DEFAULTS`. It therefore does not look into `KARAPACE_ADVERTISED_PORT`.

`KARAPACE_ADVERTISED_PORT` should work properly.

There are a few other configuration options which don't exist in `DEFAULTS` but otherwise do exist. It may make sense to leave `DEFAULTS` alone and instead have `set_settings_from_environment` look at a different (complete) list of configuration options.

tvainika:

Use `KARAPACE_ADVERTISED_PORT` environment value

Fixes #688

# Why this way

Each configuration setting must be defined in `DEFAULTS` to enable reading those from environment variables.  There was regression with typing causing it to be removed earlier.

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -90,6 +90,7 @@ SASL_PLAIN_PASSWORD = "sasl_plain_password"
 DEFAULTS: ConfigDefaults = {
     "access_logs_debug": False,
     "access_log_class": None,
+    "advertised_port": None,
     "advertised_hostname": HOSTNAME,
     "advertised_protocol": "http",
     "bootstrap_uri": "127.0.0.1:9092",
@@ -175,7 +176,8 @@ def set_config_defaults(config: ConfigDefaults) -> Config:
     new_config.update(config)
 
     # Fallback to default port if `advertised_port` is not set
-    new_config.setdefault("advertised_port", new_config["port"])
+    if new_config["advertised_port"] is None:
+        new_config["advertised_port"] = new_config["port"]
 
     # Fallback to `advertised_*` constructed URI if not set
     if new_config["rest_base_uri"] is None:


karapace/config.py
```diff
@@ -23,7 +23,7 @@ class Config(TypedDict):
     access_logs_debug: bool
     access_log_class: type | None
     advertised_hostname: str
-    advertised_port: int
+    advertised_port: int | None
```
aiven-anton, line 26: question: I believe `advertised_port` should not be ` | None` here, only in `ConfigDefaults`?

aiven-anton: I.e., the `Config` type should model a valid configuration, we should not have to do assertions when we have an instance of it, we should already know that it is a valid configuration that can run.

aiven-anton: Since we cannot run with `advertised_port=None`, we should eliminate that illegal state on the typing level.

tvainika: I didn't notice our linters would accept it that way. In my testing even if `def set_config_defaults(config: ConfigDefaults) -> Config:` would return dict with `advertised_port: None` it would pass mypy so feels a bit cheating. But maybe it is this way simpler anyway.

aiven-anton: @tvainika Gotcha, that's unfortunate, but my intention with the introduction of `TypedDict`s here was rather meant as stop gap to be able to move forward with typing, rather than as an end solution. For now, I guess the trust on the class is one-sided.

aiven-anton (approved):

karapace/config.py
aiven-anton: I.e., the `Config` class makes a promise, and we trust that promise throughout the application, but nothing is actually verifying the promise. IMO we should use Pydantic for this whenever we get the chance.

aiven-anton:
Thanks for fixing this! 🙏

ziggythehamster:
Yes, thank you! :)

PR #705, opened 2023-08-30
Fix references support in check schema registered endpoint

tvainika:

Fix references support in parsing code that checks if a schema has been registered under the specified subject.

# Why this way

Parsing code did not pass parsed references correctly in the check code used by the endpoint.


aiven-anton (approved):

aiven-anton:
Thanks for addressing this 🍰

PR #711, opened 2023-09-06
Prometheus client metrics support

libretto:
This code introduces the Metrics class, which allows clients migrating to Karapace from alternative products to collect metrics similar to those they are accustomed to. It supports using Prometheus as an alternative to StatsD for collecting statistics in Karapace.


libretto:
@jjaakola-aiven @aiven-anton could You review this PR too?

libretto:
Hello @aiven-anton, I just try answer Your comments: 
> Hello @libretto,
> 
> Sorry for the long wait for a response here. We've convened internally about how to move forward with your contribution, and I'll summarize it here.
> 
> There are three issues we that must be addressed for this to be mergable:
> 
>     * ~Expose the Prometheus exporter under existing server instead of starting a separate one. Remove the configuration options for Prometheus host and port.~ _Let's leave this as-is._
> 
>     * Incorrect metric semantics. This has been iterated a couple times in review already. This means for instance that `request-size` needs to be changed into two separate metrics: `request_size_total + `request_count`. This applies to other metrics in the PR as well.

The metric semantics applied in this context are determined by the SchemaRegistry product, as detailed in their documentation. To achieve exact compatibility with SchemaRegistry, adherence to their specified metric names is necessary.

> 
>     * Remove the psutil connection count probing. This should rather be gathered by an OS-level metric, we don't want the application probing the OS for this itself. If the application is to report this, it needs to be based on some application level metric, like some actual internal counter of handled requests, and not by probing the OS for amount of process connections.

Is there a method to determine the number of connections in the Karapace application? I haven't been able to find a way to access the list of connections within our app. Is it possible I overlooked something? 

>       Since the `psutil` and `schedule` dependencies were brought in for this purpose only, let's also skip adding those.
> 
>     * Let's skip the updates to README.md, so that we can let these changes be undocumented and subject to change while we're stabilizing it.
 
Ok 

> 
> 
> Since the branches have diverged, let's also close [the original PR](https://github.com/Aiven-Open/karapace/pull/652) now, and make any future changes to _this_ branch only.

Ok

aiven-anton:
@libretto 

> The metric semantics applied in this context are determined by the SchemaRegistry product, as detailed in their documentation. To achieve exact compatibility with SchemaRegistry, adherence to their specified metric names is necessary.

Metrics is not an area where we'll aim for 1-1 compatibility with Confluent Schema Registry.

> Is there a method to determine the number of connections in the Karapace application? I haven't been able to find a way to access the list of connections within our app. Is it possible I overlooked something?

It's possible that this would have to be developed. For now, I would I just skip this and address it later. Do note that we have a long-term plan to switch from the homegrown web framework to FastAPI, so any such work might become moot.

libretto:
> * Incorrect metric semantics. This has been iterated a couple times in review already. This means for instance that `request-size` needs to be changed into two separate metrics: `request_size_total + `request_count`. This applies to other metrics in the PR as well.

@aiven-anton Do You mean it must be coded in the following way? 

`
      def request(self, size: int) -> None:
        self.request_size_total += size
        self.request_count += 1
        if not self.is_ready or self.stats_client is None:
            return
        if not isinstance(self.stats_client, StatsClient):
            raise RuntimeError("no StatsClient available")
        self.stats_client.gauge("request-size-total", self.request_size_total)
        self.stats_client.gauge("request-count", self.request_count)

`

libretto:
@aiven-anton  done. Review please.

libretto:
BTW. Is the usage of .gauge() and .timing() in the following functions acceptable?
 ...
 
 def are_we_master(self, is_master: bool) -> None:
        if not self.is_ready or self.stats_client is None:
            return
        if not isinstance(self.stats_client, StatsClient):
            raise RuntimeError("no StatsClient available")
        self.stats_client.gauge("master-slave-role", int(is_master))

    def latency(self, latency_ms: float) -> None:
        if not self.is_ready or self.stats_client is None:
            return
        if not isinstance(self.stats_client, StatsClient):
            raise RuntimeError("no StatsClient available")
        self.stats_client.timing("latency_ms", latency_ms)
        
...

aiven-anton (dismissed):

eliax1996 (changes requested):
Can we have some tests?

libretto:
> Can we have some tests?
We can certainly add some tests. However, the current version of Karapace Stats lacks test coverage, so we have no existing framework to guide us. At this point, it seems that only unit tests are feasible. I'm unsure how to implement integration tests for this part task.

eliax1996:
> > Can we have some tests?
> 
> We can certainly add some tests. However, the current version of Karapace Stats lacks test coverage, so we have no existing framework to guide us. At this point, it seems that only unit tests are feasible. I'm unsure how to implement integration tests for this part task.

Let's kick off with the unit tests. I'll provide an example soon on how to execute an integration test. It's crucial to have a tangible test that ensures our output its compliant to the standard format.
We need to verify that compliant consumers can effectively utilize and extract the metrics.

Without integration tests, even if the feature is technically sound, we lack a mechanism preventing us from the introduction of code that might break the consumers

karapace/metrics.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __call__(cls, *args: str, **kwargs: int) -> Singleton:
+        if cls._instance is None:
+            instance = super().__call__(*args, **kwargs)
+            cls._instance = instance
+        return cls._instance
+
+
+class Metrics(metaclass=Singleton):
+    stats_client: StatsClient
```
eliax1996, line 36: `stats_client: StatsClient | None = None`, I've run this locally with `metrics_extended` set to `False` and its failing because the `stats_client` its not going to be set

libretto: Fixed and unit test for this case is added.

nosahama:
I feel this could be simplified by providing only Prometheus support and using integrations when needed which should also be left to the users, i.e. if StatsD metrics are then needed, one could add the related mapping and deploy for themselves a [statsd-exporter](https://github.com/prometheus/statsd_exporter). 
If properly tested and benchmarked, it doesn't hurt to run the WSGI server (stats) as it's own thread, but it doesn't feel right in 2024 to write this much application code just to export some metrics with all the tools available. Prometheus is pull based, and as we already have a webserver running, we need to add an endpoint to scrape the metrics and probably a directory for the prometheus multiprocess registry.

libretto:
> Thoughts:
> 
>     * The PR description makes mention of only prometheus, it's confusing to see statsd implementations.
> 
>     * It'd be nice to see also somewhere in the code, all possible metrics and their related tags if any, that way I as a programming would be working based off an interface than just passing string values to the metrics handler.
> 
>     * I feel this could be simplified by providing only Prometheus support and using integrations when needed which should also be left to the users, i.e. if StatsD metrics are then needed, one could add the related mapping and deploy for themselves a [statsd-exporter](https://github.com/prometheus/statsd_exporter).
> 
>     * If properly tested and benchmarked, it doesn't hurt to run the WSGI server (stats) as it's own thread, but it doesn't feel right in 2024 to write this much application code just to export some metrics with all the tools available.
> 
>     * Prometheus is pull based, and as we already have a webserver running, we need to add an endpoint to scrape the metrics and probably a directory for the prometheus multiprocess registry.
> 
> 
> I will continue the work by @eliax1996 and try to add a sample metric and also some integration test.

This PR introduces the Metrics class, aiming to implement metrics similar to those found in comparable products. We intend to support both StatsD and Prometheus. Are you suggesting that we should enable the concurrent use of both?

nosahama:
> > Thoughts:
> > ```
> > * The PR description makes mention of only prometheus, it's confusing to see statsd implementations.
> > 
> > * It'd be nice to see also somewhere in the code, all possible metrics and their related tags if any, that way I as a programming would be working based off an interface than just passing string values to the metrics handler.
> > 
> > * I feel this could be simplified by providing only Prometheus support and using integrations when needed which should also be left to the users, i.e. if StatsD metrics are then needed, one could add the related mapping and deploy for themselves a [statsd-exporter](https://github.com/prometheus/statsd_exporter).
> > 
> > * If properly tested and benchmarked, it doesn't hurt to run the WSGI server (stats) as it's own thread, but it doesn't feel right in 2024 to write this much application code just to export some metrics with all the tools available.
> > 
> > * Prometheus is pull based, and as we already have a webserver running, we need to add an endpoint to scrape the metrics and probably a directory for the prometheus multiprocess registry.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I will continue the work by @eliax1996 and try to add a sample metric and also some integration test.
> 
> This PR introduces the Metrics class, aiming to implement metrics similar to those found in comparable products. We intend to support both StatsD and Prometheus. Are you suggesting that we should enable the concurrent use of both? I just revised the PR description for better clarity.

I do not think it is necessary to have both StatsD and Prometheus within the application itself, especially if there are tools to integrate one on top of the other.

nosahama:
@libretto I will close this PR once #902 is merged. 
In #902 we've introduced the building blocks for instrumenting the service using Prometheus. Following this, we can add more metrics as needed, until it is sufficient enough to provide compatibility with [Confluent SR metrics](https://docs.confluent.io/platform/current/schema-registry/monitoring.html).

PR #715, opened 2023-09-15
fix: explicitly set retries to 0

eliax1996:
Currently in the backup tool we leave implicitly the setting about the retries.
This (if the default value change) could lead to duplication of messages in the restore of a topic.


tvainika (dismissed):
Please don't remove runtime/.gitignore. It is there for convenience.

eliax1996:
Thanks for the catch @tvainika, don't know why I've committed it :) and also I don't know how I've deleted it :P

tvainika (approved):

PR #716, opened 2023-09-15
fix: enable self-referencing schemas

### LINKED ISSUE #713: Nested enum support in Protobuf message
# What is currently missing?

An `enum` embedded in a `message` is currently not supported: 

```
syntax = "proto3";

package x.y.;
message MyMessage {
  message MyNestedMessage {
    enum MyNestedEnum {
      VALUE_1 = 0;
    }
    MyNestedEnum enum1 = 1;
  }
}
```

The workaround is to move the nested enum to top level:

```
syntax = "proto3";

package x.y.;
enum MyNestedEnum {
  VALUE_1 = 0;
}
message MyMessage {
  message MyNestedMessage {
    MyNestedEnum enum1 = 1;
  }
}
```

# How could this be improved?

Add the support.

eliax1996:
Before this PR the protobuf schema was able to refer only to fully qualified names or names defined in an higher leaf of the parsing tree.
This commit enable a schema to refer the same scope level if at the same leaf level there is defined exactly the field required after.

In an example:
```protobuf
enum FancyEnum {
  ...
}
message FancyMessage1 {
   FancyEnum field ...
}
```
 `FancyMessage1` is allowed to call `FancyEnum` because his scope is `/FancyMessage1` while the scope of `FancyEnum` it's `/` (the name itself doesn't count in the declarations, but take part only in field assignment inside the declarations, we could say that `field` has the scope set to `/FancyMessage1` but the type `FancyEnum` has also the scope `/`, only when referring to a type you enter in it's scope during the parsing procedure).

This was a valid declaration in the previous status.

Instead, in this example:

```protobuf
message FancyMessage1 {
   enum FancyEnum {
     ...
   }
   FancyEnum field ...
}
```

In the previous state was an invalid schema.

With this pr, since `FancyEnum field ...` it's declared inside the scope `/FancyMessage1` but also the requested type `FancyEnum` it's declared in the same scope, `/FancyMessage1` it's considered valid

Fixes #713

### DIFF
--- a/karapace/protobuf/dependency.py
+++ b/karapace/protobuf/dependency.py
@@ -8,7 +8,7 @@ See LICENSE for details
 from karapace.dependency import DependencyVerifierResult
 from karapace.protobuf.known_dependency import DependenciesHardcoded, KnownDependency
 from karapace.protobuf.one_of_element import OneOfElement
-from typing import List
+from typing import List, Optional, Set
 
 
 class ProtobufDependencyVerifier:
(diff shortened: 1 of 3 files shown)


tvainika:
This code still fails with slightly modified schema like
```
syntax = "proto3";

package fancy.company.in.party.v1;
message AnotherMessage {
  enum BamFancyEnum {
    // Hei! This is a comment!
    MY_AWESOME_FIELD = 0;
  }
  message WowANestedMessage {
    message DeeplyNestedMsg {
      BamFancyEnum im_tricky_im_referring_to_the_previous_enum = 1;
    }
  }
}
```
which I believe is as valid as your test case.

juha-aiven:
In description

> ...the avro schema was able to refer ...

You mean of course protobuf schema.

karapace/protobuf/dependency.py
```diff
@@ -35,13 +35,32 @@ def add_used_type(self, parent: str, element_type: str) -> None:
...
+            or (father_child_type is not None and father_child_type in declared_index)
+            or "." + used_type in declared_index
+        )
+
     def verify(self) -> DependencyVerifierResult:
         declared_index = set(self.declared_types)
         for used_type in self.used_types:
             delimiter = used_type.rfind(";")
+            father_child_type = None
             used_type_with_scope = ""
```
jjaakola-aiven, line 58: This `used_type_with_scope` could work similar as the `father_child_type`, set to `None` and then there is no need to pass `delimiter` to `is_type_declared`.

karapace/protobuf/schema.py
```diff
@@ -161,6 +161,9 @@ def _process_nested_type(
     ):
         verifier.add_declared_type(package_name + "." + parent_name + "." + element_type.name)
         verifier.add_declared_type(parent_name + "." + element_type.name)
+        anchestor_only = parent_name.find(".")
```
jjaakola-aiven, line 164: `ancestor`

jjaakola-aiven (approved):

PR #719, opened 2023-09-19
chore: Make pytest CI runs more verbose

aiven-anton:

Increase pytest verbosity, by removing the quiet flag as well as the `--tb=short` flag.

### Why this way

In this failing CI run, the fact that we use `--tb short` makes it impossible to tell what code path triggered the breaking call to `read_config()`.

https://github.com/Aiven-Open/karapace/actions/runs/6239301084/job/16936931606

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
-addopts = -ra -q --tb=short --numprocesses auto --import-mode=importlib
+addopts = -ra --numprocesses auto --import-mode=importlib
 timeout = 60
 timeout_func_only = true


juha-aiven:
Are you sure about removing `--tb short`? I always use that together with `-vv` and it gives me enough traceback info, at least I don't remember ever missing anything important.

aiven-anton:
@juha-aiven Yeah I really don't see the point of shortening tracebacks, especially not in CI for a parallelized test suite (due to hard to deterministically recreate exact order of execution). In the case mentioned I can't tell where the call originated due to tb short. 

It's still easy to override these settings in dev envs to have differing behavior there.

juha-aiven (approved):
Fine for me.

juha-aiven (approved):
👍🏻

tvainika (approved):

PR #722, opened 2023-09-25
test: add another couple of basic tests

eliax1996:
Even if the scope of the tests could be covered by others I think those are worth being added for a couple of reasons:
1. the avro test shows how to register two different subjects (one for the key and the other for the value) even if the schema is the same to produce correctly an avro message in a topic
2. the protobuf test shows that basic google entities are already included in the karapace schema but it works also if you register it explicitly

### DIFF
--- a/karapace/protobuf/dependency.py
+++ b/karapace/protobuf/dependency.py
@@ -11,22 +11,13 @@ from karapace.protobuf.one_of_element import OneOfElement
 from typing import List, Optional, Set
 
 
-class FieldNotUniquelyIdentifiableException(Exception):
-    pass
-
-
 class ProtobufDependencyVerifier:
     def __init__(self) -> None:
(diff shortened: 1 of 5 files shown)


jjaakola-aiven (approved):

eliax1996:
@jjaakola-aiven have you seen that the test it's failing, shouldn't be the google referenced type be one of the well known and already pre-installed? Because it's failing the test that use the import without register it first

eliax1996:
I've tried to fix that but basically doesn't seems to be easy to fix. I've even attempted during the test setup phase to create the file in the right folder but doesn't seem to be enough to make work the test with the schema added by the `proto` dependency.

To me sounds fishy that we rely on the schemas being already there but in practice we don't have a test for it (unless I've missed it).

Closing the pr unless I discover how to make the provided schema visible from the protoc executable

PR #731, opened 2023-10-04
REST proxy SASL OIDC authentication

matyaskuti:

OAuth credentials pass-through:
* Allow passing through OAuth2 credentials to the underlying Kafka service
* In case the incoming HTTP request's `Authorization` header has a `Bearer` token, will create Kafka clients with the `OAUTHBEARER` SASL mechanism and pass the token through to them
* Basic SASL authentication behaviour is unchanged
* Changes and the addition of the OAuth2 possibility only apply to the REST proxy

Proxy janitor extension:
* The user proxy janitor is extended to clean up proxies with soon-to-be-expired tokens to avoid Kafka clients going into a connection retry-loop until they are cleaned up by existing janitor logic
* Extracting the expiry time from the token doesn't include signature verification - even if the token has been tampered with, Kafka verifies the token and requests are rejected if it's invalid
* `pyjwt` is used over `python-jose` due to its smaller footprint - any future switch (eg. due to the need to actually verify credentials and use JWK) is trivial (switch dependencies, modify imports) due to `python-jose` building on `pyjwt` and having identical interfaces for the features added now

# Why this way

Preserve the current auth behaviour without duplicating the client creation, make the decision purely based on the incoming HTTP requests.

### DIFF
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -62,6 +62,7 @@ class Config(TypedDict):
     sasl_mechanism: str | None
     sasl_plain_username: str | None
     sasl_plain_password: str | None
+    sasl_oauth_token: str | None
     topic_name: str
     metadata_max_age_ms: int
     admin_metadata_max_age: int
@@ -131,6 +132,7 @@ DEFAULTS: ConfigDefaults = {
     "sasl_mechanism": None,
     "sasl_plain_username": None,
     SASL_PLAIN_PASSWORD: None,
+    "sasl_oauth_token": None,
     "topic_name": DEFAULT_SCHEMA_TOPIC,
     "metadata_max_age_ms": 60000,
     "admin_metadata_max_age": 5,
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 from contextlib import AsyncExitStack, closing
 from http import HTTPStatus
 from kafka.errors import (
+    AuthenticationFailedError,
     BrokerResponseError,
     KafkaTimeoutError,
     NoBrokersAvailable,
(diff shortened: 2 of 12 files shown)


karapace/kafka_rest_apis/auth_utils.py
jjaakola-aiven, line 1: `authentication.py`, I don't see need to `_utils` suffix.

karapace/kafka_rest_apis/auth_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    BASIC = "Basic"
+    BEARER = "Bearer"
+
+
+def raise_unauthorized() -> NoReturn:
+    raise HTTPResponse(
+        body='{"message": "Unauthorized"}',
+        status=HTTPStatus.UNAUTHORIZED,
+        content_type=JSON_CONTENT_TYPE,
+        headers={"WWW-Authenticate": 'Basic realm="Karapace REST Proxy"'},
```
jjaakola-aiven, line 30: I think the challenge is also dependant of the auth scheme, but I am not sure what to respond as this is called when authorization header is missing and actual mechanisms are defined at Kafka server.

karapace/config.py
```diff
@@ -62,6 +62,7 @@ class Config(TypedDict):
     sasl_mechanism: str | None
     sasl_plain_username: str | None
     sasl_plain_password: str | None
+    sasl_oauth_token: str | None
```
jjaakola-aiven, line 65: This would require documentation in `README.md`. Is this option needed?

jjaakola-aiven (commented):
What happens to producer/consumer when token expires?

karapace/config.py
matyaskuti: Unfortunately yes. Just like the already existing `sasl_plain_username` and `sasl_plain_password`, which are likewise extracted from an auth header and not "proper" configuration parameters, and thus undocumented in the `README.md`.

These are needed to pass info eg. from `KafkaRest` to the `UserProxy`.

karapace/kafka_rest_apis/auth_utils.py
matyaskuti: Yeah, I wasn't sure either, that's why I left it as-is. It made perfect sense while we only accepted basic auth. In theory we could have multiple of the same header, but `aiohttp` doesn't let it happen.

matyaskuti:
@jjaakola-aiven 
> What happens to producer/consumer when token expires?

Potentially producer and consumer clients could go into a retry-loop until the idle proxy janitor cleans up the `UserProxy` associated with the current Bearer token.

As discussed offline, I'll modify the current `idle_proxy_janitor` to take into account token expiry in a "preemptive" way, cleaning up soon-to-be-expired proxies before the token actually becomes invalid.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -86,6 +93,12 @@ async def _disconnect_idle_proxy_if_any(self) -> None:
         async with self._proxy_lock:
             # Always clean one at time, don't mutate dict while iterating
             for _key, _proxy in self.proxies.items():
+                # In case of an OAuth2/OIDC token, the proxy is to be cleaned up _before_ the token expires
+                # If the token is still valid within the tolerance time, idleness is still checked
+                if _proxy.auth_expiry and _proxy.auth_expiry < datetime.datetime.now() + AUTH_EXPIRY_TOLERANCE:
```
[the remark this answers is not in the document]
matyaskuti: Even without adding the TZ info here, there is no timezone related difference between this and the expiry time from the token (as that is converted from a Unix timestamp to a datetime without TZ info as well).

However, using UTC everywhere still makes sense 👍

jjaakola-aiven (approved):

jjaakola-aiven (approved):

PR #732, opened 2023-10-06
REST proxy to return 403 instead of 500 in case of deny by ACL

jclarysse:
With REST authorization enabled, users got http 500 when attempting to consume topics for which they are not authorized.

The change consists in catching the exception TopicAuthorizationFailedError as part of http 403 response. 
References: HH-2806


tvainika (approved):

tvainika:
Looks good overall, there is no much test for REST api auth, those would be nice, but not necessary to get this fixed. I can merge as soon as tests and linters pass. There seemed to be another linter failure though.

tvainika (approved):

PR #733, opened 2023-10-06
feature: replace the old parsing strategy and move towards a tree based one

eliax1996:
The scope of this pr is to replace the old parsing strategy and move towards a tree based type verification strategy.

From this commit it's possible to refer to a type using a partial path and not only a fully specified path. If we declare for e.g. the following schema:
```protobuf
syntax = "proto3";

package my.awesome.customer.plan.entity.v1beta1;

message CustomerPlan {
  string plan_name = 1;
}
```

When importing it it's sufficient to use a part of the full specified path (that in this case would be `my.awesome.customer.plan.entity.v1beta1.CustomerPlan`). As e.g. it would be enough to use `entity.v1beta1.CustomerPlan` or `v1beta1.CustomerPlan`, so the following declaration would be accepted:
```protobuf
syntax = "proto3";

package my.awesome.customer.plan.entity.v1beta1;
import "my/awesome/customer/plan/entity/v1beta1/entity.proto";

message CustomerPlanEvent {
  message Created {
    entity.v1beta1.CustomerPlan plan = 1;
  }
}
```


eliax1996:
This pr it's experimental, even if it's passing all the tests stuff can be still moved around and a final judgement should be given from @jjaakola-aiven before going forward. Please try as much valid (and invalid) schemas as possible and provide feedbacks if something it's fishy.
A sample test it's the `test_partial_path_in_protobuf`, feel free to start from here and craft new ones.

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
     return f"option {result};\n"
 
 
+@dataclasses.dataclass
```
jjaakola-aiven, line 112:
```suggestion
@dataclasses.dataclass(frozen=True)
```

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
...
+    used_attribute_type: str
+
+
+@dataclasses.dataclass(frozen=True)
+class TypeTree:
+    token: str
+    children: List["TypeTree"]
+
+    @property
+    def is_full_path_type(self) -> bool:
```
jjaakola-aiven, line 124: I am not sure of the name, just a proposal below. I also did consider here `is_leaf` or `is_terminal_node` as this is a tree structure. But domain related name here is a lot better.
```suggestion
    def is_fully_qualified_type(self) -> bool:
```

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
...
+        return len(self.children) == 0
+
+    def represent(self, level=0):
```
jjaakola-aiven, line 127:
```suggestion
    def represent(self, level=0) -> str:
```

karapace/protobuf/proto_file_element.py
```diff
@@ -29,18 +32,23 @@ def _collect_dependencies_types(compare_types: CompareTypes, dependencies: Optio
         _collect_dependencies_types(compare_types, sub_deps, is_self)
 
 
+# todo: transform those types into serious object with parsing that filters out invalid type name or invalid package name
+TypeName = str
```
aiven-anton, line 36: suggestion: Can we turn these into NewType now already? Bare type aliases like this are a bit devious as they give an impression of stricter type safety where there is none.

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
     return f"option {result};\n"
 
 
+@dataclasses.dataclass(frozen=True)
```
aiven-anton, line 112: nit (non-blocking): There exists a `@default_dataclass` that could be used here.

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
...
+        if self.is_full_path_type:
+            return f"{spacing}{self.token}"
+        child_repr = "\n".join(child.represent(level=level + 1) for child in self.children)
+        return f"{spacing}{self.token} -> \n{child_repr}"
+
+    def __repr__(self) -> str:
+        return self.represent()
+
+
+def add_new_type_r(parent_tree: TypeTree, remaining_tokens: List[str]) -> None:
```
aiven-anton, line 138: nit: What does `_r` mean here? Can it be expanded to some word?

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
...
+    if remaining_tokens:
+        token = remaining_tokens.pop()
```
aiven-anton, line 140: suggestion: This mutates a shared object. If we want do this, let's add some warning in docstring.

karapace/protobuf/schema.py
```diff
@@ -106,6 +109,49 @@ def option_element_string(option: OptionElement) -> str:
...
+@dataclasses.dataclass(frozen=True)
+class UsedType:
+    parent_object_qualified_name: str
+    used_attribute_type: str
+
+
+@dataclasses.dataclass(frozen=True)
+class TypeTree:
+    token: str
+    children: List["TypeTree"]
```
aiven-anton, line 121: suggestion (non-blocking): It would be comforting to make this immutable, but potentially requires a bit of refactoring of the constructing algorithm.

aiven-anton (commented):
I left some minor nits.

karapace/protobuf/schema.py
eliax1996: yeah, you are right 😄 was already on plan before merging, I've said that to @jjaakola-aiven before the review 😄 happy that also you noticed it, doing it now

karapace/protobuf/schema.py
eliax1996: it stands for `recursive` I don't know if there is a "pythonic" way of defining a function for the recursive side, I know I can define a function inside the `add_new_type` function but I don't really like nested functions. Maybe I can simply rename it as `_add_new_type` since it's a method internal and someone from external isn't expected to call the recursive one, wdyt?

eliax1996: the idea was just to provide an external "happy and nice" function without exposing types that make sense only when looking the recursive implementation, that to me it's an implementation detail and should't be exposed, probably the `_` prepending it's the appropiate solution

karapace/protobuf/proto_file_element.py
eliax1996: yeah, small effort, make sense to push for it

karapace/protobuf/schema.py
eliax1996: yeah, you are right 😄 was already on plan before merging, I've said that to @jjaakola-aiven before the review 😄 happy that also you noticed it, doing it now

karapace/protobuf/schema.py
eliax1996: mmm I can transform it into an immutable version by "simply" replacing the variable with `token[:-2]` but IMO we aren't adding value and we are penalizing a lot the performance, consider that this object it's created and consumptioned by the method itself (another reason to make it private)

karapace/protobuf/schema.py
aiven-anton: Yeah prepending `_` would be nice, but I'd also expand the `_r` to `_recurse`, not much gained from saving those few characters.

karapace/protobuf/schema.py
aiven-anton: That's fine, let's just add appropriate warning signs ;)

karapace/protobuf/schema.py
eliax1996: ok, experiment done and wasn't easy to rewrite the parser. I've tried to explain a little bit a possible follow-up, wdyt?

aiven-anton: Yeah, this is not a blocker, thanks for exploring the options 🙏 I'll mark as resolved.

jjaakola-aiven (approved):

Issue #724, opened 2023-09-27
Failure when producing events with the latest version 3.7.3


MSdracanovic:
Hello,
We are using Aiven Karapace REST API to publish event from our ERP to other entities in our company.
Recently with automatic maintenance update we got update from version 3.6.2 to version 3.7.3

After update we can produce any event from our system

error that we are receiving is

`{
    "error_code": 42205,
    "message": "Invalid schema. format = AVRO, schema_id = 7"
}`

We tested with simple example and got same result
schema
`{
  "fields": [
    {
      "name": "age",
      "type": "int"
    }
  ],
  "name": "Obj",
  "type": "record"
}`

payload 
`{
   "value_schema_id": 7,
   "records":[
      {
         "value":{
            "age":55
         }
      }
   ]
}`

Current Content-Type in header is `application/vnd.kafka.avro.v2+json`
When we change to `application/vnd.kafka.json.v2+json` we can produce message but our consumers can not consume it.

We want to avoid downgrade and to solve this issue.
Do you have any proposal on this.

Thanks in Advance

tvainika:
We have identified that in pull request #664 was additional validation with schema registered under matching subject, which was not fully backward compatible. This change was published with Karapace 3.7.0 in July.

If you are Aiven customer, please contact via Aiven support to get your services mitigation for this issue.

Our long term plan is to add configuration toggle for schema registered under a subject to validate the content when producing to a topic. But this can take some days to became available with next Karapace release.

MSdracanovic:
Thank for quick response.
After  we recreate schema with format name  topicname-value and topicname-key we manage to produce events.

PR #735, opened 2023-10-11
Fix subject name strategy config and add validation flag

tvainika:

Fix subject `name_strategy` configuration setting, which we have had kind of documented but not working. Now it is possible to set it, and use it. Also fixes documentation to match what has been the default in the implementation.

Add new configuration option `name_strategy_validation` (true by default) which allows skipping subject name strategy validation when producing messages with Kafka REST calls. 

References: #664 and #724

# Why this way

This enables easy backward compatible behaviour of producing with REST using any schema.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -462,6 +462,9 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
    * - ``name_strategy``
      - ``topic_name``
      - Name strategy to use when storing schemas from the kafka rest proxy service
+   * - ``name_strategy_validation``
+     - ``true``
+     - If enabled, validate that given schema is registered under used name strategy when producing messages from Kafka Rest
    * - ``master_election_strategy``
      - ``lowest``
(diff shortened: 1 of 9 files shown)


eliax1996:
Did I understood it correctly that this is globally and not for each topic? Doesn’t make sense to enable/disable it by topic? Same question for the strategy

tvainika:
Yes, this is global config that allows easy backward mode that is compatible with pre 3.7.0 karapace

eliax1996:
for disabling the schema I can agree that having something global can make sense but for the type of strategy we need to enable the customisation on a topic level. Also I would argue that someone could want to enable the validation in a topic based way since you can have flows that aren’t validated and flow that are

giuseppelillo (approved):
LGTM as a first approach for configuring subject name strategy. Topic level configuration is still needed, but should be tackled separately.

PR #736, opened 2023-10-11
Fix protobuf-compiler version to match Debian bullseye

tvainika:

Fix protobuf-compiler version to match Debian bullseye

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -32,7 +32,7 @@ RUN groupadd --system karapace \
  && chown --recursive karapace:karapace /opt/karapace /var/log/karapace
 
 # Install protobuf compiler.
-ARG PROTOBUF_COMPILER_VERSION="3.12.4-1"
+ARG PROTOBUF_COMPILER_VERSION="3.12.4-1+deb11u1"
 RUN apt-get update \
  && apt-get install --assume-yes --no-install-recommends \
     protobuf-compiler=$PROTOBUF_COMPILER_VERSION \


container/Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN groupadd --system karapace \
  && chown --recursive karapace:karapace /opt/karapace /var/log/karapace
 
 # Install protobuf compiler.
-ARG PROTOBUF_COMPILER_VERSION="3.12.4-1"
+ARG PROTOBUF_COMPILER_VERSION="3.12.4-1+deb11u1"
```
jjaakola-aiven, line 35: Maybe the version should be removed here as the compiler is from the Debian packages. There is only one version available.

jjaakola-aiven (changes requested):

container/Dockerfile
aiven-anton: I was the one adding this, without much thought to it other than to adhere to the linter: https://github.com/Aiven-Open/karapace/blob/9a4f91e48d03ad4084eb8c4d2729c06f58b0b356/.pre-commit-config.yaml#L63

We would just need to add another ignore there if we remove the version pin.

I'll refrain from commenting on whether it's a good idea to pin apt packages or not, but I'll link the rationale from the tool itself: https://github.com/hadolint/hadolint/wiki/DL3008#rationale. I'm fine with ditching this practice.

tvainika: I had rebased other PR on this just to get CI builds passing. I'm open to non-pinned versions here, but as the build is passing, I'm not planning to actively work on it now.

PR #739, opened 2023-10-17
test: new message addition its valid

eliax1996:
Add a new test for the case of adding a new message to the already created message


eliax1996:
This should be tested again once the parser its rewrite. Its a test that should be passed

PR #743, opened 2023-10-23
Return Unauthorized on malformed OAuth token

matyaskuti:

Handle JWT DecodeErrors when extracting the expiration timestamp from an OIDC/OAuth2 JWT token, otherwise this would result in an HTTP 500 response.

# Why this way

Catching PyJWT's `DecodeError` is sufficient, as all other exceptions from the library (see
https://pyjwt.readthedocs.io/en/stable/api.html#exceptions) are related to proper verification, which we do not do at the moment.

### DIFF
--- a/tests/unit/test_authentication.py
+++ b/tests/unit/test_authentication.py
@@ -21,6 +21,13 @@ import jwt
 import pytest
 
 
+def _assert_unauthorized_http_response(http_response: HTTPResponse) -> None:
+    assert http_response.body == '{"message": "Unauthorized"}'
+    assert http_response.status == HTTPStatus.UNAUTHORIZED
+    assert http_response.headers["Content-Type"] == JSON_CONTENT_TYPE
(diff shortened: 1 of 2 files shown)


tests/unit/test_authentication.py
```diff
@@ -88,6 +91,17 @@ def test_get_expiration_time_from_header(auth_header: str, expected_expiration:
     assert expiration == expected_expiration
 
 
+@pytest.mark.parametrize(
+    "auth_header",
+    (f"Bearer {jwt.encode({'exp': 1697013997}, 'secret')}XX", "Bearer NotAToken"),
+)
+def test_get_expiration_time_from_header_malformed_bearer_token_raises_unauthorized(auth_header: str) -> None:
```
aiven-anton, line 98: praise: Structuring tests this way makes it very easy to review and understand what the expected outcome is 👍

aiven-anton (approved):
This looks all good and sound 👍

PR #748, opened 2023-10-26
Bump mypy and fix problems in avro introspection code

aiven-anton:

Bumps mypy and fixes revealed problematic code in avro introspection code where wrong entities was used.

Supersedes and closes #747.

### DIFF
--- a/karapace/avro_dataclasses/introspect.py
+++ b/karapace/avro_dataclasses/introspect.py
@@ -5,7 +5,7 @@ See LICENSE for details
 
 from __future__ import annotations
 
-from .schema import AvroType, FieldSchema, RecordSchema
+from .schema import AvroType, EnumType, FieldSchema, MapType, RecordSchema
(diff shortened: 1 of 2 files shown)


karapace/avro_dataclasses/introspect.py
```diff
@@ -115,15 +115,11 @@ def _field_type(field: Field, type_: object) -> AvroType:  # pylint: disable=too
...
+        return MapType(
             {
                 "type": "map",
                 "values": _field_type(field, args[1]),
-                **(
-                    {"default": field.default_factory()}
-                    if field.default_factory is not MISSING
-                    else {}  # type: ignore[misc]
-                ),
+                **({"default": field.default_factory()} if field.default_factory is not MISSING else {}),
```
aiven-anton, line 122: Unfortunate victim to the cursed combination of 120 line-length limit + black ...

tvainika (approved):

PR #754, opened 2023-11-03
Added support to other naming strategy

eliax1996:
Before that commit the other strategies where existing but not effectively usable/tested in the codebase.
This pr adds the support and set tests to effectively test the naming strategies.
During the development I've also added a little bit more typing over the files I've touched to improve the codebase

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -461,7 +461,7 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
      - Runtime directory for the ``protoc`` protobuf schema parser and code generator
    * - ``name_strategy``
      - ``topic_name``
-     - Name strategy to use when storing schemas from the kafka rest proxy service
+     - Name strategy to use when storing schemas from the kafka rest proxy service. You can opt between ``name_strategy`` , ``record_name`` and ``topic_record_name``
    * - ``name_strategy_validation``
      - ``true``
      - If enabled, validate that given schema is registered under used name strategy when producing messages from Kafka Rest
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -6,8 +6,8 @@ See LICENSE for details
 """
 from __future__ import annotations
 
-from enum import Enum, unique
 from karapace.constants import DEFAULT_AIOHTTP_CLIENT_MAX_SIZE, DEFAULT_PRODUCER_MAX_REQUEST, DEFAULT_SCHEMA_TOPIC
+from karapace.typing import ElectionStrategy, NameStrategy
 from karapace.utils import json_decode, json_encode, JSONDecodeError
 from pathlib import Path
 from typing import IO, Mapping
@@ -158,19 +158,6 @@ class InvalidConfiguration(Exception):
     pass
 
 
-@unique
-class ElectionStrategy(Enum):
-    highest = "highest"
-    lowest = "lowest"
-
-
-@unique
-class NameStrategy(Enum):
-    topic_name = "topic_name"
-    record_name = "record_name"
-    topic_record_name = "topic_record_name"
-
-
 def parse_env_value(value: str) -> str | int | bool:
     # we only have ints, strings and bools in the config
     try:
@@ -273,8 +260,10 @@ def validate_config(config: Config) -> None:
     try:
         NameStrategy(name_strategy)
     except ValueError:
-        valid_strategies = [strategy.value for strategy in NameStrategy]
-        raise InvalidConfiguration(f"Invalid name strategy: {name_strategy}, valid values are {valid_strategies}") from None
+        valid_strategies = list(NameStrategy)
+        raise InvalidConfiguration(
+            f"Invalid default name strategy: {name_strategy}, valid values are {valid_strategies}"
+        ) from None
 
     if config["rest_authorization"] and config["sasl_bootstrap_uri"] is None:
         raise InvalidConfiguration(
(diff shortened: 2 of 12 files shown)


karapace/config.py
```diff
@@ -269,12 +281,14 @@ def validate_config(config: Config) -> None:
             f"Invalid master election strategy: {master_election_strategy}, valid values are {valid_strategies}"
         ) from None
 
-    name_strategy = config["name_strategy"]
+    deafault_name_strategy = config["name_strategy"]
```
tvainika, line 284: typo: deafault -> default

karapace/serialization.py
```diff
@@ -225,6 +227,25 @@ async def close(self):
         await self.client.close()
 
 
+def get_subject_name(
```
tvainika, line 230: I see that you only moved here code around, not changing. However now looking at this, is it correct code at all? For me it looks it takes only namespace of record, but is missing record name itself? `naming_strategy` functions take `record_name` as argument, not `namespace`. This is also reflected in the test you added. Are those expected values correct in that test is relevant question in this context.

tests/unit/test_serialization.py
```diff
@@ -310,3 +314,21 @@ async def test_deserialization_fails(default_config_path):
         await deserializer.deserialize(enc_bytes)
 
     assert mock_registry_client.method_calls == [call.get_schema_for_id(1)]
+
+
+@pytest.mark.parametrize(
+    "expected_subject,strategy,subject_type",
+    (
+        (Subject("foo-key"), NameStrategy.topic_name, SUBJECT_VALID_POSTFIX[0]),
```
tvainika, line 322: See comment related to this test.

But also this test is more readable if there is no import of `SUBJECT_VALID_POSTFIX` but instead each parametrized test here refers either to `SubjectType.key` or `SubjectType.value`, IMHO.

karapace/serialization.py
```diff
@@ -243,37 +264,17 @@ def __init__(
         else:
             registry_url = f"http://{self.config['registry_host']}:{self.config['registry_port']}"
             registry_client = SchemaRegistryClient(registry_url, session_auth=session_auth)
-        name_strategy = config.get("name_strategy", "topic_name")
-        self.subject_name_strategy = NAME_STRATEGIES.get(name_strategy, topic_name_strategy)
         self.registry_client: Optional[SchemaRegistryClient] = registry_client
         self.ids_to_schemas: Dict[int, TypedSchema] = {}
         self.ids_to_subjects: MutableMapping[int, List[Subject]] = TTLCache(maxsize=10000, ttl=600)
         self.schemas_to_ids: Dict[str, SchemaId] = {}
+        self._topic_strategy_cache: MutableMapping[TopicName, NameStrategy] = TTLCache(maxsize=10000, ttl=600)
```
tvainika, line 271: Does this belong to this PR or some other work?

tvainika (changes requested):
See inline the fundamental question how record name strategy should work to be clarified

karapace/serialization.py
eliax1996: ops, another piece I forgot. Thanks

tests/unit/test_serialization.py
eliax1996: Yeah, you are right. Initially I've wrote the test as you are suggesting, then I thought while adding the `SUBJECT_VALID_POSTFIX` variable, shouldn't we use the same also in test? But it looks terrible, going to revert

karapace/serialization.py
eliax1996: > is it correct code at all? For me it looks it takes only namespace of record, but is missing record name itself? 
good point, I've just moved stuff around and created a test to keep the behaviour. We need to find a real case where this is done and compare it to our implementation, I'm looking for it now

karapace/serialization.py
```diff
@@ -59,37 +62,59 @@ class SchemaUpdateError(SchemaError):
...
+
 
+def record_name_strategy(
+    topic_name: str,  # pylint: disable=unused-argument,
+    record_name: str | None,
+    subject_type: SubjectType,  # pylint: disable=unused-argument
+) -> Subject:
+    if record_name is None:
+        raise InvalidRecord(
+            "The provided record do not have a valid `record_name`, use another naming strategy or fix the schema"
```
giuseppelillo, line 84:
```suggestion
            "The provided record does not have a valid `record_name`, use another naming strategy or fix the schema"
```

eliax1996: maybe directly `doesn't`?

giuseppelillo (commented):
Overall LGTM. Could you also add a test on how the record and topic-record naming strategy are supposed to work with an Avro schema without namespace?

karapace/serialization.py
giuseppelillo: Whatever you prefer :D

tvainika (changes requested):
LGTM except documentation change is not correct...

README.rst
```diff
@@ -460,7 +460,7 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
      - ``runtime``
      - Runtime directory for the ``protoc`` protobuf schema parser and code generator
    * - ``name_strategy``
-     - ``topic_name``
+     - ``topic_name``, ``record_name``, ``topic_record_name``
```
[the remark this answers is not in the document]
eliax1996: 😵‍💫 pushing both suggestion/fix asap

tvainika (approved):

giuseppelillo (approved):

PR #757, opened 2023-11-08
Use confluent-kafka-python for Kafka admin client instead of kafka-python

matyaskuti:

Replace Kafka admin clients (including `KafkaRestAdminClient` and "vanilla" admin clients from `kafka-python`) with a new `confluent_kafka` based one. 

The main change is:
* The addition of the `KafkaAdminClient` in a [new module](https://github.com/Aiven-Open/karapace/pull/757/files#diff-45587f805273ca642239999779a87f5e0d0915737967a487020cec010b047af0)
* Which is more or less moving the `KafkaRestAdminClient` from its [original location](https://github.com/Aiven-Open/karapace/pull/757/files#diff-56b57faef5e16374947f786aef54c5488ffe64234ac7fe86279d01d19be127bc) and adapting to the new base class from confluent-kafka-python
* As well as splitting off the existing test suite and expanding it into [its own](https://github.com/Aiven-Open/karapace/pull/757/files#diff-957494df0582911018996b42b1d01ec6f093b95c296bd7235a1d96a94696507c).

Other changes are mostly a consequence of this, adapting to a slightly different interface, simplifying client code when it was straight forward, etc. *Many files have been changed, however, most of these are straightforward renamings or small changes to adapt to the new client.*

# Why this way

* Completely removing the dependency on `kafka-python` is out of scope for this PR, however there is only one admin client now
* The new `KafkaAdminClient` aims to keep the same interface as much as possible as it was before (exceptions include many-many `NewTopic.name` -> `NewTopic.topic` changes), also in exceptions thrown
* The underlying client from confluent-kafka-python takes care of API version negotiation (that was previously implemented in the admin client explicitly)
* When implementing and using the new admin client I aimed to contain all references to `confluent_kafka` within the new module
* Throwing meaningful errors still depends on kafka-python's `kafka.errors.for_code` and its specific exceptions
* ~No typing stubs are added for confluent-kafka-python, that's something that could potentially be done after a "full switch"~ Type stubs are going to be added in a follow-up: https://github.com/Aiven-Open/karapace/pull/760
* confluent-kafka-python is version pinned to 2.3.0 to avoid any kind of future incompatibilities without an explicit update

# Request from reviewers
Are there any "what if" kind of scenarios that might not be covered by tests or the code at the moment? (Especially backups and schema registry related.)


eliax1996 (commented):
Very nice code 😄, I've left some comments, feel free to disagree with all of them.
I've just left what I was feeling while reading the code. The code looks very auto explicative and you've added doc on all the new methods cool!

PPS before merging just make sure to sync the well doc you added to the commits with the pr description.

I need another round of reading but I've left this to flush my mental buffer :)

karapace/kafka_admin.py
```diff
@@ -0,0 +1,276 @@
...
+
+            if matches_name_filter or matches_source_filter:
+                config[config_name] = config_entry.value
+
+        return config
+
+    def get_offsets(self, topic: str, partition_id: int) -> dict[str, int]:
+        """Returns the beginning and end offsets for a topic partition.
+
+        Making two separate requests for beginning and end offsets, due to the
```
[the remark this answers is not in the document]
matyaskuti: I extended the docstring a bit

karapace/schema_reader.py
```diff
@@ -206,25 +195,28 @@ def run(self) -> None:
...
 
             schema_topic_exists = False
-            schema_topic = new_schema_topic_from_config(self.config)
-            schema_topic_create = [schema_topic]
             while not self._stop.is_set() and not schema_topic_exists:
                 try:
-                    LOG.info("[Schema Topic] Creating %r", schema_topic.name)
-                    self.admin_client.create_topics(schema_topic_create, timeout_ms=constants.TOPIC_CREATION_TIMEOUT_MS)
-                    LOG.info("[Schema Topic] Successfully created %r", schema_topic.name)
+                    LOG.info("[Schema Topic] Creating %r", self.config["topic_name"])
```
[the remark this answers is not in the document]
matyaskuti: I added logging in the admin client, but I'll still leave this here, I think it makes sense to leave these "domain specific" logs too.

karapace/kafka_admin.py
```diff
@@ -0,0 +1,276 @@
...
+        super().__init__(self._get_config_from_params(bootstrap_servers, **params))
+        self._activate_callbacks()
+        self._verify_connection()
+
+    @staticmethod
+    def _get_config_from_params(bootstrap_servers: Iterable[str] | str, **params: Unpack[AdminClientParams]) -> dict:
+        if not isinstance(bootstrap_servers, str):
+            bootstrap_servers = ",".join(bootstrap_servers)
+
+        config: dict[str, int | str | Callable | None] = {
```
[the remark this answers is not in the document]
eliax1996: Yeah, I don't really like dictionaries when you know statically the keys 😄 but I agree it's something minor that could be posticipated if you feel it's not the right time/could be heavily modified soon

karapace/schema_reader.py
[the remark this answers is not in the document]
eliax1996: Yeah agree

karapace/backup/topic_configurations.py
```diff
@@ -2,25 +2,12 @@
...
-    TOPIC_CONFIG = 1
-    DYNAMIC_BROKER_CONFIG = 2
-    DYNAMIC_DEFAULT_BROKER_CONFIG = 3
-    STATIC_BROKER_CONFIG = 4
-    DEFAULT_CONFIG = 5
-    DYNAMIC_BROKER_LOGGER_CONFIG = 6
-
-
-ALL_CONFIG_SOURCES: Final = {item.value for item in ConfigSource.__members__.values()}
+ALL_CONFIG_SOURCES: Final = ConfigSource.__members__.values()
```
aiven-anton, line 10:
```suggestion
ALL_CONFIG_SOURCES: Final = ConfigSource
```

nitpick (non-blocking): Code golfing 😉 This constant seems superfluous as enum classes are themselves valid iterables and containers of their members.

karapace/constants.py
```diff
@@ -8,7 +8,7 @@
 
 SCHEMA_TOPIC_NUM_PARTITIONS: Final = 1
 API_VERSION_AUTO_TIMEOUT_MS: Final = 30000
-TOPIC_CREATION_TIMEOUT_MS: Final = 20000
+TOPIC_CREATION_TIMEOUT_S: Final = 20
```
aiven-anton, line 11: suggestion (non-blocking): Since this is changing, I suggest using the unambiguous `datetime.timedelta` for this value.

tests/unit/test_authentication.py
```diff
@@ -107,6 +107,14 @@ def test_simple_oauth_token_provider_returns_configured_token() -> None:
     assert token_provider.token() == "TOKEN"
 
 
+def test_simple_oauth_token_provider_returns_configured_token_and_expiry() -> None:
```
aiven-anton, line 110: question (non-blocking): Is this related to this PR?

mypy.ini
```diff
@@ -89,6 +90,9 @@ ignore_errors = True
 [mypy-aiokafka.*]
 ignore_missing_imports = True
 
+[mypy-confluent_kafka.*]
+ignore_missing_imports = True
```
aiven-anton, line 94: issue: I get that it's a big ask, but I'd really prefer if we can stop introducing new ignores for dependencies. How big of a feat would it be to add stubs only for the parts we use, like we've done with isodate previously? 

https://github.com/Aiven-Open/karapace/tree/main/stubs/

karapace/kafka_admin.py
```diff
@@ -0,0 +1,290 @@
...
+
+def single_futmap_result(futmap: dict[Any, Future[T]]) -> T:
+    """Extract the result of a future wrapped in a dict.
+
+    Bulk operations of the `confluent_kafka` library's Kafka clients return results
+    wrapped in a dictionary of futures. Most often we use these bulk operations to
+    operate on a single resource/entity. This function makes sure the dictionary of
+    futures contains a single future and returns its result.
+    """
+    assert len(futmap) == 1
```
aiven-anton, line 44:
```suggestion
    operate on a single resource/entity. This function makes sure the dictionary of
    futures contains a single future and returns its result.
    """
    if not len(futmap) == 1:
        raise ValueError
```

suggestion: If it's part of the functions responsibility to "make sure the dictionary of futures contains a single future", I don't think it's appropriate with an assertion.

aiven-anton: (golfing) something would this could be a lazy way to make sure we raise when there's not exactly one item:

```python
(future,) = futmap.values()
return future.result()
```

karapace/kafka_admin.py
```diff
@@ -0,0 +1,290 @@
...
+    return next(iter(futmap.values())).result()
+
+
+def raise_from_kafkaexception(exc: KafkaException) -> NoReturn:
+    """Raises a more developer-friendly error from a `KafkaException`.
+
+    The confluent-kafka library's `KafkaException` is a wrapper around its internal
+    `KafkaError`. The resulting, raised exception however is coming from
+    `kafka-python`, due to these exceptions having human-readable names, providing
+    better context for error handling.
```
aiven-anton, line 55: issue: I'm weary about this, it looks like this will tie us into a future where we'll never be able to get rid of kafka-python, forever having to depend on both implementations.

karapace/kafka_admin.py
```diff
@@ -0,0 +1,290 @@
...
+        for _ in range(3):
+            try:
+                self.list_topics(timeout=1)
+            except KafkaException as exc:
+                LOG.info("Could not establish connection due to: %s", exc)
+                continue
+            else:
+                break
+        else:
+            raise AuthenticationFailedError()
```
aiven-anton, line 141: issue: This looks devious. There are other things than authentication that could be the cause of inability to connect, aren't we hiding that here?

karapace/kafka_admin.py
```diff
@@ -0,0 +1,290 @@
...
+            raise UnknownTopicOrPartitionError()
+
+        topics_data: dict[str, dict] = {}
+        for topic, topic_metadata in topics_metadata.items():
+            if topics is not None and topic not in topics:
+                continue
+
+            partitions_data = []
+            for partition_id, partition_metadata in topic_metadata.partitions.items():
+                partition_data = {
```
aiven-anton, line 206: issue: This looks like a place where it's not really justified to use dicts, let's use dataclasses with proper typing here?

karapace/kafka_admin.py
```diff
@@ -0,0 +1,290 @@
...
+            earliest_offset_futmap: dict[TopicPartition, Future] = self.list_offsets(
+                {
+                    TopicPartition(topic, partition_id): OffsetSpec.earliest(),
+                }
+            )
+            startoffset = single_futmap_result(earliest_offset_futmap)
+        except KafkaException as exc:
+            code = exc.args[0].code()
+            if code < 0:
+                raise UnknownTopicOrPartitionError() from exc
```
aiven-anton, line 288: issue: This looks off to me, the two codes that are `< 0` are `NONE` (no error) and `UNKNOWN_SERVER_ERROR`: https://kafka.apache.org/protocol.html#protocol_error_codes

aiven-anton (commented):
Leaving a partial review, I haven't read through everything yet. I left a few comments on the design/interface of the new admin client.

tests/unit/test_authentication.py
matyaskuti: Yes, the method tested here is used in the new admin client (slight difference between the expected "fetch token" interface).

karapace/kafka_admin.py
matyaskuti: I agree, however, I think overhauling the error handling here could be postponed until we are actually at the point when we can remove all usages of the `kafka-python` library. Until then, I think it's simpler to make use of this convenience, without having to adapt clients and wrap error-codes in meaningful exceptions, etc. (as not to have to check error-codes in client code).

karapace/kafka_admin.py
matyaskuti: Indeed, `-1` is the unknown error.
However, `confluent_kafka` has its own error codes, eg. in some cases this request would throw a `_NOENT` [KafkaError](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#kafkaerror) with the code `-156` and `"Failed to query partition leaders: No leaders found"` message.

I'll change this to explicitly check for that error code instead.

mypy.ini
matyaskuti: It wasn't a herculean effort to do it, I've added it in a separate commit. However, looking at the size of this PR I think it makes sense to split it off into its own follow-up PR.

karapace/kafka_admin.py
aiven-anton: That sounds fair to split up the effort 👍

karapace/kafka_admin.py
matyaskuti: I massaged this to be more or less the same as the old admin client when it comes to the exceptions raised. I'm not sure if we can get more detail out of the `confluent_kafka.admin.AdminClient`; even the error callback doesn't get much detailed information.

karapace/constants.py
matyaskuti: I'd leave this out for now, I think we can introduce timedeltas for these as a future improvement.

karapace/kafka_admin.py
matyaskuti: For now I wanted to have the usage of the admin client change as little as possible (there are already quite a lot of changes here) and just have an almost-drop-in-replacement (as much as possible). If you are okay with it, I'll leave a TODO here and tackle this as a follow-up.

mypy.ini
matyaskuti: Split it off into https://github.com/Aiven-Open/karapace/pull/760

aiven-anton: Nice! 👌

mypy.ini
```diff
@@ -14,6 +14,7 @@ warn_unused_ignores = True
 warn_no_return = True
 warn_unreachable = True
 strict_equality = True
+enable_incomplete_feature = Unpack
```
aiven-anton, line 17: I noticed this was just marked stable in 1.7 that was just released, oddly coincidental timing! https://mypy-lang.blogspot.com/2023/11/mypy-17-released.html

But let's keep this as-is and upgrade mypy separately to limit the scope.

aiven-anton (approved):

aiven-anton:
This looks good to me 👍

@eliax1996 Do you want to make another pass on this?

eliax1996 (approved):

PR #759, opened 2023-11-09
feat: enable users to disable validation only on specific topics

eliax1996:
This PR enables users to avoid validation only on specific topics. It is necessary because, prior to [this commit](ca65ddde04b1dd49cf4324ede66e2f5d827d3c51) it was possible to create messages with incorrect or missing subjects. By fixing the validation of subject flows that were previously functioning (but shouldn't have been) now no longer work due to the added validation.

With this PR, users can migrate their flows one by one while adding new validated flows, rather than opting out of validation completely if they wish to retain the functionality of old flows without refactoring them.

### DIFF
--- a/karapace/schema_registry.py
+++ b/karapace/schema_registry.py
@@ -27,9 +27,9 @@ from karapace.master_coordinator import MasterCoordinator
 from karapace.messaging import KarapaceProducer
 from karapace.offset_watcher import OffsetWatcher
 from karapace.schema_models import ParsedTypedSchema, SchemaType, SchemaVersion, TypedSchema, ValidatedTypedSchema
-from karapace.schema_reader import KafkaSchemaReader
+from karapace.schema_reader import KafkaSchemaReader, MessageType
 from karapace.schema_references import LatestVersionReference, Reference
-from karapace.typing import JsonObject, ResolvedVersion, SchemaId, Subject, Version
+from karapace.typing import JsonObject, ResolvedVersion, SchemaId, Subject, TopicName, Version
 from typing import Mapping, Sequence
 
 import asyncio
@@ -464,6 +464,19 @@ class KarapaceSchemaRegistry:
                 value["schemaType"] = schema.schema_type
         else:
             value = None
+        self.producer.send_message(key=key, value=value)
+
+    def is_topic_requiring_validation(self, *, topic_name: TopicName) -> bool:
+        return self.database.is_topic_requiring_validation(topic_name=topic_name)
+
+    def update_require_validation_for_topic(
+        self,
+        *,
+        topic_name: TopicName,
+        skip_validation: bool,
+    ) -> None:
+        key = {"topic": topic_name, "keytype": str(MessageType.schema_validation), "magic": 0}
+        value = {"skip_validation": skip_validation, "topic": topic_name}
         self.producer.send_message(key=key, value=value)
 
     def send_config_message(self, compatibility_level: CompatibilityModes, subject: Subject | None = None) -> None:
--- a/karapace/schema_registry.py
+++ b/karapace/schema_registry.py
@@ -476,7 +476,7 @@ class KarapaceSchemaRegistry:
         skip_validation: bool,
     ) -> None:
         key = {"topic": topic_name, "keytype": str(MessageType.schema_validation), "magic": 0}
-        value = {"skip_validation": skip_validation, "topic": topic_name}
+        value = {"skip_validation": skip_validation}
         self.producer.send_message(key=key, value=value)
 
     def send_config_message(self, compatibility_level: CompatibilityModes, subject: Subject | None = None) -> None:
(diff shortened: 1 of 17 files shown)


GSokol:
Thank you so much, @eliax1996! We are now struggling from disability of publishing new schemes for existing topics.

eliax1996:
@GSokol glad to know that this would help. I will create a new optional release in Aiven as soon as it is merged.

karapace/schema_registry.py
```diff
@@ -466,6 +466,19 @@ def send_schema_message(
...
+    def is_topic_requiring_validation(self, *, topic_name: TopicName) -> bool:
+        return self.database.is_topic_requiring_validation(topic_name=topic_name)
+
+    def update_require_validation_for_topic(
+        self,
+        *,
+        topic_name: TopicName,
+        skip_validation: bool,
+    ) -> None:
+        key = {"topic": topic_name, "keytype": str(MessageType.schema_validation), "magic": 0}
```
jjaakola-aiven, line 478: The formatting of the key needs to be aligned in the `key_format.py`. This record needs the canonical format, `topic` is an unknown key in data currently.
How does the Confluent Schema Registry react to custom message like this?

karapace/schema_registry_apis.py
```diff
@@ -28,20 +29,30 @@
...
+from karapace.typing import JsonData, JsonObject, ResolvedVersion, SchemaId, TopicName
 from karapace.utils import JSONDecodeError
 from typing import Any
+from urllib.parse import urlparse
 
 import aiohttp
 import async_timeout
 
 
+def compute_forwarded_url(master_url: str, request_url: str) -> str:
```
jjaakola-aiven, line 47: This would need a follow-up PR to change all other the forward URLs to use this.

jjaakola-aiven (commented):
Some comments, I need to do another round of reading and thinking.

karapace/schema_registry_apis.py
eliax1996: yeah, I thought about it but I didn't wanted to pollute this pr with other refactor

GSokol:
Also it's good to have a config setting, to disable schema validation by default. But it could be done in a separate PR to make this thing in faster.

eliax1996:
@GSokol this is already implemented (and merged and available in Aiven), if you look there is a config called `name_strategy_validation` that if set to `false` disable the validation. This PR is just to enable the users to explicitly disable the validation only on a specific subset of topics since using the validation brings more order in the topics (otherwise it's basically like Kafka vanilla where you can write anything in a topic)

GSokol:
@eliax1996, `name_strategy_validation` is only used in the REST API logic.

karapace/schema_registry.py
eliax1996: still a valid point, taking a look tomorrow

eliax1996: discussed offline, probably solved. Waiting an update on that

karapace/schema_registry_apis.py
eliax1996: Done in the last commit

tvainika:
Hi @GSokol, you wrote that

> `name_strategy_validation` is only used in the REST API logic.

and it made me curious about your expectations regarding this feature. This PR adds ability to disable validation per topic, but also this configuration affects only REST API logic. I hope this still matches your expectations.

karapace/schema_registry.py
```diff
@@ -466,6 +466,19 @@ def send_schema_message(
...
+
+    def update_require_validation_for_topic(
+        self,
+        *,
+        topic_name: TopicName,
+        skip_validation: bool,
+    ) -> None:
+        key = {"keytype": str(MessageType.schema_validation), "magic": 0}
+        value = {"skip_validation": skip_validation, "topic": topic_name}
+        self.producer.send_message(key=key, value=value)
```
tvainika, line 480: Depending on `skip_validation` this could also send tombstone record instead of toggling between true/false.

tvainika (changes requested):
Access control for new endpoint needs to be defined and documented.

Also new API endpoints need some documentation of semantics. For example if configuration `name_strategy_validation` is disabled, is it sill possible to enable validation for specific topics? With this code it seems that is not possible. Would it be useful, or not? Maybe per topic validation endpoints should reject setting if `name_strategy_validation` is disabled?

karapace/schema_registry.py
eliax1996: I just realized that I need to add the `topic` to the key, otherwise the compaction could delete the topic information, this doesn't change the behaviour of confluent schema registry that is simply to skip the messages he cannot understand (a behaviour that we also do while receiving messages that aren't parsable by us)

karapace/schema_registry.py
eliax1996: I don't think so, otherwise how can you rollback from disabling the validation for a certain topic?

PR #760, opened 2023-11-15
Add typing stubs for confluent-kafka

matyaskuti:

Building on https://github.com/Aiven-Open/karapace/pull/757, adding typing stubs for the confluent-kafka-python library.

# Why this way

Adding stubs only for the features we use from the library.


eliax1996 (approved):
Looked before and to me seems fine 😄
I didn't really looked every signature and every returned type in the real implementation though

aiven-anton (approved):

PR #761, opened 2023-11-20
Fix typing in protobuf enum and message elements

tvainika:

Fix typing in protobuf enum and message elements

### DIFF
--- a/karapace/protobuf/message_element.py
+++ b/karapace/protobuf/message_element.py
@@ -84,13 +84,14 @@ class MessageElement(TypeElement):
         result.append("}\n")
         return "".join(result)
 
-    def compare(self, other: MessageElement, result: CompareResult, types: CompareTypes) -> None:
+    def compare(self, other: TypeElement, result: CompareResult, types: CompareTypes) -> None:
         from karapace.protobuf.compare_type_lists import compare_type_lists
 
+        if not isinstance(other, MessageElement):
+            result.add_modification(Modification.TYPE_ALTER)
+            return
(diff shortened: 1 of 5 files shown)


karapace/protobuf/message_element.py
```diff
@@ -84,13 +84,14 @@ def to_schema(self) -> str:
         result.append("}\n")
         return "".join(result)
 
-    def compare(self, other: MessageElement, result: CompareResult, types: CompareTypes) -> None:
+    def compare(self, other: TypeElement, result: CompareResult, types: CompareTypes) -> None:
         from karapace.protobuf.compare_type_lists import compare_type_lists
 
+        if not isinstance(other, MessageElement):
+            result.add_modification(Modification.TYPE_ALTER)
```
eliax1996, line 91: question, which is the semantics of the compare? 

In case of multiple changes, like for e.g. changing type and adding a field to the message do we just return the `TYPE_ALTER`?

eliax1996 (approved):
Overall seems fine, just want to make sure that the return its always fine and trying to understand which is the hierarcy between the diff elements

karapace/protobuf/message_element.py
tvainika: This line of code should be mostly unreachable as this is invoked from `compare_type_lists` and there is basically already `if` statement which checks both `self` and `other` of `compare` call are same type. But for safety it is better to repeat that here instead of adding `assert` IMHO because this would be right thing to do.

PR #762, opened 2023-11-20
Rename internal variable to avoid overriding Thread._stop()

tvainika:

Rename internal variable to avoid overriding `Thread._stop()`

# Why this way

Don't override `Thread._stop()` with internal `Event()` instance.  This caused internal exception after process forking for producing ProtoBuf messages from REST Proxy.


eliax1996 (approved):
Great!

Issue #585, opened 2023-04-18
Proposal: first-class support for developing on Docker


aiven-anton:
I've opened a few draft PRs related to improving the Docker container, and wanted to share where I'm going with this. We decided to put those on hold until we have at least a rudimentary smoke test in place for the container image.

The purpose of this issue to serve as an umbrella for those and further changes, that will facilitate easier set up of development environments, especially for non-linux work stations. The end goal is to be able to easily run the test suite in a docker container, with local changes mounted for quick iterations. The draft PRs that are already lined-up also help in facilitating this by making sure local container rebuilds are quick.

In making these changes, I am also modernizing some of the tooling that we're using where appropriate, like using Docker's Github actions to build container metadata and versioning, using a more recent Python version, etc.

It's important to note that all of these changes are intended to only have minor effects on the resulting production container image.

### Items to address 

- [x] [Introduce smoke-test of container image](https://github.com/aiven/karapace/pull/597). Should start up container in both modes (`rest`/`registry`) with a Kafka back-end, and verify communications work through some relatively simple operations.
- [x] [Build ARM image](https://github.com/aiven/karapace/pull/578) (fixes #521).
- [x] [Optimize container build time](https://github.com/aiven/karapace/pull/579).
- [x] ~[#580](https://github.com/aiven/karapace/pull/580)~ [Narrow docker context](https://github.com/aiven/karapace/pull/579), this assures changes to unrelated files in the repository don't trigger cache eviction of Docker layers.
- [ ] Introduce a `dev` build target, separate from the production build target, that installs full test requirements in the container.
- [ ] Make sure there's a docker compose file that uses the new `dev` target, should enable running tests with something like `docker compose run --rm karapace pytest`.
- [ ] Introduce proxy targets in Makefile, so that we can easily run common make targets within a container. `make docker:unit-tests` should result in a call to `make unit-tests` within the docker-compose container.

### Why?

- Current setup leaves lingering processes on developer machines.
- No need to install, and keep versions in sync, of dependency binaries on developer machines (java, protoc, snappy, kafka).
- Alignment of developer environments, possibly easier to reason about and recreate issues.
- A streamlined dev env setup hopefully means it's easier to attract contributors.

muralibasani:
With smoke tests and docker setup files, I assume this issue can be closed.

@aiven-anton Pls reopen if we are missing something.

PR #763, opened 2023-11-24
fix: fixed various stuff that were blocking a new out of the box installation only using the `GNUmakefile`

eliax1996:
fixed errors that I've encountered by cloning the project from scratch in a new folder and using the `GNUmakefile` to set it up and running.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -86,9 +86,9 @@ cleanest: cleaner
 requirements: export CUSTOM_COMPILE_COMMAND='make requirements'
 requirements:
 	pip install --upgrade pip setuptools pip-tools
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements.in
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-dev.in
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-typing.in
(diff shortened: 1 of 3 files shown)


GNUmakefile
```diff
@@ -86,9 +86,9 @@ cleanest: cleaner
 requirements: export CUSTOM_COMPILE_COMMAND='make requirements'
 requirements:
 	pip install --upgrade pip setuptools pip-tools
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements.in
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-dev.in
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-typing.in
+	pip-compile --upgrade --resolver=backtracking requirements/requirements.in -o requirements/requirements.txt
```
aiven-anton, line 89: Note: the reason this breaks is because the use of `.ONESHELL:`. That feature doesn't exist in the version of make that ships with macs, resulting in diverging behaviors.

Once we have a [more complete docker dev env](https://github.com/Aiven-Open/karapace/issues/585), I'll argue for removing much of the makefile targets that downloads executables and sets up venvs and so on.

aiven-anton (approved):

PR #765, opened 2023-11-28
Replace sync Kafka Producers with confluent_kafka one

matyaskuti:

* Introduce a synchronous, `conlfuent_kafka`-based producer client
* Extract and split off common code from the [previously introduced](https://github.com/Aiven-Open/karapace/pull/757) `KafkaAdminClient`
* Create `karapace.kafka` to contain all Kafka client code
* Still reuse `kafka-python`'s exception classes
* Many of the changed files are only import and slight interface adjustments

# Why this way

* Removing explicit references to `kafka-python` in the code can be done after sync consumers are replaced with `confluent-kafka`-based ones too, as well as removing the reliance on `kafka-python`'s exceptions


karapace/kafka/admin.py
matyaskuti, line 1: This is moved from the pre-existing `karapace/kafka_admin.py`, with some of the code moved to `karapace/kafka/common.py`

karapace/kafka_admin.py
matyaskuti, line 1: Moved into karapace/kafka/admin.py and karapace/kafka/common.py

stubs/confluent_kafka/admin/__init__.pyi
```diff
@@ -15,13 +15,14 @@ __all__ = (
...
     "OffsetSpec",
+    "PartitionMetadata",
     "ResourceType",
     "TopicMetadata",
 )
 
 class AdminClient:
     def __init__(self, config: dict[str, str | int | Callable]) -> None: ...
-    def poll(self, timeout: float) -> int: ...
+    def poll(self, timeout: float = -1) -> int: ...
```
aiven-anton, line 25:
```suggestion
    def poll(self, timeout: float = ...) -> int: ...
```

nit (non-blocking): Default values have no effect in type hints, and are usually omitted to eliminate duplication ([see last paragraph of this section](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading) for reference).

aiven-anton (approved):
LGTM, nice work 👍

karapace/backup/api.py
```diff
@@ -388,11 +390,11 @@ def _handle_producer_send(
             key=instruction.key,
             value=instruction.value,
             partition=instruction.partition_index,
-            headers=[(key.decode() if key is not None else None, value) for key, value in instruction.headers],
-            timestamp_ms=instruction.timestamp,
-        ).add_errback(producer_error_callback)
-    except (KafkaError, AssertionError) as ex:
-        raise BackupDataRestorationError("Error while calling send on restoring messages") from ex
+            headers=[(key.decode(), value) for key, value in instruction.headers if key is not None],
```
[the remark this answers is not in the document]
eliax1996: why doing that instead of asserting that isn't `None`?
In that way if the assumptions isn't true we are skipping a record, IMO we should fail if an assumption is violated rather than proving that currently isn't violated and if the implementation changes in the future we have a skip instead of an error

PR #767, opened 2023-11-30
Replace sync Kafka consumers with confluent_kafka one 

matyaskuti:

* Implement and use a synchronous confluent-kafka based consumer client
* Building on the previous confluent-kafka changes: https://github.com/Aiven-Open/karapace/pull/765 and https://github.com/Aiven-Open/karapace/pull/757
* The new client is a drop-in replacement as much as possible, usage is slightly adjusted (hence the large number of files changed)
* PyTest timeout from 60s to 90s, since the default poll timeout for backups is 60s
* `karapace.kafka.types` module introduced to consolidate additional data types used from confluent-kafka (to a certain extent, see [comment thread](https://github.com/Aiven-Open/karapace/pull/767#discussion_r1423847487))

# Why this way

* Continuing the previous patterns towards removing kafka-python (errors and `KarapaceKafkaClient` still remain)
* ~Keeping confluent-kafka references in `karapace.kafka.*`~ this was [agreed](https://github.com/Aiven-Open/karapace/pull/767#discussion_r1423847487) to be simplified and not be a goal after all


karapace/schema_reader.py
```diff
@@ -306,49 +302,52 @@ def handle_messages(self) -> None:
             if are_we_master is True:
                 watch_offsets = True
 
-        for _, msgs in raw_msgs.items():
-            schema_records_processed_keymode_canonical = 0
-            schema_records_processed_keymode_deprecated_karapace = 0
-            for msg in msgs:
+        schema_records_processed_keymode_canonical = 0
```
matyaskuti, line 305: One level less looping, as the new client doesn't return messages separated by partition.

tests/unit/backup/backends/test_v2.py
```diff
@@ -29,7 +30,7 @@ def test_schema_backup_v2_roundtrip(tmp_path: Path) -> None:
     topic_name = "a-topic"
     partition_index = 123
     records = (
-        ConsumerRecord(
+        StubMessage(
```
matyaskuti, line 33: See `test.utils.StubMessage`, as `confluent_kafka.Message` cannot be user-instantiated

karapace/schema_reader.py
```diff
@@ -234,11 +231,8 @@ def _get_beginning_offset(self) -> int:
         assert self.consumer is not None, "Thread must be started"
 
         try:
-            offsets = self.consumer.beginning_offsets([TopicPartition(self.config["topic_name"], 0)])
-            # Offset in the response is the offset for last offset.
-            # Reduce by one for matching on startup.
-            beginning_offset = list(offsets.values())[0] - 1
-            return beginning_offset
+            beginning_offset, _ = self.consumer.get_watermark_offsets(TopicPartition(self.config["topic_name"], 0))
+            return beginning_offset - 1
```
matyaskuti, line 235: Should this be reduced by 1? 🤔

karapace/backup/api.py
```diff
@@ -282,9 +281,8 @@ def _consume_records(
     consumer: KafkaConsumer,
     topic_partition: TopicPartition,
     poll_timeout: PollTimeout,
-) -> Iterator[ConsumerRecord]:
-    start_offset: int = consumer.beginning_offsets([topic_partition])[topic_partition]
-    end_offset: int = consumer.end_offsets([topic_partition])[topic_partition]
+) -> Iterator[Message]:
+    start_offset, end_offset = consumer.get_watermark_offsets(topic_partition)
```
eliax1996, line 285: this could also return `None`, right?

eliax1996: yeah I think so:
```
static PyObject *Consumer_get_watermark_offsets (Handle *self, PyObject *args,
                                                 PyObject *kwargs) {

        TopicPartition *tp;
        rd_kafka_resp_err_t err;
        double tmout = -1.0f;
        int cached = 0;
        int64_t low = RD_KAFKA_OFFSET_INVALID, high = RD_KAFKA_OFFSET_INVALID;
        static char *kws[] = { "partition", "timeout", "cached", NULL };
        PyObject *rtup;

        if (!self->rk) {
                PyErr_SetString(PyExc_RuntimeError,
                                "Consumer closed");
                return NULL;
        }

        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|db", kws,
                                         (PyObject **)&tp, &tmout, &cached))
                return NULL;


        if (PyObject_Type((PyObject *)tp) != (PyObject *)&TopicPartitionType) {
                PyErr_Format(PyExc_TypeError,
                             "expected %s", TopicPartitionType.tp_name);
                return NULL;
        }

        if (cached) {
                err = rd_kafka_get_watermark_offsets(self->rk,
                                                     tp->topic, tp->partition,
                                                     &low, &high);
        } else {
                Py_BEGIN_ALLOW_THREADS;
                err = rd_kafka_query_watermark_offsets(self->rk,
                                                       tp->topic, tp->partition,
                                                       &low, &high,
                                                       cfl_timeout_ms(tmout));
                Py_END_ALLOW_THREADS;
        }

        if (err) {
                cfl_PyErr_Format(err,
                                 "Failed to get watermark offsets: %s",
                                 rd_kafka_err2str(err));
                return NULL;
        }

        rtup = PyTuple_New(2);
        PyTuple_SetItem(rtup, 0, PyLong_FromLongLong(low));
        PyTuple_SetItem(rtup, 1, PyLong_FromLongLong(high));

        return rtup;
}
```

eliax1996:
```
if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|db", kws,
                                         (PyObject **)&tp, &tmout, &cached))
                return NULL;
```

> Parse the parameters of a function that takes both positional and keyword parameters into local variables. The keywords argument is a NULL-terminated array of keyword parameter names. Empty names denote [positional-only parameters](https://docs.python.org/3/glossary.html#positional-only-parameter). Returns true on success; on failure, it returns false and raises the appropriate exception.

Ah ok, its only when something is making an invalid call. Weird but we should never hit this

karapace/kafka/consumer.py
```diff
@@ -0,0 +1,44 @@
...
+        if "group_id" not in params:
+            params["group_id"] = self._create_group_id()
+
+        super().__init__(bootstrap_servers, verify_connection, **params)
+
+        self.subscribe([topic])
+
+    @staticmethod
+    def _create_group_id() -> str:
+        return f"karapace-{secrets.token_hex(6)}"
```
eliax1996, line 37: optional: maybe `karapace-autogenerated-{secrets.token_hex(6)}`? Just to highlight that its automatically generated because of missing an explicit one?

karapace/schema_reader.py
eliax1996: weird, I cannot understand why trivially. If you remove the `-1` and run again the tests some of them are failing?

eliax1996: also we need to check that for an empty topic returns 0 and from a topic with 1 message return always 0

karapace/schema_reader.py
```diff
@@ -263,7 +257,7 @@ def _is_ready(self) -> bool:
             return False
         # Offset in the response is the offset for the next upcoming message.
         # Reduce by one for actual highest offset.
-        self._highest_offset = list(offsets.values())[0] - 1
+        self._highest_offset = end_offset - 1
```
eliax1996, line 260: same here, probably its worth checking if its really right and how its handled for an empty topic

karapace/schema_reader.py
```diff
@@ -306,49 +300,52 @@ def handle_messages(self) -> None:
...
+        schema_records_processed_keymode_canonical = 0
+        schema_records_processed_keymode_deprecated_karapace = 0
+        for msg in msgs:
+            try:
+                message_key = msg.key()
+                if message_key is None:
+                    continue
+                key = json_decode(message_key)
+            except JSONDecodeError:
+                LOG.exception("Invalid JSON in msg.key()")
```
eliax1996, line 312: maybe logging also the message (truncated if its very big) could be helpful?

karapace/schema_reader.py
```diff
@@ -306,49 +300,52 @@ def handle_messages(self) -> None:
...
+
+            value = None
+            message_value = msg.value()
+            if message_value:
                 try:
-                    key = json_decode(msg.key)
+                    value = self._parse_message_value(message_value)
                 except JSONDecodeError:
-                    LOG.exception("Invalid JSON in msg.key")
+                    LOG.exception("Invalid JSON in msg.value()")
```
eliax1996, line 331: and here?

eliax1996 (commented):
Almost convinced its fine, just want to discuss a little bit on the doubt I've asked 😁, when @aiven-anton take a look I'm fine with merging

karapace/backup/api.py
aiven-anton: It is documented as returning `None` for timeouts: https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.get_watermark_offsets.

We should probably handle that explicitly.

karapace/schema_reader.py
matyaskuti: Having or not having the `-1` there makes no difference in the tests, interestingly.
I added some tests for the consumer for the purpose of documenting this behaviour - `(0, 0)` is returned on an empty topic by `get_watermark_offsets`, so I think the `-1`s are correct here

aiven-anton (commented):
Partial review so far, leaving some comments.

karapace/backup/api.py
```diff
@@ -301,12 +299,11 @@ def _consume_records(
     end_offset -= 1  # high watermark to actual end offset
 
     while True:
-        records: Collection[ConsumerRecord] = consumer.poll(poll_timeout.milliseconds).get(topic_partition, [])
-        if len(records) == 0:
+        record: Message | None = consumer.poll(timeout=poll_timeout.seconds)
+        if record is None:
```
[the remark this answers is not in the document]
matyaskuti: I'm not entirely sure that in this case a stale consumer equals an error while polling. A `None` record in this case means no messages within the timeout, thus `StaleConsumerError`. I think if there is a `record.error()` present, we should raise that error separately - it rather means something went wrong than "we ran out of data to back up". 🤔 
What do you think?

karapace/backup/backends/v3/backend.py
```diff
@@ -334,27 +334,31 @@ def store_metadata(
...
-                key=record.key,
-                value=record.value,
-                headers=tuple(Header(key=key.encode(), value=value) for key, value in record.headers),
-                offset=record.offset,
-                timestamp=record.timestamp,
+                key=record_key.encode() if isinstance(record_key, str) else record_key,
+                value=record_value.encode() if isinstance(record_value, str) else record_value,
+                headers=tuple(Header(key=key.encode(), value=value) for key, value in record.headers() or []),
+                offset=record.offset(),
+                timestamp=record.timestamp()[1],
```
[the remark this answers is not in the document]
matyaskuti: I'd say we are not losing any information, we are ignoring the same information as before. In `ConsumerRecord` the timestamp type is represented as a separate field: https://github.com/dpkp/kafka-python/blob/0864817de97549ad71e7bc2432c53108c5806cf1/kafka/consumer/fetcher.py#L31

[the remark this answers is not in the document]
aiven-anton: Yeah that's what I meant, the change reveals that our implementation loses a piece of information here (both before and after).

stubs/confluent_kafka/cimpl.pyi
```diff
@@ -56,3 +62,18 @@ class Producer:
...
     def list_topics(self, topic: str | None = None, timeout: float = -1) -> ClusterMetadata: ...
     def poll(self, timeout: float = -1) -> int: ...
+
+class Consumer:
+    def subscribe(self, topics: list[str]) -> None: ...
+    def get_watermark_offsets(
+        self, partition: TopicPartition, timeout: float | None = None, cached: bool = False
+    ) -> tuple[int, int]: ...
+    def close(self) -> None: ...
+    def list_topics(self, topic: str | None = None, timeout: float = -1) -> ClusterMetadata: ...
```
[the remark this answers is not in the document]
matyaskuti: I'll clean all this up in a follow-up 👍

karapace/backup/backends/v3/backend.py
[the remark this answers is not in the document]
matyaskuti: Create an issue for this: https://github.com/Aiven-Open/karapace/issues/779

karapace/backup/api.py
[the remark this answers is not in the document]
matyaskuti: I'll make the change to raise any errors that a message might have, making sure it bubbles up from here.

As for the CLI behaviour, we have the option of not catching it, just letting it fail, although this is not the nicest.
We could also catch all Kafka errors (I don't see a simple way of narrowing it down at the moment), and print a generic error message (containing/highlighting the error code and its description) along the stacktrace.

karapace/kafka/types.py
```diff
@@ -0,0 +1,55 @@
...
+    OffsetSpec,
+    PartitionMetadata,
+    ResourceType,
+    TopicMetadata,
+)
+from confluent_kafka.error import KafkaError, KafkaException
+
+import enum
+
+__all__ = (
```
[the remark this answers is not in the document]
matyaskuti: I'm trying to make `karapace.kafka` as self-contained as possible and avoiding `confluent_kafka` references leaking anywhere else.

Having `kafka` and `aiokafka` spread all over the codebase has been a pain in gradually replacing them. Also if/when we add/use more functionality from `confluent_kafka` it's possible that we'll need to wrap some of its classes with our own to enhance them - it's going to be easier to do so if things are in one place.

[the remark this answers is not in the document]
eliax1996: I like that idea

karapace/schema_reader.py
matyaskuti: I've added the offset to the log message, as you mentioned large keys (or values below) could be tricky. `LOG.exception` includes the error, that should already help with figuring out what went wrong.

karapace/schema_reader.py
matyaskuti: Same as above.

karapace/kafka/types.py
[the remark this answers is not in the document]
aiven-anton: I don't agree with this, and I doubt the reason that kafka-python and aiokafka are hard to replace is because we import directly from them rather from an aliasing namespace like this. In fact, I think it would have been harder because in order to follow the pattern during the transitioning phase, entities from both third-party libraries would have been forced to compete for names in the same namespace.

In order for this to actually have the intended effect, we would need to introduce our own abstraction layer and explicitly implement adapters for each backend. As we don't plan to support multiple backends in the long term, I don't really see any value in this though and I think it's OK that we tie ourselves into the libraries we're using, even if it makes it a bit harder to replace them.

As there are two voices in the opposite direction from mine here, I see where this is leaning. Please consider this one more round and then mark this as resolved if that's where you're still landing 🙂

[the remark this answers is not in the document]
matyaskuti: Thanks for challenging this! I'll remove the re-exports and change back to the previous state where we use `confluent_kafka` imports throughout.

While I do prefer re-exports here and hiding `confluent_kafka` elsewhere (from a structural point of view), I prefer future-proofing more. Given this structure, using types and classes from `confluent_kafka` in other places in the codebase would have to be a 2-step process always - slightly inconvenient with the inevitable mixing of the two styles as direct imports from the library is the path of least resistance. 👍 

Will resolve the comment once the change is done.

tests/unit/backup/backends/v3/test_backend.py
```diff
@@ -33,33 +33,23 @@ def test_writer_reader_roundtrip(tmp_path: Path) -> None:
...
             partition=partition_index,
             offset=10,
-            timestamp=round(time.time()),
-            timestamp_type=None,
-            headers=(),
-            checksum=None,
-            serialized_key_size=None,
-            serialized_value_size=None,
-            serialized_header_size=None,
+            timestamp=(1, round(time.time())),  # TODO
```
aiven-anton, line 42: nit (non-blocking): Are these todos for the timestamp type? Maybe better to remove them since we have a ticket now? Otherwise should probably have more context. (2 more in same file)

tests/utils.py
```diff
@@ -307,3 +307,17 @@ def popen_karapace_all(config_path: Union[Path, str], stdout: IO, stderr: IO, **
...
+
+    Since that class cannot be instantiated, thus this is a liberal simulation
+    of its behaviour ie. its attributes are accessible via getter functions:
+    `message.offset()`."""
+
+    def __init__(self, **attrs: Any) -> None:
+        self._attrs = attrs
+
+    def __getattr__(self, key: str) -> None:
+        return lambda: self._attrs[key]
```
aiven-anton, line 323: nit (non-blocking): I assume this is safe, but given code-base will evolve, would it be safer to re-raise as `AttributeError` on a failing access here?

karapace/schema_reader.py
aiven-anton: Nice! 👍

aiven-anton (commented):
Ok, fully reviewed now, I think looks good as far as I can tell. The outstanding thing is the questions about LEO/HWM off by one ... Can we find some way to build higher confidence in those changes?

karapace/schema_reader.py
eliax1996: I would say that we should be able to distinguish between an empty topic and a topic with 1 message. IMO should give back 2 different reply

tests/unit/backup/backends/v3/test_backend.py
matyaskuti: Good catch, I missed these after introducing `Timestamp`

karapace/schema_reader.py
matyaskuti: As discussed separately with @aiven-anton and @jjaakola-aiven, this offset decrement is a result of the schema reader and offset watcher evolution and has no real effect towards behaviour (unlike decrementing the end offset). See also `OFFSET_UNINITIALIZED` and `OFFSET_EMPTY` towards the top of this file. This and the two constants can be eventually cleaned up and simplified.

Related explanation in the backups code: https://github.com/Aiven-Open/karapace/pull/781

karapace/schema_reader.py
matyaskuti: Related explanation in the backups code: https://github.com/Aiven-Open/karapace/pull/781

karapace/backup/api.py
```diff
@@ -282,9 +282,8 @@ def _consume_records(
     consumer: KafkaConsumer,
     topic_partition: TopicPartition,
     poll_timeout: PollTimeout,
-) -> Iterator[ConsumerRecord]:
-    start_offset: int = consumer.beginning_offsets([topic_partition])[topic_partition]
-    end_offset: int = consumer.end_offsets([topic_partition])[topic_partition]
+) -> Iterator[Message]:
+    start_offset, end_offset = consumer.get_watermark_offsets(topic_partition)
```
aiven-anton, line 286: I had another doubt this was correct due to mix of terminology, but have now verified it's the same underlying Protocol message involved (`ListOffsets`):

- https://github.com/confluentinc/librdkafka/blob/df6efd25e6e44c06a937cbdf8fc81715668e0ce1/src/rdkafka.c#L3597
- https://github.com/dpkp/kafka-python/blob/0864817de97549ad71e7bc2432c53108c5806cf1/kafka/consumer/fetcher.py#L573

So looks correct 👍

aiven-anton (dismissed):
LGTM, solid work! 👍

karapace/schema_reader.py
eliax1996: question, not requirement. Do you think would be useful to paste the explaination also here? I feel that its something that could save time to someone wondering why in the future

matyaskuti: @eliax1996 I added a comment here

eliax1996 (approved):

PR #775, opened 2023-12-06
Implement async confluent-kafka producer

matyaskuti:
* Create a wrapper around the existing, confluent-kafka based `KafkaProducer` with an async `send` method and a poll-thread to send messages continuously in the background
* Dropping the usage of an SSL context for the async producer, as confluent-kafka has no support for this
* Change the REST web application shutdown hook from `on_cleanup` to `on_shutdown` - this was an existing bug preventing any kind of closing of resources (Kafka clients, consumer manager, etc.) to actually run, further explanation in commit message

# Why this way
* (Mostly) drop-in replacement for existing behaviour
* Following the [confluent-kafka example](https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/asyncio_example.py)

### DIFF
--- a/karapace/kafka/common.py
+++ b/karapace/kafka/common.py
@@ -71,9 +71,12 @@ class TokenWithExpiryProvider(Protocol):
 
 
 class KafkaClientParams(TypedDict, total=False):
+    acks: int | None
     client_id: str | None
     connections_max_idle_ms: int | None
-    max_block_ms: int | None
+    compression_type: str | None
+    linger_ms: int | None
+    message_max_bytes: int | None
     metadata_max_age_ms: int | None
     retries: int | None
     sasl_mechanism: str | None
@@ -83,6 +86,7 @@ class KafkaClientParams(TypedDict, total=False):
     socket_timeout_ms: int | None
     ssl_cafile: str | None
     ssl_certfile: str | None
+    ssl_crlfile: str | None
     ssl_keyfile: str | None
     sasl_oauth_token_provider: TokenWithExpiryProvider
     # Consumer-only
@@ -121,8 +125,12 @@ class _KafkaConfigMixin:
 
         config: dict[str, int | str | Callable | None] = {
             "bootstrap.servers": bootstrap_servers,
+            "acks": params.get("acks"),
             "client.id": params.get("client_id"),
             "connections.max.idle.ms": params.get("connections_max_idle_ms"),
+            "compression.type": params.get("compression_type"),
+            "linger.ms": params.get("linger_ms"),
+            "message.max.bytes": params.get("message_max_bytes"),
             "metadata.max.age.ms": params.get("metadata_max_age_ms"),
             "retries": params.get("retries"),
             "sasl.mechanism": params.get("sasl_mechanism"),
@@ -132,6 +140,7 @@ class _KafkaConfigMixin:
             "socket.timeout.ms": params.get("socket_timeout_ms"),
             "ssl.ca.location": params.get("ssl_cafile"),
             "ssl.certificate.location": params.get("ssl_certfile"),
+            "ssl.crl.location": params.get("ssl_crlfile"),
             "ssl.key.location": params.get("ssl_keyfile"),
             "error_cb": self._error_callback,
             # Consumer-only
--- a/karapace/kafka/producer.py
+++ b/karapace/kafka/producer.py
@@ -5,15 +5,18 @@ See LICENSE for details
 
 from __future__ import annotations
 
+from collections.abc import Iterable
 from concurrent.futures import Future
 from confluent_kafka import Message, Producer
 from confluent_kafka.admin import PartitionMetadata
 from confluent_kafka.error import KafkaError, KafkaException
 from functools import partial
-from karapace.kafka.common import _KafkaConfigMixin, raise_from_kafkaexception, translate_from_kafkaerror
+from karapace.kafka.common import _KafkaConfigMixin, KafkaClientParams, raise_from_kafkaexception, translate_from_kafkaerror
+from threading import Event, Thread
 from typing import cast, TypedDict
 from typing_extensions import Unpack
 
+import asyncio
 import logging
 
 LOG = logging.getLogger(__name__)
@@ -59,3 +62,70 @@ class KafkaProducer(_KafkaConfigMixin, Producer):
             return self.list_topics(topic).topics[topic].partitions
         except KafkaException as exc:
             raise_from_kafkaexception(exc)
+
+
+class AsyncKafkaProducer:
+    """An async wrapper around `KafkaProducer` built on top of confluent-kafka.
+
+    Calling `start` on an `AsyncKafkaProducer` instantiates a `KafkaProducer`
+    and starts a poll-thread.
+
+    The poll-thread continuously polls the underlying producer so buffered messages
+    are sent and asyncio futures returned by the `send` method can be awaited.
+    """
+
+    def __init__(
+        self,
+        bootstrap_servers: Iterable[str] | str,
+        loop: asyncio.AbstractEventLoop | None = None,
+        **params: Unpack[KafkaClientParams],
+    ) -> None:
+        self.loop = loop or asyncio.get_running_loop()
+
+        self.stopped = Event()
+        self.poll_thread = Thread(target=self.poll_loop)
+
+        self.producer: KafkaProducer | None = None
+        self._bootstrap_servers = bootstrap_servers
+        self._producer_params = params
+
+    def _start(self) -> None:
+        assert not self.stopped.is_set(), "The async producer cannot be restarted"
+
+        self.producer = KafkaProducer(self._bootstrap_servers, **self._producer_params)
+        self.poll_thread.start()
+
+    async def start(self) -> None:
+        # The `KafkaProducer` instantiation tries to establish a connection with
+        # retries, thus can block for a relatively long time. Running in the
+        # default executor and awaiting makes it async compatible.
+        await self.loop.run_in_executor(None, self._start)
+
+    def _stop(self) -> None:
+        self.stopped.set()
+        if self.poll_thread.is_alive():
+            self.poll_thread.join()
+        self.producer = None
+
+    async def stop(self) -> None:
+        # Running all actions needed to stop in the default executor, since
+        # some can be blocking.
+        await self.loop.run_in_executor(None, self._stop)
+
+    def poll_loop(self) -> None:
+        """Target of the poll-thread."""
+        assert self.producer is not None, "The async producer must be started"
+
+        while not self.stopped.is_set():
+            # The call to `poll` is blocking, necessitating running this loop in its own thread.
+            # In case there is messages to be sent, this loop will do just that (equivalent to
+            # a `flush` call), otherwise it'll sleep for the given timeout (seconds).
+            self.producer.poll(timeout=0.1)
+
+    async def send(self, topic: str, **params: Unpack[ProducerSendParams]) -> asyncio.Future[Message]:
+        assert self.producer is not None, "The async producer must be started"
+
+        return asyncio.wrap_future(
+            self.producer.send(topic, **params),
+            loop=self.loop,
+        )
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -1,5 +1,3 @@
-from aiokafka import AIOKafkaProducer
-from aiokafka.errors import KafkaConnectionError
 from binascii import Error as B64DecodeError
 from collections import namedtuple
 from confluent_kafka.error import KafkaException
@@ -13,9 +11,10 @@ from kafka.errors import (
     TopicAuthorizationFailedError,
     UnknownTopicOrPartitionError,
 )
-from karapace.config import Config, create_client_ssl_context
+from karapace.config import Config
 from karapace.errors import InvalidSchema
 from karapace.kafka.admin import KafkaAdminClient
+from karapace.kafka.producer import AsyncKafkaProducer
 from karapace.kafka_rest_apis.authentication import (
     get_auth_config_from_header,
     get_expiration_time_from_header,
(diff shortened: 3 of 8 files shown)


tests/integration/test_rest.py
```diff
@@ -227,9 +227,9 @@ async def test_internal(rest_async: KafkaRest | None, admin_client: KafkaAdminCl
     assert len(results) == 1
     for result in results:
         assert "error" in result, "Invalid result missing 'error' key"
-        assert result["error"] == "Unrecognized partition"
+        assert result["error"] == "This request is for a topic or partition that does not exist on this broker."
         assert "error_code" in result, "Invalid result missing 'error_code' key"
-        assert result["error_code"] == 1
```
matyaskuti, line 232: `aiokafka` raises an `AssertionError` in this case, which is considered non-retriable, however from `confluent-kafka` an `UnknownTopicOrPartitionError` equivalent is raised, which is considered retriable - hence the difference in the error codes.

I'm not entirely sure which behaviour we'd like to go forward with, the "correct" one with code `2`, or preserve the previous. 🤔

karapace/kafka_rest_apis/authentication.py
```diff
@@ -142,6 +142,9 @@ class SimpleOauthTokenProviderAsync(AbstractTokenProviderAsync):
     async def token(self) -> str:
         return self._token
 
+    def token_with_expiry(self, _config: str | None = None) -> tuple[str, int | None]:
```
matyaskuti, line 145: This is necessary for now, as we still have the consumers from `aiokafka` present. Once that's replaced, we can simplify these token providers, to have a single one with this method, that doesn't really depend on the abstract classes from any library. (confluent-kafka only needs a callable as the OAuth callback)

karapace/kafka/common.py
```diff
@@ -71,9 +71,12 @@ def token_with_expiry(self, config: str | None) -> tuple[str, int | None]:
 
 
 class KafkaClientParams(TypedDict, total=False):
+    acks: int | Literal["all"] | None
```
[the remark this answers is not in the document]
matyaskuti: To keep things simpler, I changed it to `int | None` only and adjusted the usage in `kafka_rest_apis/__init__.py` so we use `-1` when the config has `"all"`.

The actual value could be anything between ("all")-1..1000

karapace/kafka/producer.py
```diff
@@ -59,3 +62,41 @@ def partitions_for(self, topic: str) -> dict[int, PartitionMetadata]:
...
+
+        self.stopped = Event()
+        self.poll_thread = Thread(target=self.poll_loop)
+        self.poll_thread.start()
+
+    async def stop(self) -> None:
+        self.stopped.set()
+        self.poll_thread.join()
+
+    def poll_loop(self) -> None:
```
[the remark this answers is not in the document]
matyaskuti: Yes, but also for any messages to be sent from the local buffer `poll` or `flush` (which is just `poll` in a loop until the buffer is empty) needs to be called. To achieve `asyncio` compatible behaviour then, `poll` needs to be called constantly. Otherwise the client would have to take care of this, which doesn't fit into the whole async setup.

eliax1996 (commented):
I feel confident in merging once you reply to my doubts :)

aiven-anton:
> Dropping the usage of an SSL context for the async producer, as confluent-kafka has no support for this

Just to make sure this doesn't become a future blocker, it could perhaps be worth experimenting with handling SSL external to confluent-kafka, and making it communicate through an already setup TLS connection.

karapace/kafka/producer.py
```diff
@@ -59,3 +62,41 @@ def partitions_for(self, topic: str) -> dict[int, PartitionMetadata]:
...
+
+    Starting `AIOKafkaProducer` instantiates a `KafkaProducer` and starts a poll-thread.
+
+    The poll-thread continuously polls the underlying producer so buffered messages
+    are sent and asyncio futures returned by the `send` method can be awaited.
+    """
+
+    def __init__(
+        self,
+        bootstrap_servers: Iterable[str] | str,
```
aiven-anton, line 78: thought (non-blocking): Just noting that `str` is a subtype of `Iterable[str]`, so technically we could safely replace with just `Iterable[str]`, but let's not do that ;)

On a tangent, if we wanted to be super stringent here, we could bring in this construct to model "sequence, but not a `str`": https://github.com/hauntsaninja/useful_types/blob/951d36b2aaaee6ab9fa1efcb8c4f1e2733247ff5/useful_types/__init__.py#L285.

karapace/kafka/producer.py
```diff
@@ -59,3 +62,41 @@ def partitions_for(self, topic: str) -> dict[int, PartitionMetadata]:
...
+        loop: asyncio.AbstractEventLoop | None = None,
+        **params: Unpack[KafkaClientParams],
+    ) -> None:
+        self.loop = loop or asyncio.get_running_loop()
+        self.producer = KafkaProducer(bootstrap_servers, **params)
+
+        self.stopped = Event()
+        self.poll_thread = Thread(target=self.poll_loop)
+        self.poll_thread.start()
```
aiven-anton, line 87: issue: Consider not starting thread at instantiation time, as it's usually preferable for instantiation to not have side effects.

karapace/kafka/producer.py
```diff
@@ -59,3 +62,41 @@ def partitions_for(self, topic: str) -> dict[int, PartitionMetadata]:
...
+
+    async def stop(self) -> None:
+        self.stopped.set()
+        self.poll_thread.join()
+
+    def poll_loop(self) -> None:
+        """Target of the poll-thread."""
+        while not self.stopped.is_set():
+            self.producer.poll(0.1)
```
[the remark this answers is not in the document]
matyaskuti: `poll` is blocking, that's why it needs its own thread.
In case there is messages to be sent it'll just keep doing that (equivalent behaviour to [calling `flush`](https://github.com/confluentinc/librdkafka/blob/e03d3bb91ed92a38f38d9806b8d8deffe78a1de5/src/rdkafka.c#L4379-L4387)). If there are no messages to be sent, this call will block for `0.1` second, making sleep superfluous.

tests/integration/kafka/test_producer.py
```diff
@@ -71,3 +74,58 @@ def test_partitions_for(self, producer: KafkaProducer, new_topic: NewTopic) -> N
         assert partitions[0].id == 0
         assert partitions[0].replicas == [1]
         assert partitions[0].isrs == [1]
+
+
+@pytest.fixture(scope="function", name="aioproducer")
+async def fixture_aioproducer(
```
aiven-anton, line 80: suggestion (non-blocking): Same here as Elia's comment `s/aioproducer/async_producer/`?

karapace/kafka/producer.py
```diff
@@ -59,3 +62,41 @@ def partitions_for(self, topic: str) -> dict[int, PartitionMetadata]:
             return self.list_topics(topic).topics[topic].partitions
         except KafkaException as exc:
             raise_from_kafkaexception(exc)
+
+
+class AIOKafkaProducer:
```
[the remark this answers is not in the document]
matyaskuti: I've no strong opinion on this, but do you perhaps have an alternative suggestion? 🤔

tests/integration/kafka/test_producer.py
```diff
@@ -71,3 +74,58 @@ def test_partitions_for(self, producer: KafkaProducer, new_topic: NewTopic) -> N
...
+    kafka_servers: KafkaServers,
+    loop: asyncio.AbstractEventLoop,
+) -> Iterator[AIOKafkaProducer]:
+    try:
+        aioproducer = AIOKafkaProducer(bootstrap_servers=kafka_servers.bootstrap_servers, loop=loop)
+        yield aioproducer
+    finally:
+        await aioproducer.stop()
```
aiven-anton, line 88:
```suggestion
    aioproducer = AIOKafkaProducer(bootstrap_servers=kafka_servers.bootstrap_servers, loop=loop)
    try:
        yield aioproducer
    finally:
        await aioproducer.stop()
```

issue: `aioproducer` is not necessarilly successfully assigned to in the `finally` clause unless it happens prior to entering the try-block. This might be input to the other comment about not starting the background thread during instantiation.

tests/integration/kafka/test_producer.py
```diff
@@ -71,3 +74,58 @@ def test_partitions_for(self, producer: KafkaProducer, new_topic: NewTopic) -> N
...
+        )
+        message = await aiofut
+
+        assert message.offset() == 0
+        assert message.partition() == partition
+        assert message.topic() == new_topic.topic
+        assert message.key() == key
+        assert message.value() == value
+        assert message.timestamp()[0] == Timestamp.CREATE_TIME
+        assert message.timestamp()[1] == timestamp
```
aiven-anton, line 115: suggestion (non-blocking): Should we check the headers?

karapace/kafka/producer.py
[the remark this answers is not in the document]
aiven-anton: I agree with this, I'd probably go with `AsyncKafkaProducer`.

karapace/kafka/producer.py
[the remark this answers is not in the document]
aiven-anton: This would actually be a good comment to add to this line IMO :)

tests/integration/kafka/test_producer.py
matyaskuti: `send` uses the `confluent_kafka.Producer.produce` method, which has a bit of a caveat [returning headers on a Message](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Producer.produce):

> Currently message headers are not supported on the message returned to the callback. The msg.headers() will return None even if the original message had headers set.

karapace/kafka/producer.py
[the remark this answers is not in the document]
eliax1996: > If there are no messages to be sent, this call will block for 0.1 second, making sleep superfluous.

yeah my naive question was more like: its this busy wait or its smart enough to block. But not that I think about it its fine, your your responsibility :)
I agree that the comment could be useful to have!

tests/integration/kafka/test_producer.py
aiven-anton: Gotcha ✅

karapace/kafka/producer.py
aiven-anton: The extra hoops could be mostly designed away though, by making the class a context manager. I.e. making `__enter__` call `self.start()` and `__exit__` call `.stop()`.

IMO it's bad practice to do things that require clean-up at instantiation time.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -477,27 +477,24 @@ async def _maybe_create_async_producer(self) -> AIOKafkaProducer:
...
-                    linger_ms=self.config["producer_linger_ms"],
-                    max_request_size=self.config["producer_max_request_size"],
-                    metadata_max_age_ms=self.config["metadata_max_age_ms"],
-                    security_protocol=self.config["security_protocol"],
-                    ssl_context=ssl_context,
-                    **get_kafka_client_auth_parameters_from_config(self.config),
-                )
-
                 try:
+                    producer = AsyncKafkaProducer(
```
aiven-anton, line 481: suggestion: It looks like this try-block aught to have a `finally: producer.stop()`?

aiven-anton: Now, this makes me wonder whether this is the cause of these too many open files errors we have in sentry:

https://aiven-io.sentry.io/issues/4625989357/?environment=production&project=1219632&query=is%3Aunresolved+assigned%3A%23team-helpful-husky&referrer=issue-stream&statsPeriod=30d&stream_index=2

aiven-anton, line 481: suggestion: Now that the instantiation cannot raise `(NoBrokersAvaiable, AuthenticationFailedError)` anymore, let's move it outside the try-block.

matyaskuti: For now I've only moved starting the thread, instantiating the underlying producer is not in `start` yet, work in progress.

matyaskuti: > suggestion: It looks like this try-block aught to have a finally: producer.stop()?
I think if we do that every time we start the producer we would also stop it right away. We could call stop in the except clause, but if the underlying `KafkaProducer` cannot be created the thread will not be started, so calling stop would do nothing.

If I read that Sentry error correctly it's around admin client creation in Karapace version 3.9 (although maybe it's present for other versions but Sentry is not very helpful there) 🤔

aiven-anton: Ah yes, you're right, it's only the exception cases we should close it.

> but if the underlying KafkaProducer cannot be created the thread will not be started, so calling stop would do nothing

I don't think we can say much about which resources were successfully created if an exception occurs here. For the caught ones, it definitely looks like we can have opened a socket that would be abandoned and lingering if not cleaned up, but same goes for any exception not explicitly handled.

karapace/kafka/producer.py
```diff
@@ -59,3 +62,69 @@ def partitions_for(self, topic: str) -> dict[int, PartitionMetadata]:
...
+        self.stopped = Event()
+        self.poll_thread = Thread(target=self.poll_loop)
+
+        self.producer: KafkaProducer | None = None
+        self._bootstrap_servers = bootstrap_servers
+        self._producer_params = params
+
+    def _start(self) -> None:
+        self.producer = KafkaProducer(self._bootstrap_servers, **self._producer_params)
+        self.stopped.clear()
```
aiven-anton, line 94: nit (non-blocking): Should we really allow restarting rather than erroring out, it seems some resource management will get complex if we allow restarting, and I believe the subsequent call to `self.poll_thread.start()` will currently fail because it tries restart an existing thread? Not sure we need to make something explicit, but this `.clear()` call could potentially be removed, and for safe usage it might make sense to raise if the event is set when entering `_start()`?

aiven-anton (approved):
LGTM, I added a nit-level comment, but I think this is mergable as-is. Nice work!

karapace/kafka/producer.py
matyaskuti: Good point about the thread restart, adding a guard here.

aiven-anton (approved):

eliax1996 (approved):

jjaakola-aiven (approved):

PR #777, opened 2023-12-06
Do not skip None keyed headers on restore

matyaskuti:

* Previously in https://github.com/Aiven-Open/karapace/pull/765 skipping None header keys was added. While now it's not causing any problems, in the future it could be an issue if we suppress/skip error-cases instead of explicit failure. Highlighted by https://github.com/Aiven-Open/karapace/pull/765/files/b606514604e32349d52ed757f5e0a2ad02851452#r1417211308
*  Fully return to pre https://github.com/Aiven-Open/karapace/pull/765 behaviour in the backups api
* As confluent-kafka also allows `None` header keys, despite the documentation not mentioning this, adjust typing

# Why this way

* Restoring previous behaviour


eliax1996 (approved):
Thanks for this and for didn't indulge my paranoia :)

PR #782, opened 2023-12-19
Implement async confluent-kafka consumer

matyaskuti:

* Create a wrapper around the existing `KafkaConsumer` with each operation ran in the default event loop executor
* Dropping the usage of an SSL context for the async consumer, as confluent-kafka has no support for this
* Adjust the consumer manager accordingly, simplifying where possible

# Why this way

* (Mostly) drop-in replacement for existing behaviour

Follow-ups:

* OAuth token handling cleanup: https://github.com/Aiven-Open/karapace/pull/800
* Removing `aiokafka` from dependencies is in https://github.com/Aiven-Open/karapace/pull/798


karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -486,7 +489,11 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
                 )
                 timeout_left = max(0, (start_time - time.monotonic()) * 1000 + timeout)
                 try:
-                    data = await consumer.getmany(timeout_ms=timeout_left, max_records=1)
+                    message = await consumer.poll(timeout=timeout_left / 1000)
+                    if message is None:
+                        continue
+                    if message.error() is not None:
+                        raise translate_from_kafkaerror(message.error())
```
matyaskuti, line 496: Errors can also surface this way from confluent-kafka, it's a good practice to check for `message.error()`.

However, in case of a topic regex/pattern subscription here we throw an `UnkownTopicOrPartitionError` (subclass of `KafkaError` that is handled below with an HTTP 500) if there is no topic discovered yet that matches the regex.

How fast such a topic gets discovered is dependent on `topic.metadata.refresh.interval.ms`, which is 5 minutes by default. A message with such an error could show up multiple times, thus potentially resulting us dropping (ie. not returning) messages, due to how we immediately raise on exceptions (also later in case of deserialization errors).

Should we perhaps extend the response with an `"error"` in L534 and not throw any exceptions in this case? Also since this error is considered retriable. 🤔

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -495,47 +506,44 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
...
-                    element = {
-                        "topic": tp.topic,
-                        "partition": tp.partition,
-                        "offset": msg.offset,
-                        "timestamp": msg.timestamp,
-                        "key": key,
-                        "value": value,
-                    }
-                    response.append(element)
+            for msg in poll_data:
```
matyaskuti, line 521: One less layer of looping, records are not grouped by partition.

karapace/kafka/consumer.py
```diff
@@ -66,3 +69,209 @@ def get_watermark_offsets(
             return result
         except KafkaException as exc:
             raise_from_kafkaexception(exc)
+
+    def commit(
+        self,
+        message: Message | None = None,
+        offsets: list[TopicPartition] | None = None,
+        _asynchronous: bool = False,
```
eliax1996, line 77: why this field if its not used?

karapace/kafka/consumer.py
```diff
@@ -66,3 +69,209 @@ def get_watermark_offsets(
...
+
+        More in the confluent-kafka documentation:
+        https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.subscribe
+        """
+        topics = topics or []
+        patterns = patterns or []
+        self.log.info("Subscribing to topics %s and patterns %s", topics, patterns)
+        try:
+            super().subscribe(
+                topics + [f"^{pattern}" for pattern in patterns],
```
eliax1996, line 126: maybe `.lstrip("^")` or raise if its starting with `^`? Just wondering what happens if someone use it badly

karapace/kafka/consumer.py
```diff
@@ -66,3 +69,209 @@ def get_watermark_offsets(
...
+        "^this-is-a-regex-[0-9]", thus we prefix all strings in the `patterns`
+        list with "^".
+
+        The `on_assign` and `on_revoke` callbacks are set to keep track of
+        subscriptions (topics).
+
+        More in the confluent-kafka documentation:
+        https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.subscribe
+        """
+        topics = topics or []
```
eliax1996, line 121: ```assert self._subscription == frozenset(), "expected to call `unsubscribe` before attempting a new subscription"```?

karapace/kafka/consumer.py
```diff
@@ -66,3 +69,209 @@ def get_watermark_offsets(
...
+
+    def _on_assign(self, _consumer: Consumer, partitions: list[TopicPartition]) -> None:
+        topics = frozenset(partition.topic for partition in partitions)
+        self._subscription = self._subscription.union(topics)
+
+    def _on_revoke(self, _consumer: Consumer, partitions: list[TopicPartition]) -> None:
+        topics = frozenset(partition.topic for partition in partitions)
+        self._subscription = self._subscription.difference(topics)
+
+    def subscription(self) -> frozenset[str]:
```
eliax1996, line 145: `subscriptions`?

eliax1996: not sure if its enforced by the interface, in that case simply ignore it :)

karapace/kafka/consumer.py
eliax1996: maybe better raise some error

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -495,47 +505,44 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
...
-                    for rec in records:
-                        message_count += 1
-                        read_bytes += max(0, 0 if rec.key is None else len(rec.key)) + max(
-                            0, 0 if rec.value is None else len(rec.value)
-                        )
-                        poll_data[topic].append(rec)
-                        read_buffered = True
+                message_count += 1
+                read_bytes += max(0, 0 if message.key() is None else len(message.key())) + max(
+                    0, 0 if message.value() is None else len(message.value())
```
eliax1996, line 510: why max between 0 and the `len()`, could it deliver a negative number? and when if yes?

eliax1996: idem for the other max, could a sum between a `len` and another `max` between 0 and another `len` be negative?

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -495,47 +505,44 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
...
+                    KarapaceBase.unprocessable_entity(
+                        message=f"value deserialization error for format {request_format}: {e}",
+                        sub_code=RESTErrorCodes.HTTP_UNPROCESSABLE_ENTITY.value,
+                        content_type=content_type,
+                    )
+                element = {
+                    "topic": msg.topic(),
+                    "partition": msg.partition(),
+                    "offset": msg.offset(),
+                    "timestamp": msg.timestamp()[1],
```
eliax1996, line 541: mmm the `[1]` is non trivial, could you add a comment like:
`msg.timestamp()[1] # returns a tuple of message timestamp type, and timestamp we are interested only on the latter.`

Also we don't need to parse the type of timing and do different translation based on the type of time?
Should we maybe hide this detail behind a logic and avoid to forwarding that weird type outside of your wrapper?

tests/integration/test_rest_consumer.py
```diff
@@ -87,7 +88,7 @@ async def test_subscription(rest_async_client, admin_client, producer, trail):
     topic_name = new_topic(admin_client)
     instance_id = await new_consumer(rest_async_client, group_name, fmt="binary", trail=trail)
     sub_path = f"/consumers/{group_name}/instances/{instance_id}/subscription{trail}"
-    consume_path = f"/consumers/{group_name}/instances/{instance_id}/records{trail}?timeout=1000"
```
eliax1996, line 90: interesting, why do we want to increase the default timeout? Have you noticed that the performance are degratating or there is another rationale behind this?

eliax1996 (changes requested):
First round done :)
I would like to take another look later.

Great job you almost moved entirely karapace out of the other libraries 😁 piece after piece you have done a very great and impactful job! 🧨

tests/integration/test_rest_consumer.py
matyaskuti: Because of the topic-pattern subscription later in this test we need to give the consumer a bit more time to discover everything and actually get assigned the topics.

karapace/kafka/consumer.py
matyaskuti: I think the expected behaviour and previous behaviour from the other libraries is that you can re-subscribe but that will override existing subscriptions (just like here).

An `IllegalStateError` is raised if `assign` is called after `subscribe` but I think this should be fine. 🤔

eliax1996: mmm lets just add a log of warning in that case :)

karapace/kafka_rest_apis/consumer_manager.py
eliax1996: I would also explictly check that the `[0]` element its the value expected :)

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -495,47 +505,49 @@ async def fetch(self, internal_name: Tuple[str, str], content_type: str, formats
...
+                element = {
+                    "topic": msg.topic(),
+                    "partition": msg.partition(),
+                    "offset": msg.offset(),
+                    # `confluent_kafka.Message.timestamp()` returns a tuple where the first component is
+                    # the timestamp type, see `karapace.kafka.types.Timestamp`
+                    # In case of the `NOT_AVAILABLE` type whatever the timestamp may be, it cannot be trusted
+                    # and should be ignored according to the confluent-kafka documentation:
+                    # https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/#confluent_kafka.Message
+                    "timestamp": msg.timestamp()[1] if msg.timestamp()[0] != Timestamp.NOT_AVAILABLE else None,
```
eliax1996, line 546: interesting, seems we can distinguish between when the timestamp was attached to the record from the Broker or from the Producer. I'm wondering if we can add that information in the element itself (to not lose an important information), but probably its an API that should stay constant. I'm wondering if we can add a property to the returned fetched message itself, like two properties that could be `None` or populated based on this result (like: `def produced_timestamp()` and `def broker_received_timestamp()`.

eliax1996 (dismissed):
LGTM

karapace/kafka_rest_apis/consumer_manager.py
matyaskuti: Indeed, that's available in the producer API: https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Producer.produce

In this case I think it might have less significance, since when producing via Karapace, we don't have a timestamp option, so it's always going to be from the broker.

jjaakola-aiven (approved):

PR #794, opened 2024-01-10
Replace sync Kafka consumers with confluent_kafka one

matyaskuti:

* The exact same PR as https://github.com/Aiven-Open/karapace/pull/767 (which has been reverted in https://github.com/Aiven-Open/karapace/pull/793)
* With a single line removed from backups creation (_assigning_ a specific partition to a consumer after _subscribing_ to the same topic): `karapace/backup/api.py` L533 removed, see the second commit in this PR

# Why this way

* Using `consumer.assign` after `consumer.subscribe` is not supposed to be done (further tests and guards will be introduced in a later [PR](https://github.com/Aiven-Open/karapace/pull/782/files#diff-1efd38e0519f1bead1ab4fe0996f01ed062e167700963b12d90f30ba50be74c3) which adds enriched functionality to consumers)


matyaskuti:
@jjaakola-aiven I updated the description with the location. After a short discussion with @aiven-anton I'll attempt to add a test that reveals this issue through the backups code as well.

tests/integration/backup/test_v3_backup.py
```diff
@@ -1043,3 +1046,125 @@ def test_gives_non_successful_exit_code_for_legacy_backup_format(
         assert cp.returncode == 1
         assert cp.stderr.decode() == error_message
         assert cp.stdout == b""
+
+
+def test_backup_creation_fails_with_consumer_subscribe_assign(
```
aiven-anton, line 1051: While this looks useful to pin-point the issue, as this is not a feature, I don't think it makes sense to check this test in.

What I meant to say with the need for a regression test is rather we should have a test that fails when the fix is not in place, and that passes when the fix is there. However, that should just be checked manually and we should not keep around a test to check that we're failing when the code is patched back into the bad version.

tests/integration/kafka/test_consumer.py
```diff
@@ -77,3 +78,46 @@ def test_get_watermark_offsets_topic_with_one_message(
 
         assert beginning == 0
         assert end == 1
+
+
+def test_assign_after_subscribe_consumes_from_beginning(
```
aiven-anton, line 83: Same comment applies here, this looks useful to pinpoint the issue, but I don't think we should add this test to the code-base.

tests/integration/backup/test_v3_backup.py
```diff
@@ -1043,3 +1046,125 @@ def test_gives_non_successful_exit_code_for_legacy_backup_format(
...
+            api.create_backup(
+                config=config,
+                backup_location=backup_location,
+                topic_name=api.normalize_topic_name(None, config),
+                version=BackupVersion.V3,
+                replication_factor=1,
+            )
+
+
+def test_backup_creation_succeeds_without_consumer_assign(
```
aiven-anton, line 1111: suggestion: Let's add some comment here to make it more obvious why this test is added:

"This test was added to test for regressions after ___ ..."

"The issue facilitates only for large topics because ___"

aiven-anton (changes requested):

tests/integration/backup/test_v3_backup.py
matyaskuti: Fair enough. Although my thinking was more along the lines of documenting behaviour with executable proof, rather than testing a feature. Nevertheless, I'll remove the 2 extra tests. 👍

PR #797, opened 2024-01-19
Fix schema registering with ordering change

### LINKED ISSUE #795: Schema not getting register if the order of types are changed.
# What happened?

With compatibility set to full transitive, If I try to update the proto and just change the order of message and enum type in the file, the schema registry fails to registry the schema.

For example, I have the following proto, and I registry it in the schema registry.
```
FUGA_PROTO = """
syntax = "proto3";
package tc4;

message Fred {
  HodoCode hodecode = 1;
}

enum HodoCode {
  HODO_CODE_UNSPECIFIED = 0;
}
"""
```

Now If I update the proto (adding 1 field in message _Fred_) and change the order of message and enum shown below (Note I have written enum first and message type later).
```
FUGA_UPDATED_PROTO = """
syntax = "proto3";
package tc4;

enum HodoCode {
  HODO_CODE_UNSPECIFIED = 0;
}

message Fred {
  HodoCode hodecode = 1;
  string id = 2;
}
"""
```

The registry fails to registry(or update) the schema with the following error.
```
  File "/venv/lib/python3.10/site-packages/karapace/rapu.py", line 329, in _handle_request
    data = await callback(**callback_kwargs)
  File "/venv/lib/python3.10/site-packages/karapace/schema_registry_apis.py", line 1240, in subject_post
    raise xx
  File "/venv/lib/python3.10/site-packages/karapace/schema_registry_apis.py", line 1207, in subject_post
    schema_id = await self.schema_registry.write_new_schema_local(subject, new_schema, references)
  File "/venv/lib/python3.10/site-packages/karapace/schema_registry.py", line 393, in write_new_schema_local
    result = check_compatibility(
  File "/venv/lib/python3.10/site-packages/karapace/compatibility/__init__.py", line 175, in check_compatibility
    result = check_protobuf_compatibility(
  File "/venv/lib/python3.10/site-packages/karapace/compatibility/__init__.py", line 68, in check_protobuf_compatibility
    return check_protobuf_schema_compatibility(reader, writer)
  File "/venv/lib/python3.10/site-packages/karapace/compatibility/protobuf/checks.py", line 12, in check_protobuf_schema_compatibility
    writer.compare(reader, result)
  File "/venv/lib/python3.10/site-packages/karapace/protobuf/schema.py", line 585, in compare
    return self.proto_file_element.compare(
  File "/venv/lib/python3.10/site-packages/karapace/protobuf/proto_file_element.py", line 147, in compare
    return compare_type_lists(self.types, other.types, result, compare_types)
  File "/venv/lib/python3.10/site-packages/karapace/protobuf/compare_type_lists.py", line 72, in compare_type_lists
    raise IllegalStateException("Instance of element is not applicable")
karapace.protobuf.exception.IllegalStateException: Instance of element is not applicable
```

Note: If I do not change the order of enum and message type, the code work as expected.

The expected behavior should be that the registry should not fail and it should register the schema.

I tested the same scenario in the confluent registry and it is working fine.

tvainika:

Fix schema compatibility checks if ordering of types is changed. Order of types does not affect semantical compatibility of the schemas.

Fixes #795

### DIFF
--- a/tests/unit/protobuf/test_protobuf_compatibility.py
+++ b/tests/unit/protobuf/test_protobuf_compatibility.py
@@ -2,7 +2,7 @@
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
-from karapace.protobuf.compare_result import CompareResult
+from karapace.protobuf.compare_result import CompareResult, Modification
 from karapace.protobuf.kotlin_wrapper import trim_margin
 from karapace.protobuf.location import Location
 from karapace.protobuf.proto_file_element import ProtoFileElement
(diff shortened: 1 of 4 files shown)


tests/unit/protobuf/test_protobuf_compatibility.py
```diff
@@ -188,3 +188,115 @@ def test_compatibility_enum_add():
...
+"""
+
+    self_parsed: ProtoFileElement = ProtoParser.parse(location, self_schema)
+    other_parsed: ProtoFileElement = ProtoParser.parse(location, other_schema)
+
+    result = CompareResult()
+    self_parsed.compare(other_parsed, result)
+    assert result.is_compatible()
+    assert len(result.result) == 1
+    assert result.result[0].modification == Modification.FIELD_ADD
```
eliax1996, line 264: Why should it be considered an addition of a field?

eliax1996:
Shouldn’t we sort the fully path types and then compare those? I mean, isn’t easier if we normalize the schema before comparing it and compare it only when it’s compared?
Just asking, not complaining of the current solution

tests/unit/protobuf/test_protobuf_compatibility.py
tvainika: `string id = 2` field is assed to message `Fred`.

eliax1996: ops I’ve confused the two tests. Looks good!

eliax1996 (approved):

tvainika:
> Shouldn’t we sort the fully path types and then compare those? I mean, isn’t easier if we normalize the schema before comparing it and compare it only when it’s compared?

Karapace does not support `normalize` request parameter same as CSR. I think those kind of cases should be addressed, but yet change of type order should not be considered incompatible change. So those could be sorted internally in memory before this comparing, but I think it is simply more readable code if there is no need for such internal step.

eliax1996:
> > Shouldn’t we sort the fully path types and then compare those? I mean, isn’t easier if we normalize the schema before comparing it and compare it only when it’s compared?
> 
> Karapace does not support `normalize` request parameter same as CSR. I think those kind of cases should be addressed, but yet change of type order should not be considered incompatible change. So those could be sorted internally in memory before this comparing, but I think it is simply more readable code if there is no need for such internal step.

Got it. Thanks for the reply :)

PR #800, opened 2024-01-24
OAuth token prodiver cleanup

matyaskuti:

Since we moved away from kafka-python and aiokafka based clients, there is no need to depend on either library's abstract OAuth token provider mechanism. The confluent-kafka clients take a simple callable for token fetching (pass-through in our case).

Rest of the `aiokafka` related cleanup is in https://github.com/Aiven-Open/karapace/pull/798


eliax1996 (approved):
image

PR #801, opened 2024-01-26
Handle UnknownTopicOrPartitionError in schema reader

aiven-anton:

Handle `UnknownTopicOrPartitionError` in `._is_ready()` check after seeing this in production.

### Why this way

All exceptions should be handled.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -18,6 +18,7 @@ from kafka.errors import (
     NoBrokersAvailable,
     NodeNotReadyError,
     TopicAlreadyExistsError,
+    UnknownTopicOrPartitionError,
 )
 from karapace import constants
 from karapace.config import Config
@@ -240,6 +241,8 @@ class KafkaSchemaReader(Thread):
             return beginning_offset - 1
         except KafkaTimeoutError:
             LOG.warning("Reading begin offsets timed out.")
+        except UnknownTopicOrPartitionError:
+            LOG.warning("Topic does not yet exist.")
         except Exception as e:  # pylint: disable=broad-except
             self.stats.unexpected_exception(ex=e, where="_get_beginning_offset")
             LOG.exception("Unexpected exception when reading begin offsets.")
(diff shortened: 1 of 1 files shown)


karapace/schema_reader.py
```diff
@@ -256,6 +257,9 @@ def _is_ready(self) -> bool:
         except KafkaTimeoutError:
             LOG.exception("Reading end offsets timed out.")
             return False
+        except UnknownTopicOrPartitionError:
```
matyaskuti, line 260: This would also have to be added in [`_get_beginning _offset`](https://github.com/Aiven-Open/karapace/pull/801/files#diff-d0d95f66bd9730341ddb7c386650f61768378934401a70d05ec9dac690634ea8R235)

Furthermore I think we should switch the [consumer creation](https://github.com/Aiven-Open/karapace/pull/801/files#diff-d0d95f66bd9730341ddb7c386650f61768378934401a70d05ec9dac690634ea8R178-R193) and the [topic creation](https://github.com/Aiven-Open/karapace/pull/801/files#diff-d0d95f66bd9730341ddb7c386650f61768378934401a70d05ec9dac690634ea8R195-R214) code - now we create a consumer and subscribe it to a topic that's not been ensured to exist yet, causing these errors to be thrown and retried again and again until the topic is actually available.

matyaskuti (changes requested):

karapace/schema_reader.py
aiven-anton: Both these sounds like good things to do. However, I'm just trying to deal with the unhandled exception here, we're seeing quite some large volumes of these at times. I suggest merging this as-is and doing any more involved work as follow-up.

matyaskuti: Makes sense, but let's handle the same exception in the other function as well (as these 2 are related)

aiven-anton: Ok, added equivalent handling to the other function as well :)

matyaskuti (approved):

PR #802, opened 2024-01-26
Demote various exception logs to warning and info

aiven-anton:
Exception level logging should only be used for cases where there's an actual application error, but in many cases was being used for recoverable situations and during normal operations and user errors. The right place to use exception level logging is when the exception is unknown, not just when any Python exception has occurred.

In some situations exceptions were logged before being re-raised, those were demoted to info logs since there's no value in having duplicated logging.

Overall, exception handling is not structured throughout Karapace and we at some point need to do a deep inventory. In many places we catch broad exceptions where we really should not, and this is likely having bad effects on behavior during memory constrained operations and similar.

### DIFF
--- a/karapace/kafka_rest_apis/__init__.py
+++ b/karapace/kafka_rest_apis/__init__.py
@@ -307,7 +307,7 @@ class KafkaRest(KarapaceBase):
                     if self.proxies.get(key) is None:
                         self.proxies[key] = UserRestProxy(self.config, self.kafka_timeout, self.serializer)
             except (NoBrokersAvailable, AuthenticationFailedError):
-                log.exception("Failed to connect to Kafka with the credentials")
+                log.warning("Failed to connect to Kafka with the credentials")
                 self.r(body={"message": "Forbidden"}, content_type=JSON_CONTENT_TYPE, status=HTTPStatus.FORBIDDEN)
             proxy = self.proxies[key]
             proxy.mark_used()
@@ -497,9 +497,9 @@ class UserRestProxy:
                 except (NoBrokersAvailable, AuthenticationFailedError):
                     await producer.stop()
                     if retry:
-                        log.exception("Unable to connect to the bootstrap servers, retrying")
+                        log.warning("Unable to connect to the bootstrap servers, retrying")
                     else:
-                        log.exception("Giving up after trying to connect to the bootstrap servers")
+                        log.warning("Giving up after trying to connect to the bootstrap servers")
                         raise
                     await asyncio.sleep(1)
                 except Exception:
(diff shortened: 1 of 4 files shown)


karapace/kafka_rest_apis/__init__.py
```diff
@@ -653,9 +653,9 @@ def init_admin_client(self):
                 break
             except:  # pylint: disable=bare-except
                 if retry:
-                    log.exception("Unable to start admin client, retrying")
+                    log.warning("Unable to start admin client, retrying")
                 else:
-                    log.exception("Giving up after failing to start admin client")
+                    log.info("Giving up after failing to start admin client")
```
tvainika, line 658: If retry is true, log level should be lower (info) than when retry is false (last try failing). Switch info and warning between each other here?

karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -228,9 +228,9 @@ async def create_kafka_consumer(self, fetch_min_bytes, group_name, internal_name
                 return c
             except:  # pylint: disable=bare-except
                 if retry:
-                    LOG.exception("Unable to create consumer, retrying")
+                    LOG.warning("Unable to create consumer, retrying")
                 else:
-                    LOG.exception("Giving up after failing to create consumer")
+                    LOG.info("Giving up after failing to create consumer")
```
tvainika, line 233: Same here.

karapace/kafka_rest_apis/__init__.py
```diff
@@ -497,9 +497,9 @@ async def _maybe_create_async_producer(self) -> AsyncKafkaProducer:
                 except (NoBrokersAvailable, AuthenticationFailedError):
                     await producer.stop()
                     if retry:
-                        log.exception("Unable to connect to the bootstrap servers, retrying")
+                        log.warning("Unable to connect to the bootstrap servers, retrying")
                     else:
-                        log.exception("Giving up after trying to connect to the bootstrap servers")
+                        log.info("Giving up after trying to connect to the bootstrap servers")
```
tvainika, line 502: If retry is true, log level should be lower (info) than when retry is false (last try failing). Switch info and warning between each other here?

tvainika (changes requested):
Looks good overall but some warning/info level log lines don't match in my logic.

karapace/kafka_rest_apis/__init__.py
aiven-anton: My reasoning for `info` here is because the exception is re-raised, I actually think we shouldn't log at all here due to the re-raise. I used debug at first, but went with `info` to be less invasive to behavior.

Not a very strong opinion, and I can adjust to warning.

aiven-anton: Updated to use warning in both places.

tvainika (approved):

PR #806, opened 2024-01-31
Graceful handling of broker response errors

tvainika:

BrokerResponseErrors include meaningful reason and message (e.g. topic authorization failed), which can be passed to caller of Kafka REST api here, so it is not an error in Karapace, and should not be logged as such.


aiven-anton (approved):
Looks good: Kafka returning error != application error in Karapace 👍

PR #813, opened 2024-02-08
protobuf: improve protoc compatibility

juha-aiven:

Fix a bug in _deserialize_field that causes deserialization fail in certain cases, e.g. in the case there's a nested field that is a message or an enum. The issue is that the type is optional if type_name is specified. The code falsely expects type to always be present.

Add test test_protoc_serialized_schema_deserialize to test deserialization of protoc-serialized schemas. Test a few cases that are problematic, e.g. one with an enum and another with a nested message.

Better error handling if type and type_name are missing. Raise InvalidSchema instead of using assert.

# Why this way

It would be possible to do the testing in different ways. 

1. The test would generate using protoc the "protoc" Python module (for each tested schema)
2. The test would use the generated code to serialize the tested schema
3. The test would use the just created serialized schema

The difference is that in this PR the serialized schema is hard-coded in the test. This approach was selected a) because that's how the existing tests work and b) this guarantees that exactly the schemas in the repo work. It could be that future protoc versions could create different kind of serialized schema.

### DIFF
--- a/karapace/protobuf/serialization.py
+++ b/karapace/protobuf/serialization.py
@@ -4,6 +4,7 @@ See LICENSE for details
 """
 from __future__ import annotations
 
+from karapace.errors import InvalidSchema
 from karapace.protobuf.enum_constant_element import EnumConstantElement
 from karapace.protobuf.enum_element import EnumElement
 from karapace.protobuf.field import Field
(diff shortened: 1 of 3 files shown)


tvainika (approved):
LGTM

PR #815, opened 2024-02-09
Fix internal server error if request body is non-object

tvainika:

Fix internal server error if request body is non-object

# Why this way

Passing valid JSON such as string or int caused internal server errors, because code assumed to read it as object (dictionary) with fields.

### DIFF
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -298,6 +298,14 @@ class RestApp:
                     raise HTTPResponse(  # pylint: disable=raise-missing-from
                         body="Invalid request JSON body", status=HTTPStatus.BAD_REQUEST
                     )
+
+                # Prevent string, int etc. going further from here
+                if not isinstance(rapu_request.json, dict):
+                    http_error(
+                        message="Malformed request",
+                        content_type=JSON_CONTENT_TYPE,
+                        code=HTTPStatus.BAD_REQUEST,
+                    )
             else:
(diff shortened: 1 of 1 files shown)


karapace/rapu.py
```diff
@@ -286,6 +286,15 @@ async def _handle_request(
...
                     body_string = body.decode(charset)
                     rapu_request.json = json_decode(body_string)
+
+                    # Prevent string, int etc. going further from here
+                    if not isinstance(rapu_request.json, dict):
+                        http_error(
+                            message="Malformed request",
+                            content_type=JSON_CONTENT_TYPE,
+                            code=HTTPStatus.BAD_REQUEST,
+                        )
```
aiven-anton, line 296: issue: This looks like it should not be covered by the try-block, but should rather be appended after it?

aiven-anton: The try-block catches `UnicodeDecodeError`, `LookupError`, and `ValueError`, but not `HTTPResponse`.

The issue lies in maintainability: adding code to a try-block that doesn't belong there can make it significantly harder for a reader of the code to distinguish what possible paths through the code there are. As the opposite of having the minimal amount of code being covered makes it very clear, and there are no advantages of adding unrelated code to a try-block, I don't see this as opinionated.

tvainika: Sounds reasonable, fixed now.

aiven-anton (approved):

PR #816, opened 2024-02-12
fix: harden the order of keys that needs to be serialized

eliax1996:
# Why this way
Previously we were relying on the fact that the `.keys()` order and the `json.dumps()` order are always sync. This enforces by the `OrderedDict` a fixed order. Not sure if the same odering check should be performed also for the `is_key_in_canonical_format`, probably here we want to ensure that the current dict has exactly the same order in the `keys()`.

### DIFF
--- a/src/karapace/schema_registry_apis.py
+++ b/src/karapace/schema_registry_apis.py
@@ -1307,7 +1307,7 @@ class KarapaceSchemaRegistryController(KarapaceBase):
         if auth_header is not None:
             headers["Authorization"] = auth_header
 
-        with async_timeout.timeout(timeout):
+        async with async_timeout.timeout(timeout):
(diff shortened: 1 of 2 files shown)


eliax1996:
Need to do the same for the canonical format

jjaakola-aiven (approved):
LGTM.

src/karapace/schema_registry_apis.py
```diff
@@ -1307,7 +1307,7 @@ async def _forward_request_remote(
         if auth_header is not None:
             headers["Authorization"] = auth_header
 
-        with async_timeout.timeout(timeout):
+        async with async_timeout.timeout(timeout):
```
eliax1996, line 1310: Was causing flakiness also in this pr

jjaakola-aiven (approved):

PR #817, opened 2024-02-13
fix: handle BufferError from Kafka producer

jjaakola-aiven:

Try ten times when BufferError occurs, quit if cannot progress. Call the Kafka producer poll to handle the pending requests and continue.

# Why this way

Changing to `AsyncKafkaProducer` would introduce far more changes because of async functions. I decided this immediate tactical fix instead for correct the behavior.


karapace/backup/api.py
```diff
@@ -386,17 +388,28 @@ def _handle_producer_send(
...
+            producer.send(
+                instruction.topic_name,
+                key=instruction.key,
+                value=instruction.value,
+                partition=instruction.partition_index,
+                headers=[(key.decode() if key is not None else None, value) for key, value in instruction.headers],
+                timestamp=instruction.timestamp,
+            ).add_done_callback(producer_callback)
+            # Record is in the send buffer
+            break
```
aiven-anton, line 405: suggestion: Only the `producer.send().add_done_callback()` should be covered by the try-block, the `tries += 1` should be moved prior and the `break` to an `else` clause.

karapace/backup/api.py
```diff
@@ -386,17 +388,28 @@ def _handle_producer_send(
...
+        try:
+            tries += 1
+            producer.send(
+                instruction.topic_name,
+                key=instruction.key,
+                value=instruction.value,
+                partition=instruction.partition_index,
+                headers=[(key.decode() if key is not None else None, value) for key, value in instruction.headers],
+                timestamp=instruction.timestamp,
+            ).add_done_callback(producer_callback)
```
aiven-anton, line 403: issue: Can there be odd side-effects of registering the same callback for multiple operations? It's not clear to me how this will behave.

karapace/backup/api.py
jjaakola-aiven: Done.

karapace/backup/api.py
jjaakola-aiven: I moved the callback adding to the `else`.

tests/integration/backup/test_v3_backup.py
```diff
@@ -641,6 +641,64 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+
+        def __enter__(self):
+            return self._producer
+
+        def __exit__(self, exc_type, exc_value, exc_traceback):
+            self._producer.flush()
+
+    with patch("karapace.backup.api._producer") as p:
+        p.return_value = FailToSendProducerContext()
+        with pytest.raises(BackupDataRestorationError):
```
giuseppelillo, line 694: Here it would be better to check that the correct exception has been raised.
```suggestion
        with pytest.raises(BackupDataRestorationError, match="Kafka producer buffer is full"):
```

jjaakola-aiven: Done.

giuseppelillo (approved):

PR #818, opened 2024-02-16
feat: GET /mode and GET /mode/<subject:path> endpoints

jjaakola-aiven:

Add mode GET support to Karapace. The mode is always READWRITE and response contains static `{"mode": "READWRITE"}` mode for global and subject modes.


tvainika (approved):
:rocket:

PR #821, opened 2024-02-21
protobuf: recognize tag alter modification

keejon:
This PR introduces a new `FIELD_TAG_ALTER` modification in order to recognize tag change of fields.

# Why this way
Changing the tag of a field is not backwards compatible but currently goes unnoticed because fields are compared by tag and not by field name. Because of this compatibility check will only recognize `FIELD_NAME_ALTER` modification which is backward compatible. This PR changes this so that if field name exists in both schemas it will compare the tags and add `FIELD_TAG_ALTER` modification if the tags don't match.

### DIFF
--- a/karapace/protobuf/message_element.py
+++ b/karapace/protobuf/message_element.py
@@ -94,14 +94,18 @@ class MessageElement(TypeElement):
         if types.lock_message(self):
             self_tags = {}
             other_tags = {}
+            self_names = {}
+            other_names = {}
             self_one_ofs = {}
             other_one_ofs = {}
 
             for field in self.fields:
                 self_tags[field.tag] = field
+                self_names[field.name] = field
 
             for field in other.fields:
                 other_tags[field.tag] = field
+                other_names[field.name] = field
 
             for one_of in self.one_ofs:
                 self_one_ofs[one_of.name] = one_of
@@ -125,6 +129,20 @@ class MessageElement(TypeElement):
             for tag in chain(self_tags.keys(), other_tags.keys() - self_tags.keys()):
                 result.push_path(str(tag))
 
+                if (
+                    # Check if field tag has been altered
+                    tag in self_tags
+                    and (field_name := self_tags[tag].name)
+                    and (self_field := self_names.get(field_name))
+                    and (other_field := other_names.get(field_name))
+                    and (self_field.tag != other_field.tag)
+                ):
+                    result.pop_path()
+                    result.push_path(field_name)
+                    result.add_modification(Modification.FIELD_TAG_ALTER)
+                    result.pop_path()
+                    result.push_path(str(tag))
+
                 if self_tags.get(tag) is None:
                     result.add_modification(Modification.FIELD_ADD)
                 elif other_tags.get(tag) is None:
@@ -133,6 +151,7 @@ class MessageElement(TypeElement):
                     self_tags[tag].compare(other_tags[tag], result, types)
 
                 result.pop_path()
+
             # Compare OneOfs
             for name in chain(self_one_ofs.keys(), other_one_ofs.keys() - self_one_ofs.keys()):
                 result.push_path(str(name))
--- a/tests/unit/protobuf/test_protobuf_compatibility.py
+++ b/tests/unit/protobuf/test_protobuf_compatibility.py
@@ -331,6 +331,8 @@ message Foo {
     result = CompareResult()
     self_parsed.compare(other_parsed, result)
     assert not result.is_compatible()
-    assert len(result.result) == 3  # FIELD_TAG_ALTER + FIELD_DROP + FIELD_ADD
-    assert result.result[0].modification == Modification.FIELD_TAG_ALTER
-    assert result.result[0].path == "Foo.fieldX"
+    assert {(error.modification, error.path) for error in result.result} == {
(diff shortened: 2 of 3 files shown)


tests/unit/protobuf/test_protobuf_compatibility.py
```diff
@@ -300,3 +300,37 @@ def test_compatibility_ordering_change2():
...
+def test_compatibility_field_tag_change():
+    self_schema = """\
+syntax = "proto3";
+package pkg;
+message Foo {
+    string fieldA = 1;
+    string fieldB = 2;
+    string fieldC = 3;
+    string fieldX = 4;
+}
```
eliax1996, line 314: maybe add `string fieldD = 5;` since we are interested in the tag change, not in the field removal :)

tests/unit/protobuf/test_protobuf_compatibility.py
```diff
@@ -300,3 +300,37 @@ def test_compatibility_ordering_change2():
...
+"""
+
+    self_parsed: ProtoFileElement = ProtoParser.parse(location, self_schema)
+    other_parsed: ProtoFileElement = ProtoParser.parse(location, other_schema)
+
+    result = CompareResult()
+    self_parsed.compare(other_parsed, result)
+    assert not result.is_compatible()
+    assert result.result[0].modification == Modification.FIELD_TAG_ALTER
```
eliax1996, line 335: can we assert also that the number of modifications its only 1? like: `assert len(result.result) == 1`

karapace/protobuf/message_element.py
```diff
@@ -129,10 +133,21 @@ def compare(self, other: TypeElement, result: CompareResult, types: CompareTypes
...
+                elif (
+                    # Check if field tag has been altered
+                    (field_name := self_tags[tag].name)
+                    and (self_field := self_names.get(field_name))
+                    and (other_field := other_names.get(field_name))
+                    and (self_field.tag != other_field.tag)
+                ):
+                    result.pop_path()
+                    result.push_path(field_name)
+                    result.add_modification(Modification.FIELD_TAG_ALTER)
```
eliax1996, line 145: don't you need to restore the path after pushing the alteration? What happens in neasted messages? Aren't we prepending always that field name to it?
I think we should restore the path after adding the `Modification`

eliax1996 (changes requested):
Great, understood the implementation by reading the [protobuf doc](https://protobuf.dev/programming-guides/proto3/). Leaving inlined the doc as reference to understand the implementation.

1. From the protobuf doc on common errors:

> renumbering fields (sometimes done to achieve a more aesthetically pleasing number order for fields). Renumbering effectively deletes and re-adds all the fields involved in the renumbering, resulting in incompatible wire-format changes.


2. And also, under the message structure description:

> Message Structure

> A protocol buffer message is a series of key-value pairs. The binary version of a message just uses the field’s number as the key – the name and declared type for each field can only be determined on the decoding end by referencing the message type’s definition (i.e. the .proto file). Protoscope does not have access to this information, so it can only provide the field numbers.

> When a message is encoded, each key-value pair is turned into a record consisting of the field number, a wire type and a payload. The wire type tells the parser how big the payload after it is. This allows old parsers to skip over new fields they don’t understand. This type of scheme is sometimes called [Tag-Length-Value](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value), or TLV.






Other that that I have a couple of doubts/requests that I've left as comment to the review.
Thanks for you work! Great job 😄

karapace/protobuf/message_element.py
keejon: I don't think so. In line 149 there is another pop_path called, so immediately after this it will pop and push the next field on the path

tests/unit/protobuf/test_protobuf_compatibility.py
keejon: I just took over the example from the ticket. But true it's kinda confusing. I will just remove fieldD

eliax1996 (dismissed):

tests/unit/protobuf/test_protobuf_compatibility.py
```diff
@@ -300,3 +300,37 @@ def test_compatibility_ordering_change2():
...
+}
+"""
+
+    self_parsed: ProtoFileElement = ProtoParser.parse(location, self_schema)
+    other_parsed: ProtoFileElement = ProtoParser.parse(location, other_schema)
+
+    result = CompareResult()
+    self_parsed.compare(other_parsed, result)
+    assert not result.is_compatible()
+    assert len(result.result) == 3  # FIELD_TAG_ALTER + FIELD_DROP + FIELD_ADD
```
eliax1996, line 334: I know I'm pedantic, could you translate this comment into an assertion?

Like: `set((error.modification, error.path) for error in result.result) == {(FIELD_TAG_ALTER, "Foo.fieldX"), (FIELD_DROP, "Foo.fieldX"), (FIELD_ADD: "Foo.fieldX")}`

eliax1996: not sure the path its the right one for `FIELD_ADD` and `FIELD_DROP` :) probably you need to replace the field name with the tag

eliax1996 (approved):

PR #822, opened 2024-02-26
Randomize test run order

matyaskuti:

Run tests in a random order.

Reproduce rest consumer failure in random order: `--random-order-bucket=global --random-order-seed=472073` (with process number 2)

Reproduce unittest failure in random order: `--random-order --random-order-seed=275491` (bucket is default `module`)


eliax1996 (approved):
LGTM

PR #792, opened 2024-01-04
fix: Match schemas by parsed schema object and references when queried

jjaakola-aiven:
# About this change - What it does

When schema is queried with schema and references the match must happen on the parsed form and with references matching the version.

Closes #791
References: #791

PR #823, opened 2024-02-27
fix: resolving references in subjects_schema_post

keejon:
Method subjects_schema_post was using the dependencies of the schema in body to parse all the schemas that are matched. This leads to error if both schemas don't share dependencies.


tvainika (approved):

tvainika:
It looks to me that #792 was only partial fix for somewhat similar case. @jjaakola-aiven do you want to quickly check this before merge?

PR #824, opened 2024-02-27
Update GitHub actions versions

matyaskuti:

Update GitHub actions to the latest versions.


aiven-anton (approved):
👍 Would be nice to make Dependabot handle this.

Issue #812, opened 2024-02-05
schema reader can not replay schema topic


dada-engineer:
# What happened?

We have upgraded karapace from version 3.10.4 to 3.10.5 (and 3.10.6 respectively) but the schema reader hangs while replaying showing a progress of minus 0.39 percent.

`karapace.schema_reader	schema-reader	INFO    	Replay progress (0.29): -1/254 (-0.39 %)`

So the schemas are not replayed and nothing happens.

# What did you expect to happen?

I would expect to start at the first schema message in the topic so that all schemas can be rebuilt.

# What else do we need to know?

Running with official docker image v3.10.5 and 3.10.6

matyaskuti:
@dada-engineer Could you provide more of the logs? Is there anything unusual before the replay progress log lines?

dada-engineer:
Not that I can see @matyaskuti. Some rejoining but this is probably because the k8s pod running rolled over.

[Explore-logs-2024-02-06 15_54_34.txt](https://github.com/Aiven-Open/karapace/files/14182014/Explore-logs-2024-02-06.15_54_34.txt)

matyaskuti:
@dada-engineer Thanks for the logs, a couple more things:

* Is there any custom (ie. not default) configuration of the schema registry that you are running, could you share these?
* The issue persists even when Karapace is restarted?
* If the replay progress is stuck on -1, it means that the schema reader's consumer can make a connection to the cluster, read the watermark offsets, but doesn't receive any messages from the schemas topic despite polling constantly...
* ...is it possible that the messages in the topic got deleted/compacted (eg. because of the retention period)?

dada-engineer:
* config is pretty standard except for SSL cert auth but same config is working with 3.10.4
* Restarting does not fix it
* Messages are not retained in this topic and working with 3.10.4

As everything is working in 3.10.4 I am unsure it's indeed a config issue

michael-zucchetta:
Hi, 

I encountered the same issue passing from 3.9.0 to 3.10.3, then I tried to update to 3.10.8, and I obtained a similar situation:

```
kafka.coordinator   	Thread-2	INFO    	(Re-)joining group karapace-registry
karapace.schema_reader	schema-reader	INFO    	Replay progress (0.2): -1/1029 (-0.1 %)
karapace.schema_reader	schema-reader	INFO    	Replay progress (0.2): -1/1029 (-0.1 %)
[...]

PR #825, opened 2024-02-28
Log message error on schema replay

matyaskuti:

It's a valid scenario that a message with `error()` set is consumed, however these are retriable in most cases, thus only logging for now.

Related to https://github.com/Aiven-Open/karapace/issues/812

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -317,6 +317,7 @@ class KafkaSchemaReader(Thread):
             try:
                 message_key = msg.key()
                 if message_key is None:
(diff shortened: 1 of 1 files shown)


karapace/schema_reader.py
```diff
@@ -317,6 +317,7 @@ def handle_messages(self) -> None:
             try:
                 message_key = msg.key()
                 if message_key is None:
+                    LOG.warning("Empty message when consuming from topic %s, error: %s", msg.topic(), msg.error())
```
jjaakola-aiven, line 320: `Empty message key`?

keejon (approved):
LGTM 👍

PR #832, opened 2024-03-04
Handle schema reader errors

matyaskuti:

Following up on https://github.com/Aiven-Open/karapace/issues/812, better handle message errors and extend the README with the schema registry's ACL requirements.
More in commit messages.


eliax1996 (approved):
LGTM

PR #834, opened 2024-03-06
Add tests for empty payload consumer commit

matyaskuti:

Following up on #833, adding tests around the fix that was hastily implemented there.

### DIFF
--- a/tests/integration/test_rest_consumer.py
+++ b/tests/integration/test_rest_consumer.py
@@ -252,6 +252,80 @@ async def test_offsets(rest_async_client, admin_client, trail):
 
 
 @pytest.mark.parametrize("trail", ["", "/"])
+async def test_offsets_no_payload(rest_async_client, admin_client, producer, trail):
+    group_name = "offset_group_no_payload"
+    fmt = "binary"
+    header = REST_HEADERS[fmt]
+    instance_id = await new_consumer(
+        rest_async_client,
+        group_name,
+        fmt=fmt,
+        trail=trail,
+        # By default this is true
+        payload_override={"auto.commit.enable": "false"},
+    )
+    topic_name = new_topic(admin_client)
+    offsets_path = f"/consumers/{group_name}/instances/{instance_id}/offsets{trail}"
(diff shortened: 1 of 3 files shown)


tests/integration/test_rest_consumer.py
```diff
@@ -251,6 +251,80 @@ async def test_offsets(rest_async_client, admin_client, trail):
...
+        group_name,
+        fmt=fmt,
+        trail=trail,
+        # By default this is true
+        payload_override={"auto.commit.enable": "false"},
+    )
+    topic_name = new_topic(admin_client)
+    offsets_path = f"/consumers/{group_name}/instances/{instance_id}/offsets{trail}"
+    assign_path = f"/consumers/{group_name}/instances/{instance_id}/assignments{trail}"
+    consume_path = f"/consumers/{group_name}/instances/{instance_id}/records{trail}?timeout=5000"
```
eliax1996, line 270: Why the optional trailing?

tests/integration/test_rest_consumer.py
```diff
@@ -251,6 +251,80 @@ async def test_offsets(rest_async_client, admin_client, trail):
     assert "partition" in data and data["partition"] == 0, f"Unexpected partition {data}"
 
 
+@pytest.mark.parametrize("trail", ["", "/"])
```
eliax1996, line 254: I cannot see the value of that parameter

eliax1996 (changes requested):
Looks fine except for the trail that I feel it’s useless. It is?

tests/integration/test_rest_consumer.py
matyaskuti: I've followed the pattern of the other tests here, which all have it. I sort of understand why that is there, with some frameworks this can go wrong fairly easy :(

aiven-anton: I guess what's odd is that we support both, but since we do, I guess it makes sense to test it.

eliax1996 (approved):
See you spaceboy 🔫

PR #835, opened 2024-03-07
fix: handle InMemoryDatabase find_schemas latest when no versions available for subject

biggusdonzus:
This PR fixes an `IndexError: list index out of range` exception raised when calling `InMemoryDatabase.find_schemas` with argument `latest_only = True` and there's a schema with no versions.

### DIFF
--- a/tests/unit/test_in_memory_database.py
+++ b/tests/unit/test_in_memory_database.py
@@ -1,0 +1,14 @@
+"""
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+from karapace.in_memory_database import InMemoryDatabase, Subject
+
+
+class TestFindSchemas:
+    def test_returns_empty_list_when_no_schemas(self) -> None:
+        database = InMemoryDatabase()
+        subject = Subject("hello_world")
+        database.insert_subject(subject=subject)
+        expected = {subject: []}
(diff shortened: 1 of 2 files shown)


tests/unit/test_in_memory_database.py
```diff
@@ -0,0 +1,13 @@
+"""
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+from karapace.in_memory_database import InMemoryDatabase, Subject
+
+
+def test_find_schemas() -> None:
```
aiven-anton, line 8: issue: As this test has a very specific purpose: verify we don't blow up when given an empty string, I think we should have more specific naming of this test. I find this to be the best way to structure test for maximizing communication of original intent:

```python
class TestFindSchemas:  # <- _what_ are we testing
    def test_returns_empty_list_when_no_schemas(  # <- what _expectation_ do we have on the tested thing
        self,
    ) -> None:
        ... 
    # Future tests can easily be added here:
    def test_raises_foo_error_when_stars_are_not_aligned() -> None: ...
```

aiven-anton (commented):
Thanks, this is great. I could merge this as-is, but would suggest adding some more context to the test, such that future readers understand your intent.

aiven-anton (approved):

PR #837, opened 2024-03-12
422 instead of 500 on subject schema error

gpkc:
When a user sends a bad schema, we return a 500 even though the error is handled.

Here we return a 422 instead.

### DIFF
--- a/karapace/schema_registry_apis.py
+++ b/karapace/schema_registry_apis.py
@@ -1104,14 +1104,14 @@ class KarapaceSchemaRegistryController(KarapaceBase):
                 schema_type=schema_type, schema_str=schema_str, references=references, dependencies=new_schema_dependencies
             )
         except InvalidSchema:
-            self.log.exception("No proper parser found")
+            self.log.warning("Invalid schema: %r", schema_str)
             self.r(
                 body={
-                    "error_code": SchemaErrorCodes.HTTP_INTERNAL_SERVER_ERROR.value,
+                    "error_code": SchemaErrorCodes.INVALID_SCHEMA.value,
                     "message": f"Error while looking up schema under subject {subject}",
                 },
                 content_type=content_type,
-                status=HTTPStatus.INTERNAL_SERVER_ERROR,
+                status=HTTPStatus.UNPROCESSABLE_ENTITY,
             )
         except InvalidReferences:
             human_error = "Provided references is not valid"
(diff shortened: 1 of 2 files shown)


tests/integration/test_schema.py
```diff
@@ -1461,7 +1461,7 @@ async def test_schema_subject_post_invalid(registry_async_client: Client) -> Non
         f"subjects/{subject_1}",
         json={"schema": json.dumps({"type": "invalid_type"})},
     )
-    assert res.status_code == 500, "Invalid schema for existing subject should return 500"
+    assert res.status_code == 422, "Invalid schema for existing subject should return 500"
```
tvainika, line 1464: This change does not match with assert message (500)

karapace/schema_registry_apis.py
```diff
@@ -1104,14 +1104,14 @@ async def subjects_schema_post(
...
+            self.log.warning("Invalid schema: %r", schema_str, exc_info=True)
             self.r(
                 body={
-                    "error_code": SchemaErrorCodes.HTTP_INTERNAL_SERVER_ERROR.value,
+                    "error_code": SchemaErrorCodes.INVALID_SCHEMA.value,
                     "message": f"Error while looking up schema under subject {subject}",
                 },
                 content_type=content_type,
-                status=HTTPStatus.INTERNAL_SERVER_ERROR,
+                status=HTTPStatus.UNPROCESSABLE_ENTITY,
```
aiven-anton, line 1114: suggestion: I think we should use HTTP 400 here. It's a non-transient error that clearly points at client error. Unprocessable entity is nice to reserve for situations where "request is valid, but with current server state, it cannot be processed". With an invalid schema, it's not really dependent on server state.

jjaakola-aiven: This is aligned with Confluent Schema Registry and it does return `422` on `RestInvalidSchemaException`. I'd advice not changing this.

tvainika (approved):

karapace/schema_registry_apis.py
aiven-anton: Got it TIL 👍

aiven-anton (approved):

karapace/schema_registry_apis.py
```diff
@@ -1104,14 +1104,14 @@ async def subjects_schema_post(
                 schema_type=schema_type, schema_str=schema_str, references=references, dependencies=new_schema_dependencies
             )
         except InvalidSchema:
-            self.log.exception("No proper parser found")
+            self.log.warning("Invalid schema: %r", schema_str, exc_info=True)
```
aiven-anton, line 1107: suggestion: I noticed this after approving ... But passing `exc_info=True` here will put a traceback in the logs. For an expected error like this I think that's too verbose.

aiven-anton (changes requested):

aiven-anton (approved):

PR #848, opened 2024-04-08
normalization: add normalization of the options

eliax1996:
The options could be of different types:
1. string
2. boolean
3. number
4. enum
5. map
6. list
7. option

Here I've added an example for each one of them:

1. STRING:

```
option example_option = "hello world";
```

2. BOOLEAN:

```
option example_option = true;
```

3. NUMBER:

```
option example_option = 42;
```

4. ENUM:

```
enum ExampleEnum {
  OPTION_ONE = 1;
  OPTION_TWO = 2;
}

option example_option = OPTION_ONE;
```

5. MAP:

```
message ExampleMessage {
  map<string, int32> example_map = 1;
}

option (map_option) = {
  key: "key1"
  value: 42
};
```

6. LIST:

```
message ExampleMessage {
  repeated string example_list = 1;
}

option (list_option) = [
  "item1",
  "item2",
  "item3"
];
```

7. OPTION:

```
option (custom_option) = {
  string_option: "hello world",
  enum_option: OPTION_ONE,
  boolean_option: true
};
```

The goal of this pr its to start the process of normalization of the schemas initially just for the options in the protobuf types.

It changes how the protobuf schema gets normalized before being saved

# Why this way

Seems that sorting alphabetically the options its a good way to provide an invariant over the order of the options.

For the reviewers: I'm not completely familiar with protobuf nor the normalizations, please read each line of documentation (of the code and of the pr) and if anything doesn't sound right or you found something that seems wrong please discuss about it!
I've added an example for each type of option just to make sure we are not loosing any important corner case. To me sounds reasonable to order those in any case. Additional normalizations could be added in another pr

### DIFF
--- a/karapace/protobuf/proto_file_element.py
+++ b/karapace/protobuf/proto_file_element.py
@@ -13,7 +13,7 @@ from karapace.protobuf.option_element import OptionElement
 from karapace.protobuf.service_element import ServiceElement
 from karapace.protobuf.syntax import Syntax
 from karapace.protobuf.type_element import TypeElement
-from typing import Dict, List, NewType, Optional
+from typing import Dict, List, NewType, Optional, Sequence
 
 
 def _collect_dependencies_types(compare_types: CompareTypes, dependencies: Optional[Dict[str, Dependency]], is_self: bool):
@@ -37,17 +37,21 @@ PackageName = NewType("PackageName", str)
 
 
 class ProtoFileElement:
+    types: Sequence[TypeElement]
+    services: Sequence[ServiceElement]
+    extend_declarations: Sequence[ExtendElement]
+
     def __init__(
         self,
         location: Location,
         package_name: Optional[PackageName] = None,
         syntax: Optional[Syntax] = None,
-        imports: Optional[List[TypeName]] = None,
-        public_imports: Optional[List[TypeName]] = None,
-        types: Optional[List[TypeElement]] = None,
-        services: Optional[List[ServiceElement]] = None,
-        extend_declarations: Optional[List[ExtendElement]] = None,
-        options: Optional[List[OptionElement]] = None,
+        imports: Optional[Sequence[TypeName]] = None,
+        public_imports: Optional[Sequence[TypeName]] = None,
+        types: Optional[Sequence[TypeElement]] = None,
+        services: Optional[Sequence[ServiceElement]] = None,
+        extend_declarations: Optional[Sequence[ExtendElement]] = None,
+        options: Optional[Sequence[OptionElement]] = None,
     ) -> None:
         if types is None:
             types = list()
--- a/karapace/protobuf/proto_normalizations.py
+++ b/karapace/protobuf/proto_normalizations.py
@@ -2,6 +2,10 @@
 Copyright (c) 2024 Aiven Ltd
 See LICENSE for details
 """
+
+from __future__ import annotations
+
+from karapace.dependency import Dependency
 from karapace.protobuf.enum_constant_element import EnumConstantElement
 from karapace.protobuf.enum_element import EnumElement
 from karapace.protobuf.extend_element import ExtendElement
@@ -12,23 +16,86 @@ from karapace.protobuf.one_of_element import OneOfElement
 from karapace.protobuf.option_element import OptionElement
 from karapace.protobuf.proto_file_element import ProtoFileElement
 from karapace.protobuf.rpc_element import RpcElement
+from karapace.protobuf.schema import ProtobufSchema
 from karapace.protobuf.service_element import ServiceElement
 from karapace.protobuf.type_element import TypeElement
-from karapace.typing import StrEnum
-from typing import List
+from karapace.schema_references import Reference
+from typing import Mapping, Sequence
 
-
-class ProtobufNormalisationOptions(StrEnum):
-    sort_options = "sort_options"
+import abc
 
 
 def sort_by_name(element: OptionElement) -> str:
     return element.name
 
 
-def type_field_element_with_sorted_options(type_field: FieldElement) -> FieldElement:
+class NormalizedRpcElement(RpcElement):
+    pass
+
+
+class NormalizedServiceElement(ServiceElement):
+    rpcs: Sequence[NormalizedRpcElement] | None = None
+
+
+class NormalizedFieldElement(FieldElement):
+    pass
+
+
+class NormalizedExtendElement(ExtendElement):
+    fields: Sequence[NormalizedFieldElement] | None = None
+
+
+class NormalizedTypeElement(TypeElement, abc.ABC):
+    nested_types: Sequence[NormalizedTypeElement]
+
+
+class NormalizedProtoFileElement(ProtoFileElement):
+    types: Sequence[NormalizedTypeElement]
+    services: Sequence[NormalizedServiceElement]
+    extend_declarations: Sequence[NormalizedExtendElement]
+
+
+class NormalizedMessageElement(MessageElement, NormalizedTypeElement):
+    nested_types: Sequence[NormalizedTypeElement]
+    fields: Sequence[NormalizedFieldElement]
+    one_ofs: Sequence[OneOfElement]
+    groups: Sequence[GroupElement]
+
+
+class NormalizedEnumConstantElement(EnumConstantElement):
+    pass
+
+
+class NormalizedEnumElement(EnumElement, NormalizedTypeElement):
+    constants: Sequence[NormalizedEnumConstantElement]
+
+
+class NormalizedGroupElement(GroupElement):
+    fields: Sequence[NormalizedFieldElement] | None = None
+
+
+class NormalizedProtobufSchema(ProtobufSchema):
+    proto_file_element: NormalizedProtoFileElement
+
+    def __init__(
+        self,
+        schema: str,
+        references: Sequence[Reference] | None = None,
+        dependencies: Mapping[str, Dependency] | None = None,
+        proto_file_element: ProtoFileElement | None = None,
+    ) -> None:
+        super().__init__(schema, references, dependencies, proto_file_element)
+        self.proto_file_element = normalize(self.proto_file_element)
+
+
+class NormalizedOneOfElement(OneOfElement):
+    fields: Sequence[NormalizedFieldElement]
+    groups: Sequence[NormalizedGroupElement]
+
+
+def type_field_element_with_sorted_options(type_field: FieldElement) -> NormalizedFieldElement:
     sorted_options = None if type_field.options is None else list(sorted(type_field.options, key=sort_by_name))
-    return FieldElement(
+    return NormalizedFieldElement(
         location=type_field.location,
         label=type_field.label,
         element_type=type_field.element_type,
@@ -41,9 +108,9 @@ def type_field_element_with_sorted_options(type_field: FieldElement) -> FieldEle
     )
 
 
-def enum_constant_element_with_sorted_options(enum_constant: EnumConstantElement) -> EnumConstantElement:
+def enum_constant_element_with_sorted_options(enum_constant: EnumConstantElement) -> NormalizedEnumConstantElement:
     sorted_options = None if enum_constant.options is None else list(sorted(enum_constant.options, key=sort_by_name))
-    return EnumConstantElement(
+    return NormalizedEnumConstantElement(
         location=enum_constant.location,
         name=enum_constant.name,
         tag=enum_constant.tag,
@@ -52,14 +119,14 @@ def enum_constant_element_with_sorted_options(enum_constant: EnumConstantElement
     )
 
 
-def enum_element_with_sorted_options(enum_element: EnumElement) -> EnumElement:
+def enum_element_with_sorted_options(enum_element: EnumElement) -> NormalizedEnumElement:
     sorted_options = None if enum_element.options is None else list(sorted(enum_element.options, key=sort_by_name))
     constants_with_sorted_options = (
         None
         if enum_element.constants is None
         else [enum_constant_element_with_sorted_options(constant) for constant in enum_element.constants]
     )
-    return EnumElement(
+    return NormalizedEnumElement(
         location=enum_element.location,
         name=enum_element.name,
         documentation=enum_element.documentation,
@@ -68,11 +135,11 @@ def enum_element_with_sorted_options(enum_element: EnumElement) -> EnumElement:
     )
 
 
-def groups_with_sorted_options(group: GroupElement) -> GroupElement:
+def groups_with_sorted_options(group: GroupElement) -> NormalizedGroupElement:
     sorted_fields = (
         None if group.fields is None else [type_field_element_with_sorted_options(field) for field in group.fields]
     )
-    return GroupElement(
+    return NormalizedGroupElement(
         label=group.label,
         location=group.location,
         name=group.name,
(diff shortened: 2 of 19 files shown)


tvainika:
Support for `normalize` flag is also needed in `POST /subjects/<subject>` which returns ID of schema if already registered.

jjaakola-aiven (commented):
Options are allowed on message and field level, I'd add the support for these also.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+    # doing it here since the subtypes do not declare the nested_types property
+    type_element.nested_types = sorted_neasted_types
+
+    if isinstance(type_element, EnumElement):
+        return enum_element_with_sorted_options(type_element)
+
+    if isinstance(type_element, MessageElement):
+        return message_element_with_sorted_options(type_element)
+
+    raise ValueError("Unknown type element")  # tried with assert_never but it did not work
```
eliax1996, line 139: @aiven-anton any suggestion on how I can tell `mypy` that all the concrete instances are managed? I've also tried with `assert type(type_element) != TypeElement` (since its a concrete class and could be instantiated) but didn't helped

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+def type_element_with_sorted_options(type_element: TypeElement) -> TypeElement:
+    sorted_neasted_types: List[TypeElement] = []
+
+    for nested_type in type_element.nested_types:
+        if isinstance(nested_type, EnumElement):
+            sorted_neasted_types.append(enum_element_with_sorted_options(nested_type))
+        elif isinstance(nested_type, MessageElement):
+            sorted_neasted_types.append(message_element_with_sorted_options(nested_type))
+        else:
+            raise ValueError("Unknown type element")  # tried with assert_never but it did not work
```
eliax1996, line 128: same here @aiven-anton

tests/unit/protobuf/test_protobuf_normalization.py
```diff
@@ -0,0 +1,520 @@
...
+
+location: Location = Location("some/folder", "file.proto")
+
+
+# this would be a good case for using a property based test with a well-formed message generator
+# (hypothesis could work if the objects representing the language
+# do always correspond to a valid proto schema, that for a correct parser should always be true)
+# we should create a set of random valid DSLs and check that the normalization is giving back a sorted options list
+#
+# nb: we could use the protoc compiler as an oracle to check that the files are always valid protobuf schemas
```
eliax1996, line 20: moment: 
immagine

eliax1996 (commented):
ready to be reviewed again :)

karapace/protobuf/proto_normalizations.py
aiven-anton: @eliax1996 The only way to define a hierarchy that works like ADTs in Python is to define base class and union separately. There's been proposals to introduce a concept of sealed classes but so far it has met a fair bit of justified critique, and it was [raised again recently](https://discuss.python.org/t/draft-pep-sealed-decorator-for-static-typing/49206/20) without much excitement.

What you can do currently is the below, and use `Thing` as type.

```python

class BaseThing: ...

class MemberA(BaseThing): ...
class MemberB(BaseThing): ...

Thing = MemberA | MemberB
```

The reason mypy rejects your code is it fails to prove there aren't any other subclasses of `TypeElement` it hasn't found yet. Theoretically this could be proven, but not with reasonable performance. You should also be able to see this with a `reveal_type(type_element)` inserted above the `raise ValueError`. I would expect the type to be `TypeElement` there.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+    return EnumConstantElement(
+        location=enum_constant.location,
+        name=enum_constant.name,
+        tag=enum_constant.tag,
+        documentation=enum_constant.documentation,
+        options=sorted_options,
+    )
+
+
+def enum_element_with_sorted_options(enum_element: EnumElement) -> EnumElement:
```
aiven-anton, line 55: It would be valuable if we can encode normalization in the type system, so that instead of having functions like `(EnumElement) -> EnumElement` here we would have functions like `(EnumElement) -> NormalizedEnumElement`. That way we can potentially make use of the type system to bring confidence we always have normalized items when normalization is required. Presumably we would also need an additional type at the "top".

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+from karapace.protobuf.option_element import OptionElement
+from karapace.protobuf.proto_file_element import ProtoFileElement
+from karapace.protobuf.rpc_element import RpcElement
+from karapace.protobuf.service_element import ServiceElement
+from karapace.protobuf.type_element import TypeElement
+from karapace.typing import StrEnum
+from typing import List
+
+
+class ProtobufNormalisationOptions(StrEnum):
```
jjaakola-aiven, line 21: Not used, I'd remove. I also want to challenge the need for different normalization options also in future. I think there should be only one normalized/canonical format that Karapace follows, this would make the output stable.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+        documentation=one_ofs.documentation,
+        fields=sorted_fields,
+        groups=sorted_groups,
+        options=sorted_options,
+    )
+
+
+def message_element_with_sorted_options(message_element: MessageElement) -> MessageElement:
+    sorted_options = None if message_element.options is None else list(sorted(message_element.options, key=sort_by_name))
+    sorted_neasted_types = [type_element_with_sorted_options(nested_type) for nested_type in message_element.nested_types]
```
jjaakola-aiven, line 101:
```suggestion
    sorted_nested_types = [type_element_with_sorted_options(nested_type) for nested_type in message_element.nested_types]
```

karapace/protobuf/proto_normalizations.py
jjaakola-aiven: The entrypoint to normalize is `def normalize_options_ordered`

How about having
```python
class NormalizedProtobufSchema(ProtobufSchema): ...

class NormalizedProtoFileElement(ProtoFileElement):
     
class NormalizedEnumElement(EnumElement): ...

...

Usage would be at the `schema_models.parse_protobuf_schema_definition`:
```python
protobuf_schema = ProtobufSchema(schema_definition, references, dependencies) if not normalize else NormalizedProtobufSchema(schema_definition, references, dependencies)
    if validate_references:
    ...
```


This would give the same interface for normalized versions and give typing to some extent.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+        reserveds=message_element.reserveds,
+        fields=sorted_fields,
+        one_ofs=sorted_one_ofs,
+        extensions=message_element.extensions,
+        groups=message_element.groups,
+    )
+
+
+def type_element_with_sorted_options(type_element: TypeElement) -> TypeElement:
+    sorted_neasted_types: List[TypeElement] = []
```
jjaakola-aiven, line 120:
```suggestion
    sorted_nested_types: List[TypeElement] = []
```

tvainika: Also in test files multiple times same typo

tests/integration/test_schema_protobuf.py
```diff
@@ -1270,3 +1270,106 @@ async def test_protobuf_update_ordering(registry_async_client: Client) -> None:
...
+    assert res.status_code == 404
+
+    res = await registry_async_client.post(f"subjects/{subject}?normalize=true", json=body)
+
+    assert res.status_code == 200
+    assert "id" in res.json()
+    assert original_schema_id == res.json()["id"]
+
+
+async def test_protobuf_normalization_of_options_specify_version(registry_async_client: Client) -> None:
```
tvainika, line 1328: I failed to see what was beef in this second test compared to first? There seems to be same schemas used so maybe just one test with more API calls, and/or some comments about intention?

Also integration test could fetch the schema by id after registering it with `normalize=true` to verify that fetch returns normalized version.

karapace/protobuf/proto_normalizations.py
tvainika: When loading already registered schemas from `_schemas` how do we know if those are normalized or not, and what is gained by having normalized schema type? I don't see much value here.

For example if loading earlier registered schema from `_schemas` we do not know if it was registered as with normalize flag or not. What is the type of it, and if compared to newly registering schema with normalize flag true or false, how does these types affect equality check?

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,220 @@
...
+    return ServiceElement(
+        location=service_element.location,
+        name=service_element.name,
+        documentation=service_element.documentation,
+        rpcs=sorted_rpc,
+        options=sorted_options,
+    )
+
+
+def normalize_options_ordered(proto_file_element: ProtoFileElement) -> ProtoFileElement:
```
tvainika, line 185: This can be simply named as `normalize`, no need for different kind of normalizations?

But should it be method of ProtobufSchema or ProtoFileElement instead of separate function?

karapace/protobuf/proto_normalizations.py
jjaakola-aiven: You are right. The property of normalization is not essential nor does provide any value at consuming side. Considering your comment and I also think that this is functional property at registration.

jjaakola-aiven: I don't think so, they would be just different and not equal?

karapace/protobuf/proto_normalizations.py
eliax1996: initially I was considering of implementing the different types of normalization into different prs and with different goals, like normalizing the option orders, normalize the string names etc etc. The normalize function would be much more harder than this one and I thought about separating the tests and the guarantees of those phases. Probably isn't a good idea though :)

tests/integration/test_schema_protobuf.py
eliax1996: The `test_protobuf_normalization_of_options` was verifying that producing a schema with the normalized does correctly register a normalized schema because by trying to seek for the schema later without normalizing it we didn't find it while by normalizing it we can and so its registered. While the `test_protobuf_normalization_of_options_specify_version` verify that its idempotent and the second time we try to register the same normalized schema the returned id its the same, so the schema its not registered twice. Probably the second one isn't really specific to the `normalize` option and can be transformed into verify that the returned schema its a normalized schema

karapace/protobuf/proto_normalizations.py
eliax1996: I've learned that nested doesn't have the `a` :D

eliax1996: I've learned that nested doesn't have the `a` :D

karapace/protobuf/proto_normalizations.py
eliax1996: got it, I was thinking python was based a closed world model and so the classes where implicitly sealed. But isn't true, in fact you can always use a library and define a child of a class belonging to that library, so the checking isn't always exaustive

karapace/protobuf/proto_file_element.py
aiven-anton, line 1: Note that type checking is not enabled yet for this file.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class NormalizedMessageElement(NormalizedTypeElement, MessageElement):
+    nested_types: Sequence[NormalizedTypeElement]
+    fields: Sequence[NormalizedFieldElement]
+    one_ofs: Sequence[OneOfElement]
+    groups: Sequence[GroupElement]
+
+    def to_schema(self) -> str:
+        return MessageElement.to_schema(self)
```
aiven-anton, line 67:
```suggestion
class NormalizedMessageElement(MessageElement, NormalizedTypeElement):
    nested_types: Sequence[NormalizedTypeElement]
    fields: Sequence[NormalizedFieldElement]
    one_ofs: Sequence[OneOfElement]
    groups: Sequence[GroupElement]
```

issue: This looks like wrong MRO.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,282 @@
...
+
+class NormalizedTypeElement(TypeElement):
+    nested_types: Sequence["NormalizedTypeElement"]
+
+    def to_schema(self) -> str:
+        """Convert the object to valid protobuf syntax.
+
+        This must be implemented by subclasses.
+        """
+        raise NotImplementedError()
```
aiven-anton, line 51:
```suggestion
```

issue: This is identical to inherited (non-) implementation?

eliax1996: pylint was complaining that wasn’t implemented, how can I get rid of that problem without replicating the non implementation? With an @abstract annotation?

karapace/protobuf/proto_file_element.py
eliax1996: How can we enable it?

aiven-anton: The ignores live here: https://github.com/Aiven-Open/karapace/blob/29221bf73ce62e609d001036efdd82c1f03e11f9/mypy.ini#L26

That's essentially a somewhat hidden todo list :)

karapace/protobuf/proto_normalizations.py
eliax1996: ah you are right, thanks! Multi inheritance its complex :P

eliax1996: same its true also for `NormalizedEnumElement`

karapace/protobuf/proto_normalizations.py
aiven-anton: Ah, yes it might be happy with a `@abstractmethod` (`abc.abstractmethod`), it's appropriate but I don't know whether pylint understands it.

aiven-anton: Oh wait, what you'd need is to add `abc.ABC` as base to the class.

karapace/protobuf/proto_normalizations.py
aiven-anton: Unmarked as resolved as the point here was also that you should not need to define `to_schema` when the MRO is correct.

karapace/protobuf/proto_normalizations.py
```diff
@@ -0,0 +1,277 @@
...
+class NormalizedExtendElement(ExtendElement):
+    fields: Sequence[NormalizedFieldElement] | None = None
+
+
+class NormalizedTypeElement(TypeElement):
+    nested_types: Sequence[NormalizedTypeElement]
+
+    @abstractmethod
+    def to_schema(self) -> str:
+        pass
```
aiven-anton, line 52:
```suggestion
class NormalizedTypeElement(TypeElement, abc.ABC):
    nested_types: Sequence[NormalizedTypeElement]
```

suggestion: If pylint complains, I'd just ignoring it. We should prefer marking the class final over the repeating the empty definition.

aside: In ABCs, it's a good idea to raise `NotImplementedError` in abstract definitions.

aiven-anton (approved):

jjaakola-aiven (approved):

PR #851, opened 2024-04-17
Fix error message formatting variable expansion

tvainika:

Fix formatting of error message variable expansion if subject level compatibility mode is not configured.

# Why this way

Old code was literally asserting that error message was `Subject '%s' does not have subject-level compatibility configured`


aiven-anton (approved):
Good, updated tests will behave more predictably 👍

PR #855, opened 2024-04-30
README: Fix badge for workflow run of `tests.yml`

andygrunwald:

Set the correct repository links after the repository has been moved from @aiven to @aiven-open

References: None

# Why this way

Avoiding redirects or broken images (e.g. when image redirects not work)


keejon (approved):
👍

PR #862, opened 2024-05-06
fix: handle ConnectionError in rapu._handle_messages

jjaakola-aiven:

If client connection is broken there is nothing the server can do. Handle the specific case and silence stats and reporting. The aiohttp expects a response object, although this is not and cannot be written to client as the connection is already broken.

### DIFF
--- a/tests/unit/test_rapu.py
+++ b/tests/unit/test_rapu.py
@@ -2,7 +2,11 @@
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
+from aiohttp.web import Request
+from karapace.config import DEFAULTS
+from karapace.karapace import KarapaceBase
 from karapace.rapu import HTTPRequest, REST_ACCEPT_RE, REST_CONTENT_TYPE_RE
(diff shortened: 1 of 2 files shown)


aiven-anton (dismissed):

tests/unit/test_rapu.py
```diff
@@ -154,3 +157,21 @@ def test_content_type_re():
         "serialization_format": "json",
         "general_format": None,
     }
+
+
+async def test_raise_connection_error_handling() -> None:
+    request_mock = Mock()
```
nosahama, line 163: **[non-blocking]**
Specifying `spec` could aid understanding of what is mocked directly in the test.
I see there's a `rapu.HTTPRequest`, but that doesn't have a `.read()` method.

nosahama (dismissed):

tests/unit/test_rapu.py
jjaakola-aiven: Done.

nosahama (approved):

aiven-anton (approved):

PR #872, opened 2024-05-13
chore: bump kafka-python:2.0.2 to kafka-python-ng:2.2.2

jjaakola-aiven:

The `kafka-python` is unmaintained. Fork of it has been created as `kafka-python-ng`: https://github.com/wbarnha/kafka-python-ng
See note at `kafka-python`: https://github.com/dpkp/kafka-python/issues/2431

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -4,6 +4,7 @@ aiohttp<4
 confluent-kafka==2.3.0
 isodate<1
 jsonschema<5
+kafka-python-ng==2.2.2
 lz4
 networkx<4
 protobuf<4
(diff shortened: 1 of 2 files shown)


aiven-anton:
This seems like a good idea, however, are all fixes we have on our fork merged in -ng? We have a bunch of them and it would be sad to see regressions introduced:

https://github.com/dpkp/kafka-python/compare/master...aiven:kafka-python:main

jjaakola-aiven:
Socks5 wrapper is missing from `kafka-python-ng`. This requires some work in the Aiven fork of `kafka-python`.

PR #873, opened 2024-05-13
test: store soft deleted schema

### LINKED ISSUE #170: Write regression test for soft deletes bug
PR fixing the issue: https://github.com/aiven/karapace/pull/169

Previously Karapace would ignore soft deleted schemas (schemas with `deleted` flag set) if the schemas was not seen before. This was a problem since the `_shemas` topic is ` log.cleanup.policy=compact`, and Karapace would only see the deleted schema. That meant the schema would not be available through the `schemas/ids/:id` which broke some use cases.

jjaakola-aiven:

When _schemas topic has been compacted and only soft deleted version of the schema is present the schema must be stored in the in memory database.

Fixes #170.

### DIFF
--- a/tests/unit/test_schema_reader.py
+++ b/tests/unit/test_schema_reader.py
@@ -20,9 +20,11 @@ from karapace.schema_reader import (
     OFFSET_EMPTY,
     OFFSET_UNINITIALIZED,
 )
+from karapace.typing import SchemaId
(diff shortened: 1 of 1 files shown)


keejon (dismissed):
LGTM 👍

keejon:
@jjaakola-aiven some conflicts have to be resolved first

keejon (approved):

PR #874, opened 2024-05-13
chore: update kafka-python with vendored six 1.16.0

jjaakola-aiven:

Update `kafka-python` with vendored `six==1.16.0`. This enables update of `pytest` to `8.x`.


keejon (approved):
LGTM 👍

PR #875, opened 2024-05-13
fix: Restore pip-compile invocation

aiven-anton:
In eb9ee31b3ff5879546d70737903cd6941562fb2d, pip-compile was changed to no longer cd into the requirements directory. The reason for this change was because of the use of `ONESHELL` which causes diverging behavior in environments with make versions that don't yet support this mode.

This commit reverts the removal of the `cd` invocations, and additionally removes the `ONESHELL` option which should make behavior aligned across dev envs.

References: #xxxxx

# Why this way


eliax1996 (approved):
LGTM

PR #877, opened 2024-05-14
fix: consume only one record at a time after startup

jjaakola-aiven:

If Karapace would expect to consume multiple records after startup, there is extra latency of the consumption timeout as consumer does not return control back to schema reader until it happens. This would cause extra latency when storing schema.
On normal operating mode after startup there expectation is to process single record at a time and return control immediately to schema reader when record is available.

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -58,6 +58,16 @@ OFFSET_EMPTY: Final = -1
 KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS: Final = 2.0
 SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS: Final = 5.0
 
+# For good startup performance the consumption of multiple
+# records for each consume round is essential.
+# Consumer default is 1 message for each consume call and after
+# startup the default is a good value. If consumer would expect
+# more messages it would return control back after timeout and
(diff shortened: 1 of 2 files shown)


karapace/schema_reader.py
```diff
@@ -58,6 +58,16 @@
...
 SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS: Final = 5.0
 
+# For good startup performance the consumption of multiple
+# records for each consume round is essential.
+# Consumer default is 1 message for each consume call and after
+# startup the default is a good value. If consumer would expect
+# more messages it would return control back after timeout and
+# making schema storing latency to be `processing time + timeout`.
+MAX_MESSAGES_TO_CONSUME_ON_STARTUP: Final = 1000
+MAX_MESSAGES_TO_CONSUME_AFTER_STARTUP = 1
```
eliax1996, line 68: nit: `Final`?

eliax1996 (changes requested):
almost perfect! If you can fix the `Final` I'm ready to merge

karapace/schema_reader.py
jjaakola-aiven: Done.

eliax1996 (approved):

PR #879, opened 2024-05-14
Update schema_registry.py

antoniocali:
validate_version now check if version_number is equal to -1 and in case forward "latest"

`validate_version` check now if version is equals to -1 and in case forward "latest"
I've modifies a test that delete a version via -1 instad of latest.

This should work as intended

# Why this way

Confluent Schema Registry accept -1 as version for latest schema version


antoniocali:
I'm not sure why one suite of tests failed. it looks more a timeout to me

nosahama:
#668 will be fixed with #885, closing this PR.

PR #883, opened 2024-05-24
fix: log InvalidSchema exception on INFO level

jjaakola-aiven:

The Sentry SDK integration by default sends the warning level logs to Sentry backend. Posted input data is handled and correct 422 response is sent back to client and there is no need to create entry for this to Sentry.


jjaakola-aiven:
This is fixed in PR: https://github.com/Aiven-Open/karapace/commit/82a7c44af0f747cded11b6076738830c19a47ccb
There was a call to `log.exception()'.

PR #884, opened 2024-05-24
Fix java multiple files protobuf option rendering

jjaakola-aiven:

The protobuf rendering did not change the boolean value to lowercase as expected in the IDL file.


keejon (approved):
LGTM 👍

PR #893, opened 2024-06-05
refactor: remove int inheritance from Version

jjaakola-aiven:

Make Version a container class for integer for stronger typing.

### DIFF
--- a/karapace/typing.py
+++ b/karapace/typing.py
@@ -9,6 +9,8 @@ from karapace.errors import InvalidVersion
 from typing import ClassVar, Dict, List, Mapping, NewType, Sequence, Union
 from typing_extensions import TypeAlias
 
+import functools
+
 JsonArray: TypeAlias = List["JsonData"]
 JsonObject: TypeAlias = Dict[str, "JsonData"]
 JsonScalar: TypeAlias = Union[str, int, float, None]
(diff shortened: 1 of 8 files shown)


karapace/typing.py
```diff
@@ -59,23 +59,47 @@ class Mode(StrEnum):
...
+        return self._value >= other.value
+
+    def __lt__(self, other: Version) -> bool:
+        return self._value < other.value
+
+    def __le__(self, other: Version) -> bool:
+        return self._value <= other.value
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Version):
```
nosahama, line 92: **[non-blocking]**
Would be nice to also have this check on the other comparison methods.

nosahama (dismissed):

karapace/typing.py
jjaakola-aiven: Done.

karapace/typing.py
```diff
@@ -59,23 +59,55 @@ class Mode(StrEnum):
...
+        return f"{int(self._value)}"
 
     def __repr__(self) -> str:
-        return f"Version={int(self)}"
+        return f"Version({int(self._value)})"
+
+    def __gt__(self, other: Version) -> bool:
+        if isinstance(other, Version):
+            return self._value > other.value
+        raise NotImplemented(f">Cannot compare {repr(other)} with {repr(self)}")
```
aiven-anton, line 82: issue: `NotImplemented` is not an exception, and it should not be raised, it should be returned.

karapace/typing.py
```diff
@@ -59,23 +59,55 @@ class Mode(StrEnum):
...
 
     def __str__(self) -> str:
-        return f"{int(self)}"
+        return f"{int(self._value)}"
 
     def __repr__(self) -> str:
-        return f"Version={int(self)}"
+        return f"Version({int(self._value)})"
+
+    def __gt__(self, other: Version) -> bool:
```
aiven-anton, line 79: Usually you do not need to implement all of these methods, you can just implement one of them and decorate the class with `@functools.total_ordering`.

karapace/typing.py
keejon: @jjaakola-aiven think you have to either raise NotImplementedError or return NotImplemented

keejon: ah Anton was faster :-)

karapace/typing.py
aiven-anton: It also isn't callable, these lines should be just simply `return NotImplemented`.

```pycon
>>> NotImplemented("Foo")
  File "<stdin>", line 1, in <module>
TypeError: 'NotImplementedType' object is not callable
```

karapace/typing.py
nosahama: `@functools.total_ordering` comes with some performance overhead, docs advise to implement these functions if the performance hit is too much.

nosahama (approved):

keejon (approved):

PR #894, opened 2024-06-06
fix: move offset always, also with invalid processed records

jjaakola-aiven:

Move the offset in schema reader also when the record key or value is not valid JSON.
The reading can get stuck as offset is not moved.


keejon (approved):
LGTM 👍

PR #896, opened 2024-06-07
rapu: catch `aiohttp.ClientError` with a `503` response status

nosahama:
This catches `aiohttp.ClientError` and returns `HTTPStatus<503>` rather than `HTTPStatus<500>`. 
We also catch the base `aiohttp.ClientError` rather than just `aiohttp.ClientConnectorError`, as for us, most client errors will also be TCP connection related, rather than actual invalid request data, etc and we include the exception information with the logged message, this can further aid debugging. 

References: [HH-3866](https://aiven.atlassian.net/browse/HH-3866)

# Why this way
Reflects properly the behaviour of the server via the response status.

### DIFF
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -393,11 +393,11 @@ class RestApp:
             )
             headers = {"Content-Type": "application/json"}
             resp = aiohttp.web.Response(body=body, status=status.value, headers=headers)
-        except ConnectionError as connection_error:
+        except (ConnectionError, aiohttp.ClientError) as connection_error:
             # TCP level connection errors, e.g. TCP reset, client closes connection.
             self.log.debug("Connection error.", exc_info=connection_error)
             # No response can be returned and written to client, aiohttp expects some response here.
-            resp = aiohttp.web.Response(text="Connection error", status=HTTPStatus.INTERNAL_SERVER_ERROR.value)
+            resp = aiohttp.web.Response(text="Connection error", status=HTTPStatus.SERVICE_UNAVAILABLE.value)
         except asyncio.CancelledError:
             self.log.debug("Client closed connection")
             raise
--- a/tests/unit/test_rapu.py
+++ b/tests/unit/test_rapu.py
@@ -2,11 +2,14 @@
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
+from aiohttp.client_exceptions import ClientConnectionError
 from aiohttp.web import Request
 from karapace.config import DEFAULTS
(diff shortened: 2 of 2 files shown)


karapace/rapu.py
```diff
@@ -393,11 +393,11 @@ async def _handle_request(
             )
             headers = {"Content-Type": "application/json"}
             resp = aiohttp.web.Response(body=body, status=status.value, headers=headers)
-        except ConnectionError as connection_error:
+        except (ConnectionError, aiohttp.ClientError) as connection_error:
             # TCP level connection errors, e.g. TCP reset, client closes connection.
             self.log.debug("Connection error.", exc_info=connection_error)
             # No response can be returned and written to client, aiohttp expects some response here.
```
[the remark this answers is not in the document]
nosahama: I would not think that they differ too much, they're both handling the TCP connection errors, one raised at the python interpreter level and the other raised by the client library. The only concern for a split would be if we do not want to return the same status code for both error types.

karapace/rapu.py
```diff
@@ -393,11 +393,11 @@ async def _handle_request(
             )
             headers = {"Content-Type": "application/json"}
             resp = aiohttp.web.Response(body=body, status=status.value, headers=headers)
-        except ConnectionError as connection_error:
+        except (ConnectionError, aiohttp.ClientError) as connection_error:
```
eliax1996, line 396: Do you really want to catch all the `aiohttp.ClientError` children? Or just `aiohttp.ClientConnectorError`?

eliax1996: it could also probably be fine with the `ClientError` in general -> https://docs.aiohttp.org/en/stable/_modules/aiohttp/client_exceptions.html#ClientError

just wondering if all the subcases matches well with the 503

eliax1996 (changes requested):
Just asking to update the description of the pr :) the change by itself its an improvement.
Giving @jjaakola-aiven the possibility to reply on his raised point even if to me it sounds reasonable to catch both exceptions in the same scope, maybe would be useful to highlight a case where the behaviour its different and the management should be done in a different way

karapace/rapu.py
nosahama: I've gone with catching the base so as to clean up our issue tracker and we can always further investigate with the exception information added to the base, else, we'd need to branch for each possible connection error or client error, as we are the client in this case.

eliax1996 (approved):
lgtm

jjaakola-aiven (approved):

PR #898, opened 2024-06-10
Fix fixture order for consumer tests

jjaakola-aiven:

Consumer close can get stuck if the referenced topic is deleted from broker before.

References: #xxxxx

# Why this way

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -52,7 +52,7 @@ KAFKA_WAIT_TIMEOUT = 60
 KAFKA_SCALA_VERSION = "2.13"
 # stdout logger is disabled to keep the pytest report readable
 KAFKA_LOG4J = TEST_INTEGRATION_DIR / "config" / "log4j.properties"
-WORKER_COUNTER_KEY = "worker_counter"
+WORKER_IDS_KEY = "workers"
 
 REST_PRODUCER_MAX_REQUEST_BYTES = 10240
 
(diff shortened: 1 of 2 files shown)


nosahama (approved):

tests/integration/kafka/test_consumer.py
```diff
@@ -25,7 +25,11 @@ class TestPartitionsForTopic:
     def test_partitions_for_returns_empty_for_unknown_topic(self, consumer: KafkaConsumer) -> None:
         assert consumer.partitions_for_topic("nonexistent") == {}
 
-    def test_partitions_for(self, consumer: KafkaConsumer, new_topic: NewTopic) -> None:
+    def test_partitions_for(
+        self,
+        new_topic: NewTopic,
+        consumer: KafkaConsumer,
+    ) -> None:
```
eliax1996, line 32: its actually the order of the fixture that causes the problem?

eliax1996: this is reaaaaally bad, I would expect this to be uninfluent. The order of parameters in a function shouldn't change the semantics of a program.
I wasn't quite liking the idea of fixture but this is much more dangerous.
Shouldn't we create a higher order fixture that gives the parameters and enforces the right order?

IMO if we need a topological sort of the fixture generation we should enforce it by definition

eliax1996 (approved):
Its a clear improvement even if I don't really like the idea of the test being relying on the order of the input parameters.
Pragmatically this is a big improvement though

tests/integration/conftest.py
```diff
@@ -227,8 +227,10 @@ def fixture_consumer(
         # Speed things up for consumer tests to discover topics, etc.
         topic_metadata_refresh_interval_ms=200,
     )
-    yield consumer
-    consumer.close()
+    try:
+        yield consumer
+    finally:
+        consumer.close()
```
aiven-anton, line 233: Just noting that this shouldn't have any impact, fixtures do not work like a context manager, and the pytest framework will resume the generator on clean-up. So the `finally:` here doesn't make any difference (but also doesn't hurt).

PR #890, opened 2024-06-03
Example metric integration test

eliax1996:
# About this change - What it does

References: #xxxxx

# Why this way

Issue #204, opened 2021-04-27
Add observability


mortenlj:
# What is currently missing?

Observability into the performance and operation of Karapace.

# How could this be improved?

It would be very useful to add some observability, like number of requests, number of messages produced/consumed, schemas etc.

For my particular use case, it would be great to have a prometheus endpoint with these metrics.

# Is this a feature you would work on yourself?

Probably not at this time, but if my backlog clears up and nobody has done anything here, I might take a look at it. 

Would welcome some design guidance beforehand.

lornajane:
Thanks for suggesting this and giving the specific examples of the data that we could start with, it would be excellent to have this for Karapace!

chrisVia:
Lack of telemetry with Karapace is definitely a bit disappointing and would be a welcome addition! For some perspective, the other two mature projects in this space, RedHat's Apicurio and Confluent Schema-Registry both provide telemetry (Apicurio as prometheus metrics and SR via JMX MBeans).

YuWang24:
Hey, I am just wondering do we have metrics now? Something like the Confluent Schema-Registry's JMX MBeans metrics. 
I saw there is a StatsClient. What is this StatsClient used for? By looking at the code, is that only used for reporting the unexpected exceptions?

hackaugusto:
@YuWang24 , `StatsClient` is a class that can publish metrics using the [statsd](https://github.com/statsd/statsd) format and send tracebacks to sentry. But as you noticed there are not that many metrics being published as of now.

YuWang24:
@hackaugusto Thank you for answering.

varunghai-instaclustr:
Hey @lornajane, i noticed that some work on this has been recently done as part of https://github.com/aiven/karapace/pull/453. Would you be able to share any design or roadmap for further work on implementing support for additional metrics? We (Instaclustr) would be happy to help contribute towards it.

lornajane:
This is great! We're working on a roadmap to share for this project, should be something in the next few weeks and it would be good to get your input when we do.

varunghai-instaclustr:
Nice, thanks! We were in the early stages of planning work on this and having some insight into your roadmap/design will help.

varunghai-instaclustr:
@lornajane Any progress on the roadmap? Is there something we can do to help expedite it?

lornajane:
I'm going to loop in @filip-yonov-aiven here to get us something we can look at. I think it would be great to collaborate on our plans!

[...]

PR #902, opened 2024-06-18
feature, observability: Added Prometheus Metrics & Instrumentation

nosahama:

Add Observability to the Karapace services (`registry`, `rest`) via prometheus metrics and instrumentation.

# Why this way

Using prometheus, we can take advantage of it's labels, dimensional data model, available tools and ecosystem. This provides the building blocks of adding observability into the service, with focus on extensibility and testing (unit and integration).

## Caveats

- Right now, we have added a few metric types majorly for the HTTP requests, but this can be extended with more metrics for the different operations of the system, i.e. `schema reader` metrics, etc.
- We use the `karapace` prefix to act as a namespace and identifier for the service metrics.
- A sample prometheus service is added to the docker compose setup to show the pulled metrics via the `prometheus.scrape_configs`, the UI is shown in the screenshot below. 
- To support StatsD, we can add a [statsd-exporter](https://github.com/prometheus/statsd_exporter) service to the docker compose setup to test the provided mapping.
- We've added scrape jobs for both the `karapace-rest` and `karapace-registry` services, both services expose metrics via the `/metrics` endpoint. The screenshot below shows the local prometheus UI:

  Screenshot 2024-06-18 at 14 37 04

## Follow ups

- See if we want to toggle the metrics enablement via environment variables
- Add more metrics to the service, [checkout the confluent metrics for SR](https://docs.confluent.io/platform/current/schema-registry/monitoring.html)
- Configure buckets for the `Histogram` metric
- Add grafana dashboards to the docker compose setup
- Test the StatsD exporter if we truly want to still support StatsD
- Add a few prometheus recording rules and alerts

## References: 

- Task: [EC-299](https://aiven.atlassian.net/browse/EC-299)
- Solves: #204, but we would leave the issue open while me migrate existing metrics and implement new ones. 
- Closes: #711 #890

### DIFF
--- a/karapace/instrumentation/prometheus.py
+++ b/karapace/instrumentation/prometheus.py
@@ -4,13 +4,14 @@ karapace - prometheus instrumentation
 Copyright (c) 2024 Aiven Ltd
 See LICENSE for details
 """
+# mypy: disable-error-code="call-overload"
 
 from __future__ import annotations
 
-from aiohttp.web import middleware, Request, RequestHandler, Response
+from aiohttp.web import middleware, Request, Response
 from karapace.rapu import RestApp
 from prometheus_client import CollectorRegistry, Counter, Gauge, generate_latest, Histogram
-from typing import ClassVar
+from typing import Awaitable, Callable, Final
 
 import logging
 import time
@@ -19,34 +20,34 @@ LOG = logging.getLogger(__name__)
 
 
 class PrometheusInstrumentation:
-    METRICS_ENDPOINT_PATH: ClassVar[str] = "/metrics"
-    START_TIME_REQUEST_KEY: ClassVar[str] = "start_time"
+    METRICS_ENDPOINT_PATH: Final[str] = "/metrics"
+    START_TIME_REQUEST_KEY: Final[str] = "start_time"
 
-    registry: ClassVar[CollectorRegistry] = CollectorRegistry()
+    registry: Final[CollectorRegistry] = CollectorRegistry()
 
-    karapace_http_requests_total: ClassVar[Counter] = Counter(
+    karapace_http_requests_total: Final[Counter] = Counter(
         registry=registry,
         name="karapace_http_requests_total",
         documentation="Total Request Count for HTTP/TCP Protocol",
         labelnames=("method", "path", "status"),
     )
 
-    karapace_http_requests_latency_seconds: ClassVar[Histogram] = Histogram(
+    karapace_http_requests_duration_seconds: Final[Histogram] = Histogram(
         registry=registry,
-        name="karapace_http_requests_latency_seconds",
+        name="karapace_http_requests_duration_seconds",
         documentation="Request Duration for HTTP/TCP Protocol",
         labelnames=("method", "path"),
     )
 
-    karapace_http_requests_in_progress: ClassVar[Gauge] = Gauge(
+    karapace_http_requests_in_progress: Final[Gauge] = Gauge(
         registry=registry,
         name="karapace_http_requests_in_progress",
-        documentation="Request Duration for HTTP/TCP Protocol",
+        documentation="In-progress requests for HTTP/TCP Protocol",
         labelnames=("method", "path"),
     )
 
     @classmethod
-    def setup_metrics(cls: PrometheusInstrumentation, *, app: RestApp) -> None:
+    def setup_metrics(cls, *, app: RestApp) -> None:
         LOG.info("Setting up prometheus metrics")
         app.route(
             cls.METRICS_ENDPOINT_PATH,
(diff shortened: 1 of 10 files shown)


karapace/instrumentation/prometheus.py
```diff
@@ -0,0 +1,100 @@
+"""
+karapace - prometheus instrumentation
+
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+# mypy: disable-error-code="call-overload"
```
nosahama, line 7: As the `aiohttp.web.Application` and `aiohttp.web.Request` objects both support [collections.abc.MutableMapping](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping), we can set attributes at runtime, which is beneficial and used within the HTTP requests instrumentation. 
MyPy complains about missing `overloads` and as these object are within 3rd party library, we cannot update the declarations and thus I disable this error for the whole file.

karapace/instrumentation/prometheus.py
```diff
@@ -0,0 +1,100 @@
...
+        registry=registry,
+        name="karapace_http_requests_latency_seconds",
+        documentation="Request Duration for HTTP/TCP Protocol",
+        labelnames=("method", "path"),
+    )
+
+    karapace_http_requests_in_progress: ClassVar[Gauge] = Gauge(
+        registry=registry,
+        name="karapace_http_requests_in_progress",
+        documentation="Request Duration for HTTP/TCP Protocol",
```
nosahama, line 45: **[issue]**
small copy paste error here on the `documentation`, i'd fix after review.

karapace/instrumentation/prometheus.py
```diff
@@ -0,0 +1,100 @@
...
+
+import logging
+import time
+
+LOG = logging.getLogger(__name__)
+
+
+class PrometheusInstrumentation:
+    METRICS_ENDPOINT_PATH: ClassVar[str] = "/metrics"
+    START_TIME_REQUEST_KEY: ClassVar[str] = "start_time"
```
eliax1996, line 24: just wondering, why are you setting those to be variant?

tests/unit/instrumentation/test_prometheus.py
```diff
@@ -0,0 +1,119 @@
...
+
+import aiohttp.web
+import logging
+import pytest
+
+
+class TestPrometheusInstrumentation:
+    @pytest.fixture
+    def prometheus(self) -> PrometheusInstrumentation:
+        return PrometheusInstrumentation()
```
eliax1996, line 22: just a matter of taste, I prefer having just plain functions in the tests, they avoid creating implicit dependencies between the tests (by the state of the class). Since you aren't adding any state to this, do you mind change this and transform the tests that are methods of the class into plain functions? I'm not strict with this, just to avoid the temptation in the future of placing a state in the class :) (and obviously we need to be careful also about global variables, it doesn't remove completely the problem)

eliax1996 (dismissed):
Well done, other than this I see no reason for not merging! Great job

karapace/instrumentation/prometheus.py
nosahama: what do you mean variant?

eliax1996: `ClassVar`

eliax1996: ah sorry, missread, this is `ClassVar` to mark it as variant you need to mark it as `TypeVar`

eliax1996: its to indicate that all the instances share the same variable, make sense

eliax1996: not sure if you can combine it with `Final` but I would prefer it, does it make sense to change the content? I would suggest to change it into `ClassVar[Final[str]]` or `Final[str]` if the first isn't allowed

nosahama: I remember I ran into an issue with mypy complaining about combining both `ClassVar` and `Final`, ideally as this would not have an instance, except for the class instance itself, one would have to directly change the class variable and this we can disallow if seen in a review.
If i would change at all, i'd change to `Final[T]`, wdyt?

eliax1996: yeah, I would vote for changing into `Final`, we should understand why its complaining and probably there is a good reason 😄

tests/unit/instrumentation/test_prometheus.py
nosahama: I understand the reasoning. If you see that in the integration tests, I do not use a Test Class. I have this attitude towards unit tests, as we are mostly testing objects in isolation, thus it makes sense to group these tests, share dependencies between related tests, etc. 
Every test in this class `TestPrometheusInstrumentation` is only relating to the prometheus changes and thus you see the fixture is visually scoped to the class body. It's kind of a rationale similar to [unittest.TestCase](https://docs.python.org/3/library/unittest.html#unittest.TestCase) but very lightweight and more of a group for all related tests.

nosahama: Imagine we had other objects which are used for the metrics within the `prometheus.py` module, then we'd have a lot of `test_` functions, some of which at some point would not be related to the test above or below, i've found this sometimes to be confusing for unit tests. 
E2E/Integration tests on the other hand mostly encapsulate the whole test path within a single `test_` or multiple related consecutive tests functions.

eliax1996: I see, again no strong feeling about this :) your explaination make sense, the long result its what count, with both the approaches you can reach monsters or well written tests. I can mark this as resolved since isn't anything other than taste

tests/unit/instrumentation/test_prometheus.py
```diff
@@ -0,0 +1,119 @@
...
+    async def test_http_request_metrics_middleware(
+        self,
+        mock_time: MagicMock,
+        prometheus: PrometheusInstrumentation,
+    ) -> None:
+        mock_time.time.return_value = 10
+        request = AsyncMock(
+            spec=aiohttp.web.Request, app=AsyncMock(spec=aiohttp.web.Application), path="/path", method="GET"
+        )
+        handler = AsyncMock(spec=aiohttp.web.RequestHandler, return_value=MagicMock(status=200))
```
nosahama, line 91: **[issue]**
I need to assert that the `handler` is called with the request, i forgot this, will add.

PR #903, opened 2024-06-19
chore, rapu: handle `asyncio.TimeoutError`

nosahama:

Handle `asyncio.TimeoutError`.
- we also prevent reraising `asyncio.CanclledError` as it is not necessary within this context, this could cause the app to crash or other unexpected behaviour.

### DIFF
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -393,14 +393,12 @@ class RestApp:
             )
             headers = {"Content-Type": "application/json"}
             resp = aiohttp.web.Response(body=body, status=status.value, headers=headers)
-        except (ConnectionError, aiohttp.ClientError) as connection_error:
-            # TCP level connection errors, e.g. TCP reset, client closes connection.
-            self.log.debug("Connection error.", exc_info=connection_error)
+        except (ConnectionError, aiohttp.ClientError, asyncio.CancelledError, asyncio.TimeoutError) as exc:
+            # TCP level connection errors and timeouts, e.g. TCP reset, client closes connection, task takes too long.
+            error_msg = "Unexpected connection or timeout error"
+            self.log.debug(error_msg, exc_info=exc)
             # No response can be returned and written to client, aiohttp expects some response here.
-            resp = aiohttp.web.Response(text="Connection error", status=HTTPStatus.SERVICE_UNAVAILABLE.value)
-        except asyncio.CancelledError:
-            self.log.debug("Client closed connection")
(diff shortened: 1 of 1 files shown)


karapace/rapu.py
```diff
@@ -393,14 +393,12 @@ async def _handle_request(
...
-        except (ConnectionError, aiohttp.ClientError) as connection_error:
-            # TCP level connection errors, e.g. TCP reset, client closes connection.
-            self.log.debug("Connection error.", exc_info=connection_error)
+        except (ConnectionError, aiohttp.ClientError, asyncio.CancelledError, asyncio.TimeoutError) as exc:
+            # TCP level connection errors and timeouts, e.g. TCP reset, client closes connection, task takes too long.
+            error_msg = "Unexpected connection or timeout error"
+            self.log.debug(error_msg, exc_info=exc)
             # No response can be returned and written to client, aiohttp expects some response here.
-            resp = aiohttp.web.Response(text="Connection error", status=HTTPStatus.SERVICE_UNAVAILABLE.value)
-        except asyncio.CancelledError:
```
nosahama, line 401: No need to raise `asyncio.CancelledError` here, we can also relate it to the wildcard exception handler below, there we return a response, we should do the same here too, else, we risk possibly crashing the service.

karapace/rapu.py
```diff
@@ -393,14 +393,12 @@ async def _handle_request(
             )
             headers = {"Content-Type": "application/json"}
             resp = aiohttp.web.Response(body=body, status=status.value, headers=headers)
-        except (ConnectionError, aiohttp.ClientError) as connection_error:
-            # TCP level connection errors, e.g. TCP reset, client closes connection.
-            self.log.debug("Connection error.", exc_info=connection_error)
+        except (ConnectionError, aiohttp.ClientError, asyncio.CancelledError, asyncio.TimeoutError) as exc:
+            # TCP level connection errors and timeouts, e.g. TCP reset, client closes connection, task takes too long.
+            error_msg = "Unexpected connection or timeout error"
+            self.log.debug(error_msg, exc_info=exc)
```
eliax1996, line 399: maybe error log?

nosahama: `error` would go to sentry, as this is handled, we can bump to info if you want, but i'd leave as is for CI costs. Prometheus would handle these error metrics once we've added them.

eliax1996 (approved):
LGTM

PR #906, opened 2024-06-25
WIP

eliax1996:

References: #xxxxx

# Why this way


eliax1996:
We decided to reimplement the parsing part in Go, too big as an effort to not resolving completely all the issues in the parser. A rewrite its needed

PR #912, opened 2024-07-03
fix: get rid of the path for fully qualified names.

eliax1996:
This is not complete normalization feature but rather more a fix for the fully qualified paths vs simple name references. This fix not manage the various way a type can be expressed with a partial path.

Long explanation below:

If we accept the specifications from buf as correct, we can look at how a type reference is defined [here](https://protobuf.com/docs/language-spec#type-references).

The main problem with the previous implementation is that the current parser can't tell if a fully-qualified type reference in dot notation is the same as one identified by name alone aka simple name notation ([fully qualified reference](https://protobuf.com/docs/language-spec#fully-qualified-references)).

The fix do not consider all the different ways users can define a relative reference, schemas
 with different way of expressing a [relative reference](https://protobuf.com/docs/language-spec#relative-references) even if normalized,
 for now will keep being considered different.

Right now, our logic removes the `.package_name` (+ the Message scope) part
 before comparing field modifications (in fact it re-writes the schema using the simple name notation).

Even though the TypeTree (trie data structure) could help resolve [relative references](https://protobuf.com/docs/language-spec#relative-references), we don't want to add this feature in the python implementation now because it might cause new bugs due to the non-trivial behaviour of the protoc compiler.

We plan to properly normalize the protobuf schemas later.

We'll use protobuf descriptors (after the compilation and linking step) to gather type references already resolved, and we will threaten all the protobuf using always the fully qualified names.
Properly handling all path variations means reimplementing the protoc compiler behavior, we prefer relying on the already processed proto descriptor.

So, for now, we'll only implement a normalization for the fully-qualified references in dot notation and by simple name alone.

**NB:** This is not changing the semantics of the message since the local scope its always the one
 with max priority, so if you get rid of the fully-qualified reference protoc will resolve
 the reference with the one specified in the package scope.

References: #xxxxx

# Why this way

### DIFF
--- a/karapace/protobuf/proto_normalizations.py
+++ b/karapace/protobuf/proto_normalizations.py
@@ -5,12 +5,12 @@ See LICENSE for details
 
 from __future__ import annotations
 
-from karapace.dependency import Dependency
 from karapace.protobuf.enum_constant_element import EnumConstantElement
 from karapace.protobuf.enum_element import EnumElement
 from karapace.protobuf.extend_element import ExtendElement
 from karapace.protobuf.field_element import FieldElement
 from karapace.protobuf.group_element import GroupElement
+from karapace.protobuf.known_dependency import KnownDependency
 from karapace.protobuf.message_element import MessageElement
 from karapace.protobuf.one_of_element import OneOfElement
 from karapace.protobuf.option_element import OptionElement
@@ -19,8 +19,9 @@ from karapace.protobuf.rpc_element import RpcElement
 from karapace.protobuf.schema import ProtobufSchema
 from karapace.protobuf.service_element import ServiceElement
 from karapace.protobuf.type_element import TypeElement
-from karapace.schema_references import Reference
-from typing import Mapping, Sequence
+from karapace.protobuf.type_tree import TypeTree
+from karapace.utils import remove_prefix
+from typing import Sequence
 
 import abc
 
@@ -77,15 +78,9 @@ class NormalizedGroupElement(GroupElement):
 class NormalizedProtobufSchema(ProtobufSchema):
     proto_file_element: NormalizedProtoFileElement
 
-    def __init__(
-        self,
-        schema: str,
-        references: Sequence[Reference] | None = None,
-        dependencies: Mapping[str, Dependency] | None = None,
-        proto_file_element: ProtoFileElement | None = None,
-    ) -> None:
-        super().__init__(schema, references, dependencies, proto_file_element)
-        self.proto_file_element = normalize(self.proto_file_element)
+    @staticmethod
+    def from_protobuf_schema(protobuf_schema: ProtobufSchema) -> NormalizedProtobufSchema:
+        return normalize(protobuf_schema)
 
 
 class NormalizedOneOfElement(OneOfElement):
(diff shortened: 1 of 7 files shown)


karapace/protobuf/proto_normalizations.py
```diff
@@ -93,12 +93,17 @@ class NormalizedOneOfElement(OneOfElement):
     groups: Sequence[NormalizedGroupElement]
 
 
-def type_field_element_with_sorted_options(type_field: FieldElement) -> NormalizedFieldElement:
+def type_field_element_with_sorted_options(type_field: FieldElement, scope: str) -> NormalizedFieldElement:
     sorted_options = None if type_field.options is None else list(sorted(type_field.options, key=sort_by_name))
+    field_type_normalized = type_field.element_type.removeprefix(
+        "."
+    ).removeprefix(  # getting rid of the optional `.` before the package name
```
eliax1996, line 100: all the fully qualified references always start with a `.`, like for the example of: `.google.protobuf.Timestamp` that its referring to the package `google.protobuf`.
In this case normalizing it by removing the `.` can lead to shadowing if the import of the google message its after another import that defines a scope that contains google, like for e.g.:

```
package my.awesome.tricky.company.google.protobuf;

message Timestamp {
   // definition...
}
```

Even if this is possible its unlikely to happen, we may even reduce that removal of the `.` just for types that do not have other anchestors in type type trie

eliax1996: Implemented it by the type tree and implemented a test

karapace/protobuf/proto_normalizations.py
```diff
@@ -77,28 +78,41 @@ class NormalizedGroupElement(GroupElement):
...
+        return normalize(protobuf_schema)
 
 
 class NormalizedOneOfElement(OneOfElement):
     fields: Sequence[NormalizedFieldElement]
     groups: Sequence[NormalizedGroupElement]
 
 
-def type_field_element_with_sorted_options(type_field: FieldElement) -> NormalizedFieldElement:
+def type_field_element_with_sorted_options(
```
keejon, line 91: Should we consider renaming this to `normalize_type_field_element` as it is not exclusively sorting options anymore?

tests/unit/protobuf/test_protobuf_normalization.py
```diff
@@ -511,10 +516,200 @@
...
+    tricky_dep = Dependency("TrickyNestedValue.proto", Subject("tricky_nested_value"), Version(1), tricky_no_ref_schema)
+    dependencies = {"NestedValue.proto": dep, "TrickyNestedValue.proto": tricky_dep}
+    fully_qualitifed_dot_notation = parse_protobuf_schema_definition(
+        schema_definition=PROTO_WITH_FULLY_QUALIFIED_PATHS_AND_TRICKY_DEPENDENCY,
+        references=None,
+        dependencies=dependencies,
+        validate_references=True,
+        normalize=True,
+    )
+    normalized_schema = normalize(fully_qualitifed_dot_notation).to_schema()
```
keejon, line 713: isn't this normalized already because `normalze=True`?

eliax1996: ah yeah, you are right

keejon (approved):
LGTM 👍

PR #915, opened 2024-07-12
POC: use Buf and protocompile for protobuf schemas

keejon:
This is a POC to validate if it is feasible to replace existing protobuf parser with protocompile and compatibility checks with Buf breaking changes.


muralibasani:
@keejon we can try recreating this instead of fixing conflicts. what do you think. 
If so, I can take a look.

keejon:
Implemented in this PR: https://github.com/Aiven-Open/karapace/pull/933

PR #920, opened 2024-07-23
fix: change schema coordinator error logs to warning

jjaakola-aiven:
The logs changed are warnings, the coordinator should correct the issue.

References: #xxxxx

# Why this way


eliax1996 (approved):
Change its good, would be nice to have a metric about how long this state its kept and raise an escalation if its more than a threshold

keejon (approved):
LGTM 👍

PR #921, opened 2024-07-24
fix: catch all exceptions in forked process

jjaakola-aiven:
If base exception is raised the communication queue will block in the calling process and will cause timeouts. This is prevalent in unit test for protobuf serialization in Github Actions, caused by the protobuf compiler work directory existence. The fix is to use pathlib.Path with exists=True and parents=True for removing the FileExistsError and also changing the error handling logic to pass back also BaseExceptions through the multiprocess queue.

References: #xxxxx

# Why this way

### DIFF
--- a/karapace/protobuf/io.py
+++ b/karapace/protobuf/io.py
@@ -13,13 +13,13 @@ from karapace.protobuf.protobuf_to_dict import dict_to_protobuf, protobuf_to_dic
 from karapace.protobuf.schema import ProtobufSchema
 from karapace.protobuf.type_element import TypeElement
 from multiprocessing import Process, Queue
+from pathlib import Path
 from typing import Dict, Final, Generator, Iterable, Protocol
 from typing_extensions import Self, TypeAlias
 
 import hashlib
 import importlib
 import importlib.util
-import os
 import subprocess
 import sys
 
@@ -96,7 +96,7 @@ def get_protobuf_class_instance(
     class_name: str,
     cfg: Config,
 ) -> _ProtobufModel:
-    directory = cfg["protobuf_runtime_directory"]
+    directory = Path(cfg["protobuf_runtime_directory"])
     deps_list = crawl_dependencies(schema)
     root_class_name = ""
     for value in deps_list.values():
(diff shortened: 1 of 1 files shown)


karapace/protobuf/io.py
```diff
@@ -127,16 +125,18 @@ def get_protobuf_class_instance(
...
                 cwd=work_dir,
             )
 
-    # todo: This will leave residues on sys.path in case of exceptions. If really must
-    #  mutate sys.path, we should at least wrap in try-finally.
-    sys.path.append(f"./runtime/{proto_name}")
+    runtime_proto_path = f"./runtime/{proto_name}"
+    if runtime_proto_path not in sys.path:
+        # todo: This will leave residues on sys.path in case of exceptions. If really must
+        # mutate sys.path, we should at least wrap in try-finally.
```
eliax1996, line 138: cannot we do as final action a tree delete marking the protobuf folder as the father folder to delete? In that way (unless the path of the dependencies contain `../../name` we can delete everything eventually)

eliax1996: I know you didn't introduce the comment but can be solved in that way

karapace/protobuf/io.py
```diff
@@ -233,34 +237,37 @@ def __init__(
     def read(self, bio: BytesIO) -> dict:
         if self._reader_schema is None:
             self._reader_schema = self._writer_schema
-        return reader_mp(self.config, self._writer_schema, self._reader_schema, bio)
+        return read_in_forked_process_multiprocess_process(self.config, self._writer_schema, self._reader_schema, bio)
```
eliax1996, line 240: good rename!

karapace/protobuf/io.py
```diff
@@ -274,14 +281,18 @@ def writer_mp(
     # To avoid problem with enum values for basic SerDe support we
     # will isolate work with call protobuf libraries in child process.
     if __name__ == "karapace.protobuf.io":
-        queue: _WriterQueue = Queue()
-        p = Process(target=writer_process, args=(queue, config, writer_schema, message_name, datum))
+        writer_queue: _WriterQueue = Queue(1)
```
eliax1996, line 284: why only 1 message in the queue?

eliax1996: does it make sense to proceed with just 1 at a time?

eliax1996 (commented):
Everything fine except for the queue size, not sure if the change for setting the `max_size` to `1` can decrease the performances

karapace/protobuf/io.py
jjaakola-aiven: I think the protobuf files is ok to keep, if the one exists it will be reused and `protoc` is not called.
The mangling of `sys.path` and the lines below should be wrapped in `try-except-else-finally` block to remove the added path if something fails. But not removed if it already was there.

karapace/protobuf/io.py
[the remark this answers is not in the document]
jjaakola-aiven: I could remove the extra `_process_` though.

karapace/protobuf/io.py
jjaakola-aiven: There is nothing else put to queue than one message at a time which the caller waits for. I decided just to make the size explicit here.

karapace/protobuf/io.py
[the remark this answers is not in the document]
jjaakola-aiven: Done.

karapace/protobuf/io.py
eliax1996: didn't notice wasn't reused and we have 1 process for each message

eliax1996 (approved):

PR #924, opened 2024-07-25
fix: `Version` int log formatting in `schema_reader.py`

nosahama:
We use the repr `%r` formatting instead of int `%d` for the affected log line and add a test to catch this and act as a reference for testing log statements. 

### Related traceback
```
TypeError: %d format: a real number is required, not Version')
```


eliax1996 (approved):
LGTM great! Nice also the test

jjaakola-aiven (approved):

PR #926, opened 2024-07-27
Support of References for AVRO Schemas 

libretto:
# This change adds support for references for AVRO schemas

# We aim to add support for references for AVRO schemas to Karapace using a straightforward approach by leveraging the same API used in Protobuf references. Although Avro does not natively support references, we can implement this functionality by utilizing the property of Avro schemas that allows combining types. By including multiple types in unions, we can enable referencing from one union element to another.


jjaakola-aiven (changes requested):
Cleanup commits to single/coherent commits. Add tests for better coverage.

libretto:
> Cleanup commits to single/coherent commits. Add tests for better coverage.
Could you suggest the tests that should be added?

libretto:
@jjaakola-aiven I'm still waiting for your suggestions regarding additional tests for the new functionality. Please keep in mind that we already have integration tests covering various aspects, including:

    Basic Avro references
    Avro references compatibility
    Avro union references
    Avro incompatible names in references

Additionally, the new code has successfully passed all existing tests related to Avro functionality. I also added a few unit tests for the AvroMerge class today.

karapace/schema_models.py
```diff
@@ -180,6 +183,44 @@ def schema(self) -> Draft7Validator | AvroSchema | ProtobufSchema:
...
 
+class AvroMerge:
+    def __init__(self, schema_str: str, dependencies: Mapping[str, Dependency] | None = None):
+        self.schema_str = json_encode(json_decode(schema_str), compact=True, sort_keys=True)
+        self.dependencies = dependencies
+        self.unique_id = 0
+
+    def union_safe_schema_str(self, schema_str: str) -> str:
+        # in case we meet union - we use it as is
+        regex = re.compile(r"^\s*\[")
```
nosahama, line 194: As this will be called multiple times, maybe move this regex compilation out of the function and into the `__init__()`, else, we compile every time this is called.

karapace/schema_registry_apis.py
```diff
@@ -1055,7 +1055,7 @@ def _validate_references(
                 content_type=content_type,
                 status=HTTPStatus.BAD_REQUEST,
             )
-        if references and schema_type != SchemaType.PROTOBUF:
+        if references and schema_type != SchemaType.PROTOBUF and schema_type != SchemaType.AVRO:
```
nosahama, line 1058: This condition is wrong, we will never have both schema types within the same request, thus this will not work.

karapace/schema_models.py
```diff
@@ -180,6 +183,44 @@ def schema(self) -> Draft7Validator | AvroSchema | ProtobufSchema:
...
+    def builder(self, schema_str: str, dependencies: Mapping[str, Dependency] | None = None) -> str:
+        """To support references in AVRO we iteratively merge all referenced schemas with current schema"""
+        stack: list[tuple[str, Mapping[str, Dependency] | None]] = [(schema_str, dependencies)]
+        merged_schemas = []
+
+        while stack:
+            current_schema_str, current_dependencies = stack.pop()
+            if current_dependencies:
+                stack.append((current_schema_str, None))
+                for dependency in reversed(current_dependencies.values()):
```
nosahama, line 212: Why are you doing the reverse?

karapace/schema_models.py
```diff
@@ -188,21 +229,41 @@ def parse(
     references: Sequence[Reference] | None = None,
     dependencies: Mapping[str, Dependency] | None = None,
     normalize: bool = False,
+    dependencies_compat: bool = False,
```
nosahama, line 232: I am guessing this is some flag you want to introduce so we can opt-in/out of the referencing logic, as it is it is hard coded, so maybe you can add it to the config.

karapace/schema_models.py
```diff
@@ -188,21 +229,41 @@ def parse(
...
 ) -> ParsedTypedSchema:
     if schema_type not in [SchemaType.AVRO, SchemaType.JSONSCHEMA, SchemaType.PROTOBUF]:
         raise InvalidSchema(f"Unknown parser {schema_type} for {schema_str}")
-
+    parsed_schema_result: Draft7Validator | AvroSchema | ProtobufSchema
     parsed_schema: Draft7Validator | AvroSchema | ProtobufSchema
     if schema_type is SchemaType.AVRO:
         try:
+            if dependencies or dependencies_compat:
```
nosahama, line 240: So I could have `dependencies` but not opt-in to the `dependencies_compat` and you will still run this logic, so what is the essence of `dependencies_compat`?

karapace/schema_models.py
```diff
@@ -188,21 +229,41 @@ def parse(
...
+                wrapped_schema_str = schema_str
             parsed_schema = parse_avro_schema_definition(
-                schema_str,
+                wrapped_schema_str,
                 validate_enum_symbols=validate_avro_enum_symbols,
                 validate_names=validate_avro_names,
             )
+            if dependencies or dependencies_compat:
+                if isinstance(parsed_schema, avro.schema.UnionSchema):
+                    parsed_schema_result = parsed_schema.schemas[-1].fields[0].type.schemas[-1]
```
nosahama, line 251: Please what is going on here? Can you give some hint as to what you want to achieve, taking into account also the checks above?
```python
if dependencies or dependencies_compat:
    if isinstance(parsed_schema, avro.schema.UnionSchema):
```

karapace/schema_models.py
```diff
@@ -264,9 +325,10 @@ def __init__(
         schema: Draft7Validator | AvroSchema | ProtobufSchema,
         references: Sequence[Reference] | None = None,
         dependencies: Mapping[str, Dependency] | None = None,
+        schema_wrapped: Draft7Validator | AvroSchema | ProtobufSchema | None = None,
```
nosahama, line 328: This and it's respective instance variables are not used anywhere in the code, why are you adding this?

tests/integration/test_schema_avro_references.py
```diff
@@ -0,0 +1,295 @@
...
+    "namespace": "com.netapp",
+    "fields": [
+        {"name": "name", "type": "string"},
+        {"name": "age", "type": "int"},
+        {"name": "address", "type": "Address"},
+        {"name": "job", "type": "Job"},
+    ],
+}
+
+SCHEMA_PERSON_INT_LONG = {
```
nosahama, line 57: `_INT_LONG` is a bit hard to know what it should be doing, I see you updated the `age` type to `long`, maybe `_AGE_LONG`

tests/integration/test_schema_avro_references.py
```diff
@@ -0,0 +1,295 @@
...
+        "namespace": "com.netapp",
+        "fields": [
+            {"name": "name", "type": "string"},
+            {"name": "age", "type": "int"},
+            {"name": "address", "type": "Address"},
+        ],
+    },
+]
+
+SCHEMA_ADDRESS_INCOMPATIBLE = {
```
nosahama, line 128: I cannot find the incompatibility in this schema.

nosahama:
@libretto shall we rebase this work so we get upstream changes and then let's discuss around the comments if you can tell me when you'd like to, we can spare sometime and discuss in one thread here.

karapace/schema_registry_apis.py
nosahama: Let me take another look, because I see you also use the API client in the integration tests and if the tests pass then this should be correct.

karapace/schema_models.py
libretto: Using reverse allows us to preserve the original order of including dependencies in the merged schema. This approach was chosen because the performance of using append() and pop() in lists is better than that of insert(), plus the merging code looks slightly more elegant.

karapace/schema_models.py
libretto: Seems I forgot to remove it from the code. It was used in some iteration of the code.

karapace/schema_models.py
libretto: "To achieve schema merging with references (dependencies), we wrap the top-level schema and its references in a special way, then merge all schemas into a Union, where the top-level schema is the last, and all reference schemas are appended in a specific order with the rule that all referenced schemas must appear higher in the Union list than their referers (I mentioned this order in my answer about 'reverse' above). Then we parse this merged schema, resulting in an avro.schema.UnionSchema object, where the top-level schema is last in the union.

The AVRO object is constructed in such a way that dependent schemas are stored inside the AVRO object, and the top-level schema references them in a way that allows schema comparison without issues. However, this union also includes all dependency items at the same level as the top-level schema. So, we remove these items and extract only the top-level schema object from the union. All references remain present and work perfectly. As a result, we have no union, but we have a correct AVRO schema with dependencies.

However, there is a case where the top-level schema may have no references but is a Union itself. In this case, the code above checks if we have a UnionSchema as a result, and if this schema has references, we extract the top-level schema from it. If it is a Union without references, then we do nothing."

karapace/schema_models.py
libretto: Seems it must be removed.

karapace/schema_registry_apis.py
nosahama: I guess the two `and` bugs me out, maybe a check for inclusion, i.e. `if references and (schema_type not in {SchemaType.PROTOBUF, SchemaType.AVRO})`, but it's also fine as the tests confirm. 
I think when I was testing the API via postman against the docker compose of your fork, the request might have still been hitting the docker compose I had running via the upstream.

tests/integration/test_schema_avro_references.py
libretto: It simple test which shows compatibility.

tests/integration/test_schema_avro_references.py
libretto: This test shows incompatibility in the naming of the field... We changed the name from "Address" to "ChangeAddress".

karapace/schema_models.py
libretto: Yes, i see.

tests/unit/test_avro_merge.py
```diff
@@ -0,0 +1,83 @@
...
+    @pytest.fixture
+    def setup_avro_merge(self):
+        schema_str = '{"type": "record", "name": "Test", "fields": [{"name": "field1", "type": "string"}]}'
+        dependencies = {"dependency1": MagicMock(schema=MagicMock(schema_str='{"type": "string"}', dependencies=None))}
+        return AvroMerge(schema_str, dependencies)
+
+    def test_init(self, setup_avro_merge):
+        avro_merge = setup_avro_merge
+        assert avro_merge.schema_str == json_encode(json_decode(avro_merge.schema_str), compact=True, sort_keys=True)
+        assert avro_merge.dependencies == avro_merge.dependencies
```
nosahama, line 26: You are asserting the same thing.

tests/unit/test_avro_merge.py
```diff
@@ -0,0 +1,83 @@
...
+from unittest.mock import MagicMock
+
+import pytest
+
+
+class TestAvroMerge:
+    @pytest.fixture
+    def setup_avro_merge(self):
+        schema_str = '{"type": "record", "name": "Test", "fields": [{"name": "field1", "type": "string"}]}'
+        dependencies = {"dependency1": MagicMock(schema=MagicMock(schema_str='{"type": "string"}', dependencies=None))}
```
nosahama, line 20: Do you really need to mock the dependency in this simple case?

tests/unit/test_avro_merge.py
```diff
@@ -0,0 +1,83 @@
...
+        return AvroMerge(schema_str, dependencies)
+
+    def test_init(self, setup_avro_merge):
+        avro_merge = setup_avro_merge
```
nosahama, line 24: this is not necessary.

nosahama:
The unit test for `AvroMerge` is gone and it seems that you have reverted some of your changes with your recent push.

nosahama:
@libretto The unit test for `AvroMerge` is gone and it seems that you have reverted some of your changes with your recent push.

libretto:
@nosahama 
> @libretto The unit test for `AvroMerge` is gone and it seems that you have reverted some of your changes with your recent push.

Yes it was an issue with rebase I just restored banch, synced it with main and try rebase again...

tests/unit/test_avro_merge.py
libretto: hmm yes, removed

tests/unit/test_avro_merge.py
libretto: It is a basic unit test for the AvroMerge functionality.

tests/unit/test_avro_merge.py
```diff
@@ -0,0 +1,83 @@
...
+        assert avro_merge.schema_str == json_encode(json_decode(avro_merge.schema_str), compact=True, sort_keys=True)
+        assert avro_merge.dependencies == avro_merge.dependencies
+        assert avro_merge.unique_id == 0
+
+    def test_union_safe_schema_str_no_union(self, setup_avro_merge):
+        avro_merge = setup_avro_merge
+        result = avro_merge.union_safe_schema_str('{"type": "string"}')
```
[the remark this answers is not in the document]
libretto: it is a basic AVRO schema that represents a string... In this case, we test avro_merge so it is acceptable for this test.

tests/unit/test_avro_merge.py
libretto: Probably. Need to test. I do not remember why I used such construction...

karapace/schema_models.py
libretto: fixed

tests/integration/test_schema_avro_references.py
libretto: fixed

libretto:
> @libretto shall we rebase this work so we get upstream changes and then let's discuss around the comments if you can tell me when you'd like to, we can spare sometime and discuss in one thread here.

rebased

davide-armand:
@libretto Hi, sorry for the delay, this took some time to properly review.

I think that the implementation using Avro unions is risky and might not work in all cases (hard to tell).
A more "standard" solution using the Avro Python library instead of unions seems a better choice.

I have opened a draft PR implementing the merging using Avro lib here:
https://github.com/Aiven-Open/karapace/pull/987

Please have a look and consider that approach instead.

libretto:
https://github.com/Aiven-Open/karapace/pull/987# implemented there

src/karapace/schema_models.py
```diff
@@ -207,18 +242,33 @@ def parse(
...
+                schemas_list = AvroResolver(schema_str, dependencies).resolve()
+                names = AvroNames(validate_names=validate_avro_names)
+                merged_schema = None
+                for schema in schemas_list:
+                    # Merge dep with all previously merged ones
+                    merged_schema = make_avsc_object(json.loads(schema), names)
+                merged_schema_str = str(merged_schema)
+            else:
+                merged_schema_str = schema_str
             parsed_schema = parse_avro_schema_definition(
```
davide-armand, line 258: I have realized now that it is possible to pass a list of schema strings to `avro.schema.parse()`:
https://github.com/aiven/avro/blob/5a82d57f2a650fd87c819a30e433f1abb2c76ca2/lang/py/avro/schema.py#L1192

We currently use `parse()` to parse a single schema (see `parse_avro_schema_definition()`).

When called with a list of schemas it returns a `UnionSchema` (through `make_avsc_object()`), which itself contains a list of parsed schemas (`UnionSchema.schemas`).
The construct of `UnionSchema` seems to be doing what we want (loop on the schemas, call `make_avsc_object()` and remember `names`).

I think passing to `parse()` the reference schemas + the main schema (the main schema should be passed as last in the list) and then getting the last schema from `UnionSchema.schemas` should be equivalent to what's the PR now.

If it works then we can delegate that logic to the Avro library

davide-armand (commented):
@libretto Hi, I left a few comments, please have a look

src/karapace/schema_models.py
```diff
@@ -195,6 +199,37 @@ def schema(self) -> Draft7Validator | AvroSchema | ProtobufSchema:
         return parsed_typed_schema.schema
 
 
+class AvroResolver:
+    def __init__(self, schema_str: str, dependencies: Mapping[str, Dependency] | None = None):
+        self.schema_str = json_encode(json_decode(schema_str), compact=True, sort_keys=True)
+        self.dependencies = dependencies
+        self.unique_id = 0
```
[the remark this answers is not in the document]
libretto: fixed

src/karapace/schema_models.py
```diff
@@ -195,6 +199,37 @@ def schema(self) -> Draft7Validator | AvroSchema | ProtobufSchema:
...
+        self.regex = re.compile(r"^\s*\[")
```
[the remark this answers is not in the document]
libretto: fixed

src/karapace/schema_models.py
```diff
@@ -195,6 +199,37 @@ def schema(self) -> Draft7Validator | AvroSchema | ProtobufSchema:
...
+                stack.append((current_schema_str, None))
+                for dependency in reversed(current_dependencies.values()):
+                    stack.append((dependency.schema.schema_str, dependency.schema.dependencies))
+            else:
+                self.unique_id += 1
+                merge.append(current_schema_str)
+
+        return merge
+
+    def resolve(self) -> list:
```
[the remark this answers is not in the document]
libretto: @davide-armand After this change, we have a pylint issue but all tests are passed. Please check.

cjrolo:
Hello, 
I would like to know if it is possible to have this re-reviewed? It seems that feedback was incorported into the lastest changes.

Thanks,

Carlos

src/karapace/schema_models.py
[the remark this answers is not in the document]
davide-armand: @libretto  Do you mean the error `No overload variant of "reversed"`?
It can be fixed by passing a list to `reversed`

tests/test_avro_references.py
```diff
@@ -0,0 +1,339 @@
...
+
+def test_avro_reference() -> None:
+    country_schema = ValidatedTypedSchema.parse(
+        schema_type=SchemaType.AVRO,
+        schema_str=textwrap.dedent(
+            """\
+            {
+                "type": "record",
+                "name": "Country",
+                "namespace": "com.netapp",
```
davide-armand, line 107: Please don't use reference to real companies (`com.netapp`).
Applies to in other fixtures as well.

cjrolo: Thanks! Will look into this!

cjrolo:
Fixed accordingly to the review. Hope I didn't miss anything, let me know!

Thanks for the help!

davide-armand:
@cjrolo @libretto
Thanks.
Please solve the conflicts and I will have a final review.

davide-armand:
@cjrolo @libretto Hi, Just a reminder about the PR conflicts.
The PR looks good but it is blocked by that.
Thanks!

PR #934, opened 2024-08-15
chore, schema-coordinator: handle `aiokafka` retriable errors

nosahama:

Handle `aiokafka` retriable errors, we see from [here](https://github.com/aio-libs/aiokafka/blob/master/aiokafka/consumer/consumer.py#L680) the `coordinator.check_errors()` is used in the consumer when getting the position, so this will always raise the errors that are registered with `coordinator._push_error_to_user()`, that way we do not flood application monitoring systems and instead log a warning.

# Service logs
Showing the now `warning` logs for the `GroupCoordinatorNotAvailableError`, we can also add a catch for `UnknownTopicOrPartitionError` which gets raised [here](https://github.com/Aiven-Open/karapace/blob/main/karapace/schema_reader.py#L351).

Screenshot 2024-08-20 at 15 23 05

We also reduce the log level for the `master_coordinator` bootstrap errors as they [will be retried]
(https://github.com/Aiven-Open/karapace/blob/main/karapace/coordinator/master_coordinator.py#L54)

Screenshot 2024-08-20 at 15 44 23

### DIFF
--- a/karapace/coordinator/master_coordinator.py
+++ b/karapace/coordinator/master_coordinator.py
@@ -50,7 +50,7 @@ class MasterCoordinator:
                 await self._kafka_client.bootstrap()
                 break
             except KafkaConnectionError:
-                LOG.exception("Kafka client bootstrap failed.")
+                LOG.warning("Kafka client bootstrap failed.")
                 await asyncio.sleep(0.5)
 
         while not self._kafka_client.cluster.brokers():
--- a/karapace/coordinator/schema_coordinator.py
+++ b/karapace/coordinator/schema_coordinator.py
@@ -181,6 +181,10 @@ class SchemaCoordinator:
 
         self._metadata_snapshot: list[Assignment] = []
 
+        self.SCHEMA_COORDINATOR_RETRIEABLE_ERRORS: set[Errors.KafkaError] = {
+            Errors.NodeNotReadyError,
+        }
+
     def start(self) -> None:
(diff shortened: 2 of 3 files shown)


karapace/coordinator/schema_coordinator.py
```diff
@@ -227,6 +231,9 @@ def check_errors(self) -> None:
         if self._pending_exception is not None:
             exc = self._pending_exception
             self._pending_exception = None
+            if (exc in self.SCHEMA_COORDINATOR_RETRIEABLE_ERRORS) or (getattr(exc, "retriable", False)):
```
nosahama, line 234: Should we also use something like, [coordinator_dead()](https://github.com/Aiven-Open/karapace/blob/main/karapace/coordinator/schema_coordinator.py#L429) or [request_rejoin()](https://github.com/Aiven-Open/karapace/blob/main/karapace/coordinator/schema_coordinator.py#L447), ideally the coordinator will retry some errors.

karapace/coordinator/master_coordinator.py
```diff
@@ -50,7 +50,7 @@ async def start(self) -> None:
                 await self._kafka_client.bootstrap()
                 break
             except KafkaConnectionError:
-                LOG.exception("Kafka client bootstrap failed.")
```
nosahama, line 53: Majorly for sentry denoising.

PR #935, opened 2024-08-16
test pr: copy content of branch protobuf formatter normalization

eliax1996:

References: #xxxxx

# Why this way


eliax1996:
The CI doesn't start. We need to figure out why

PR #936, opened 2024-08-21
schema-reader: Shutdown service if corrupt entries in `_schemas` topic

nosahama:

#### These breaking changes are guarded by environmental variables

These can be toggled by the various users, the default values are shown below:
```yaml
KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE: false
KARAPACE_KAFKA_RETRIABLE_ERRORS_SILENCED: true
```

#### The logic below only applies when `KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE` is set to `true`, everything else remains the same, thus no test changes.

Previously, when we encounter errors within the `_schemas` topic, we would continue the message loading and skip the problematic schema. This is not ideal as it might leave the application with corrupt schema data and the side-effects could be grave. 

What we do now is to kill the service, log the errors and allow a graceful shutdown. We will follow this work by adding metrics for such cases.

**This will also stop the service post backup-v1 restore if there are any corrupt schemas present in the backup log file**

We can see below that the shutdown is graceful:
Screenshot 2024-08-21 at 15 15 24

Adding `restart: always` to docker compose shows the behaviour, the service never proceeds past that stage, i think based on the restart behaviour for `systemd`, we might need to rely on the alerts and then intervene otherwise it'd leave the service in a crash loop, we need to verify if there are SLOs or metrics setup somewhere to track at least service uptime, which will be affected by this.

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -80,6 +80,8 @@ services:
       KARAPACE_COMPATIBILITY: FULL
       KARAPACE_STATSD_HOST: statsd-exporter
       KARAPACE_STATSD_PORT: 8125
+      KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE: false
+      KARAPACE_KAFKA_RETRIABLE_ERRORS_SILENCED: true
 
   karapace-rest:
     image: ghcr.io/aiven-open/karapace:develop
@@ -106,6 +108,8 @@ services:
       KARAPACE_LOG_LEVEL: WARNING
       KARAPACE_STATSD_HOST: statsd-exporter
       KARAPACE_STATSD_PORT: 8125
+      KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE: false
+      KARAPACE_KAFKA_RETRIABLE_ERRORS_SILENCED: true
 
   prometheus:
     image: prom/prometheus
--- a/karapace/config.py
+++ b/karapace/config.py
@@ -80,6 +80,8 @@ class Config(TypedDict):
     protobuf_runtime_directory: str
     statsd_host: str
     statsd_port: int
+    kafka_schema_reader_strict_mode: bool
+    kafka_retriable_errors_silenced: bool
 
     sentry: NotRequired[Mapping[str, object]]
     tags: NotRequired[Mapping[str, object]]
@@ -154,6 +156,8 @@ DEFAULTS: ConfigDefaults = {
     "protobuf_runtime_directory": "runtime",
     "statsd_host": "127.0.0.1",
     "statsd_port": 8125,
+    "kafka_schema_reader_strict_mode": False,
+    "kafka_retriable_errors_silenced": True,
 }
 SECRET_CONFIG_OPTIONS = [SASL_PLAIN_PASSWORD]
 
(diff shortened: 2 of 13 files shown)


karapace/schema_reader.py
```diff
@@ -352,19 +357,19 @@ def handle_messages(self) -> None:
 
                 assert message_key is not None
                 key = json_decode(message_key)
-            except JSONDecodeError:
+            except JSONDecodeError as exc:
                 # Invalid entry shall also move the offset so Karapace makes progress towards ready state.
```
jjaakola-aiven, line 361: Comment is wrong after this change, there is no need to move offset as Karapace exits.

karapace/schema_reader.py
```diff
@@ -242,6 +244,9 @@ def run(self) -> None:
                     self.offset = self._get_beginning_offset()
                 try:
                     self.handle_messages()
+                except ShutdownException:
+                    self._stop_schema_reader.set()
+                    shutdown()
                 except Exception as e:  # pylint: disable=broad-except
```
jjaakola-aiven, line 250: Adding the shutdown exception raises the question that what would be the action to take on the fallback `Exception`?

nosahama: Yes totally agree 👍 , normally in scenarios like this, we should fire a metric to track the exceptions and then form an alert around the various exceptions. We can perform another iteration, ideally, this work will trigger the review of our instrumentation.

container/compose.yml
```diff
@@ -106,6 +108,8 @@ services:
       KARAPACE_LOG_LEVEL: WARNING
       KARAPACE_STATSD_HOST: statsd-exporter
       KARAPACE_STATSD_PORT: 8125
+      KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE: false
```
nosahama, line 111: These might not apply on the rest proxy, but doesn't hurt and I didn't verify.

container/compose.yml
```diff
@@ -80,6 +80,8 @@ services:
       KARAPACE_COMPATIBILITY: FULL
       KARAPACE_STATSD_HOST: statsd-exporter
       KARAPACE_STATSD_PORT: 8125
+      KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE: false
+      KARAPACE_KAFKA_RETRIABLE_ERRORS_SILENCED: true
```
MichaelTansiniAiven, line 84: n00b question, why is this true by default?

nosahama: not at all, it's a valid question. So this doesn't affect the business logic of the service, so we enable it by default. This will cause us to reduce the alerts on our issue tracking (i.e. sentry) and as these errors are retried, it fine to silence them if the user wants, once again, they can opt out from this.

karapace/config.py
```diff
@@ -80,6 +80,8 @@ class Config(TypedDict):
     protobuf_runtime_directory: str
     statsd_host: str
     statsd_port: int
+    kafka_schema_reader_strict_mode: bool
+    kafka_retriable_errors_silenced: bool
```
jjaakola-aiven, line 84: The require documentation in the `README.md`.

jjaakola-aiven (approved):
LGTM

PR #938, opened 2024-08-21
fix: duplicated schemas on rapid elections while continuous produce of records

eliax1996:
coordinator rewrite
1. moving the coordinator in a separate thread
2. adding a waiting time between when the master its elected and the master can act. This has been done to avoid rapid elections of master that may produce schemas with different ids.

Example of what could happpen without the delay:

```md
|--------------------------------------|
|Node | Node1    | Node2    | Node3    |
|Role | Master   | Follower | Follower |
|--------------------------------------|


Node1 -> Send Message A{id=max(current_ids)} to kafka

where the max(current_ids) = 10

---------------------------------------

Node1 its disconnected, the message its still in the producer queue of Node1

---------------------------------------

Node2 its elected master

|--------------------------------------|
|Node | Node1    | Node2    | Node3    |
|Role | Follower | Master   | Follower |
|--------------------------------------|

----------------------------------------


Node2 produces a message B{id=max(current_ids)} to kafka

Because the message A isn't yet delivered to Node2, the max(current_ids) returns still 10.
And we have an ID clash.
```

The solution its simple, each master should wait a reasonable high number of milliseconds before acting as a master.
So that all the in-flight messages are delivered to kafka + the reasonable delay of the consumer for the master node before noticing that a message has been produced

### DIFF
--- a/src/karapace/coordinator/master_coordinator.py
+++ b/src/karapace/coordinator/master_coordinator.py
@@ -13,25 +13,42 @@ from aiokafka.protocol.commit import OffsetCommitRequest_v2 as OffsetCommitReque
 from karapace.config import Config
 from karapace.coordinator.schema_coordinator import SchemaCoordinator, SchemaCoordinatorStatus
 from karapace.kafka.types import DEFAULT_REQUEST_TIMEOUT_MS
+from karapace.typing import SchemaReaderStoppper
+from threading import Thread
 from typing import Final
 
 import asyncio
 import logging
+import time
 
 __all__ = ("MasterCoordinator",)
+
 
 LOG = logging.getLogger(__name__)
 
 
 class MasterCoordinator:
-    """Handles primary election"""
+    """Handles primary election
+
+    The coordination is run in own dedicated thread, under stress situation the main
+    eventloop could have queue of items to work and having own thread will give more
+    runtime for the coordination tasks as Python intrepreter will switch the active
+    thread by the configured thread switch interval. Default interval in CPython is
+    5 milliseconds.
+    """
 
     def __init__(self, config: Config) -> None:
         super().__init__()
         self._config: Final = config
         self._kafka_client: AIOKafkaClient | None = None
-        self._running = True
         self._sc: SchemaCoordinator | None = None
+        self._closing = asyncio.Event()
+        self._thread: Thread = Thread(target=self._start_loop, daemon=True)
+        self._loop: asyncio.AbstractEventLoop | None = None
+        self._schema_reader_stopper: SchemaReaderStoppper | None = None
+
+    def set_stoppper(self, schema_reader_stopper: SchemaReaderStoppper) -> None:
+        self._schema_reader_stopper = schema_reader_stopper
 
     @property
     def schema_coordinator(self) -> SchemaCoordinator | None:
@@ -41,7 +58,18 @@ class MasterCoordinator:
     def config(self) -> Config:
         return self._config
 
-    async def start(self) -> None:
+    def start(self) -> None:
+        self._thread.start()
+
+    def _start_loop(self) -> None:
+        # we should avoid the reassignment otherwise we leak resources
+        assert self._loop is None, "Loop already started"
+        self._loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self._loop)
+        self._loop.create_task(self._async_loop())
+        self._loop.run_forever()
+
+    async def _async_loop(self) -> None:
         self._kafka_client = self.init_kafka_client()
         # Wait until schema coordinator is ready.
         # This probably needs better synchronization than plain waits.
@@ -61,10 +89,22 @@ class MasterCoordinator:
             await asyncio.sleep(0.5)
 
         self._sc = self.init_schema_coordinator()
-        while True:
-            if self._sc.ready():
-                return
-            await asyncio.sleep(0.5)
+
+        # keeping the thread sleeping until it die.
+        # we need to keep the schema_coordinator running
+        # it contains the `heartbeat` and coordination logic.
+        await self._closing.wait()
+
+        LOG.info("Closing master_coordinator")
+        if self._sc:
+            await self._sc.close()
+        while self._loop is not None and not self._loop.is_closed():
+            self._loop.stop()
+            if not self._loop.is_running():
+                self._loop.close()
+            time.sleep(0.5)
+        if self._kafka_client:
+            await self._kafka_client.close()
 
     def init_kafka_client(self) -> AIOKafkaClient:
         ssl_context = create_ssl_context(
@@ -90,8 +130,10 @@ class MasterCoordinator:
 
     def init_schema_coordinator(self) -> SchemaCoordinator:
         assert self._kafka_client is not None
+        assert self._schema_reader_stopper is not None
         schema_coordinator = SchemaCoordinator(
             client=self._kafka_client,
+            schema_reader_stopper=self._schema_reader_stopper,
             election_strategy=self._config.get("master_election_strategy", "lowest"),
             group_id=self._config["group_id"],
             hostname=self._config["advertised_hostname"],
@@ -99,6 +141,7 @@ class MasterCoordinator:
             port=self._config["advertised_port"],
             scheme=self._config["advertised_protocol"],
             session_timeout_ms=self._config["session_timeout_ms"],
+            waiting_time_before_acting_as_master_ms=self._config["waiting_time_before_acting_as_master_ms"],
         )
         schema_coordinator.start()
         return schema_coordinator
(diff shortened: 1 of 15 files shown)


eliax1996:
immagine

jclarysse:
Please also add the new config `waiting_time_before_acting_as_master_ms` to https://github.com/Aiven-Open/karapace/blob/main/README.rst#configuration-keys

karapace/coordinator/master_coordinator.py
```diff
@@ -42,6 +59,17 @@ def config(self) -> Config:
         return self._config
 
     async def start(self) -> None:
+        self._thread.start()
+
+    def _start_loop(self) -> None:
+        # we should avoid the reassignment otherwise we leak resources
+        assert self._loop is None, "Loop already started"
+        self._loop = asyncio.new_event_loop()
```
nosahama, line 67: The call below to set the event loop might have some sideeffects beyond this object, but I see the idea is to set this new loop within the spawned thread.
Do you think it is possible to use the `loop` already on the `app` object? 
If we can, we probably can use [loop.run_in_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) to run the secluded code. 
We also could try to get the current running loop - `asyncio.get_running_loop()` and use the loop to run the threaded code, but it's also possible that at that time the application server hasn't started it's own loop.

nosahama: Not necessary, but I would prefer to have a single loop where it's lifecycle is already handled by the application server `aiohttp`, but I know you want to run it all in it's own thread, which leaves it properly isolated, so it's fine and you close the loop also 👍

karapace/coordinator/schema_coordinator.py
```diff
@@ -146,7 +147,17 @@ def __init__(
         self.scheme: Final = scheme
         self.master_eligibility: Final = master_eligibility
         self.master_url: str | None = None
-        self.are_we_master = False
+        self._schema_reader_stopper = schema_reader_stopper
+        self._are_we_master: bool | None = False
+        # a value that its strictly higher than any clock, so we are sure
+        # we are never going to consider this the leader without explictly passing
+        # from False to True for the `_are_we_master` variable.
+        self._initial_election_sec: float | None = float("infinity")
```
nosahama, line 155: The type cannot be `None` if there's a default value set.

karapace/schema_reader.py
```diff
@@ -315,6 +319,14 @@ def _is_ready(self) -> bool:
     def highest_offset(self) -> int:
         return max(self._highest_offset, self._offset_watcher.greatest_offset())
 
+    def ready(self) -> bool:
```
nosahama, line 322: I might be wrong, but as this is not a `property` object, I guess it is the counterpart of `set_not_ready()`, if so, maybe we should also set `self._ready = True` before returning it.

src/karapace/coordinator/schema_coordinator.py
```diff
@@ -434,13 +542,20 @@ def coordinator_dead(self) -> None:
...
         """Coordinator did not recognize either generation or member_id. Will
         need to re-join the group.
         """
+        LOG.info("Resetting generation status")
+        # this is called immediately after the election, we shouldn't reset this
+        # until a new node its elected aka the other path where a new node its elected
+        # otherwise this its called at each round and we keep not counting the 5 seconds
+        # required before the election.
+        # self._are_we_master = False
         self.generation = OffsetCommitRequest.DEFAULT_GENERATION_ID
```
eliax1996, line 559: this was called because we were exiting the election loop in the `_async_loop` of the `master_coordinator.py`. We must keep running the thread/algorithm otherwise we are always electing a new node since that causes a rebalance and the rebalance causes a new election (the rebalance happen because we sent the `reset_generation` as a side effect of closing the heartbeat task)

src/karapace/coordinator/master_coordinator.py
```diff
@@ -61,10 +89,25 @@ async def start(self) -> None:
...
-        while True:
-            if self._sc.ready():
-                return
+        while self._running:
+            # why do we need to close?
+            # we just need to keep running even when the schema registry its ready
+            # otherwise we cause a rebalance and a new election. This should run until
+            # karapace is restarted
+            # if self._sc.ready():
+            #    break
```
eliax1996, line 98: this question its mainly for @jjaakola-aiven. I inherited the initial implementation from him. I think we shouldn't exit but I wait for him to reply here

src/karapace/schema_registry_apis.py
```diff
@@ -206,6 +207,13 @@ def _add_schema_registry_routes(self) -> None:
             schema_request=True,
             auth=self._auth,
         )
+        self.route(
+            "/master_available",
+            callback=self.master_available,
+            # post because they cannot be cached, need to be sure always gathering the real value of that property
+            method="POST",
```
keejon, line 214: Not a big fan of this because it violates semantics of POST. Maybe set a `Cache-Control: no-store, no-cache, must-revalidate` header instead?

eliax1996: Agreed, I'll change it

src/karapace/schema_reader.py
```diff
@@ -433,9 +443,10 @@ def consume_messages(self, msgs: list[Message], watch_offsets: bool) -> None:
             # Default keymode is CANONICAL and preferred unless any data consumed
             # has key in non-canonical format. If keymode is set to DEPRECATED_KARAPACE
             # the subsequent keys are omitted from detection.
-            if not self.ready and self.key_formatter.get_keymode() == KeyMode.CANONICAL:
-                if msg_keymode == KeyMode.DEPRECATED_KARAPACE:
-                    self.key_formatter.set_keymode(KeyMode.DEPRECATED_KARAPACE)
+            with self._ready_lock:
+                if not self._ready and self.key_formatter.get_keymode() == KeyMode.CANONICAL:
```
jjaakola-aiven, line 447: With the changes done the key mode detection could be run during the normal operation. Not anymore only at startup.

src/karapace/coordinator/schema_coordinator.py
```diff
@@ -182,6 +194,49 @@ def __init__(
...
+            # `time.monotonic()` because we don't want the time to go back or forward because of
+            #  e.g. ntp
+            if time.monotonic() > self._initial_election_sec + (self._waiting_time_before_acting_as_master_ms / 1000):
+                # set the value to `None` since it's expensive to call each time the monotonic clock.
+                LOG.info("Declaring myself as master since %s are passed!", self._waiting_time_before_acting_as_master_ms)
+                self._initial_election_sec = None
+                # this is the last point in time were we wait till to the end of the log queue for new
+                # incoming messages.
+                self._schema_reader_stopper.set_not_ready()
+                # flag that its set at startup, fix this
```
jjaakola-aiven, line 229: A fix comment here, not sure what needs to be fixed?

tests/integration/test_schema_coordinator.py
```diff
@@ -155,14 +170,41 @@ async def test_coordinator_workflow(
...
-    await wait_for_primary_state(secondary)
+    now = time.time()
+    while time.time() - now < waiting_time_before_acting_as_master_sec:
+        assert not secondary.are_we_master(), (
+            f"Cannot become master before {waiting_time_before_acting_as_master_sec} seconds "
+            f"for the late records that can arrive from the previous master"
+        )
+        await asyncio.sleep(0.5)
+
+    while not secondary.are_we_master():
```
jjaakola-aiven, line 206: This will loop until test timeout. Maybe this could abort earlier if secondary is not set in reasonable time?

src/karapace/coordinator/schema_coordinator.py
eliax1996: missed to delete, thanks. Fixing

src/karapace/schema_reader.py
```diff
@@ -433,9 +443,8 @@ def consume_messages(self, msgs: list[Message], watch_offsets: bool) -> None:
             # Default keymode is CANONICAL and preferred unless any data consumed
             # has key in non-canonical format. If keymode is set to DEPRECATED_KARAPACE
             # the subsequent keys are omitted from detection.
-            if not self.ready and self.key_formatter.get_keymode() == KeyMode.CANONICAL:
-                if msg_keymode == KeyMode.DEPRECATED_KARAPACE:
-                    self.key_formatter.set_keymode(KeyMode.DEPRECATED_KARAPACE)
+            if self.key_formatter.get_keymode() == KeyMode.CANONICAL and msg_keymode == KeyMode.DEPRECATED_KARAPACE:
+                self.key_formatter.set_keymode(KeyMode.DEPRECATED_KARAPACE)
```
eliax1996, line 447: I think its better to keep the behaviour in that way, otherwise based on the restarts the cluster can behave in different ways. Lets make an example:
1. Karapace starts, all the messages are `CANONICAL`
2. Karapace become ready
3. Someone produce a `DEPRECATED` record (its an illegal behaviour but its possible to have it in production). Now, since `ready` its `True` Karapace will produce always only `CANONICAL` records
4. Karapace restarts, repeat the read
5. Now it produces only `DEPRECATED` records


We have proved that this "state" its not based on the input but on the sequence of events and its not stable, a restart can change the behaviour of the node causing an extremely hard issue to detect a runtime.

The proper fix should be to keep the info about witch key mode to use in the message itself, so karapace will always select the right formatter once a record its updated/changed/deleted

PR #939, opened 2024-08-23
metrics, visualization: StatsD Exporter, Prometheus Rules & Grafana Dashboard

nosahama:
This PR adds a `statsd-exporter` to the service, which will translate existing statsd metrics in influxdb style to prometheus metrics. We also add a simple grafana dashboard to Karapace, thus we can visualize the various metrics. 

We also now take the statsd host and port values from the config/env, rather than being hard coded within the client.

Sample prometheus `rules` file was added and a calculated metric is used within the dashboard. Spin up the container, send some requests and see the metrics on grafana and see the firing alerts on the prometheus alert manager. 

Screenshot 2024-08-23 at 10 53 48

Screenshot 2024-08-23 at 10 54 06

Alerts firing from the prometheus `rules` file, added with a low threshold so we see it fire:
Screenshot 2024-08-23 at 10 54 44

This is not final, the recorded metrics could use better promql techniques, this is just a base for any future work and at least we visualize what already exists in the service, i.e. the Schema Reader panel uses the exported statsd metrics and the HTTP requests use the added prometheus metrics, thus we have 2 metric handlers within the service and should migrate to one in the long run. 

A good question would be how and where will karapace send it's metrics while running within the Aiven environment.

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -78,6 +78,8 @@ services:
       KARAPACE_TOPIC_NAME: _schemas
       KARAPACE_LOG_LEVEL: WARNING
       KARAPACE_COMPATIBILITY: FULL
+      KARAPACE_STATSD_HOST: statsd-exporter
+      KARAPACE_STATSD_PORT: 8125
 
   karapace-rest:
     image: ghcr.io/aiven-open/karapace:develop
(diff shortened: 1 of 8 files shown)


eliax1996 (approved):
Awesome, LGTM

eliax1996:
Waiting a little bit if someone else want to take a look, then gonna merge it

PR #940, opened 2024-08-26
rapu: fire shutdown metric on app shutdown

nosahama:
- we add the `karapace_shutdown_count` metric, to the base app and fire the metric on app shutdown
- the `close_by_app()` is added to the app shutdown hooks and thus is called on app shutdown

### Shutdown logs
Screenshot 2024-08-26 at 16 18 22

### Active shutdown alert after 2 manual shutdowns
Screenshot 2024-08-26 at 16 20 37

### Firing shutdown alert after 2 manual shutdowns
Screenshot 2024-08-26 at 16 27 31

# Why this way
This way we have a metric to count the app shutdown ands this is handled on the app server level, rather than in application code.

### DIFF
--- a/karapace/rapu.py
+++ b/karapace/rapu.py
@@ -164,6 +164,7 @@ class RestApp:
         self.app_name = app_name
         self.config = config
         self.app_request_metric = f"{app_name}_request"
+        self.app_shutdown_count_metric = f"{app_name}_shutdown_count"
         self.app = self._create_aiohttp_application(config=config)
         self.log = logging.getLogger(self.app_name)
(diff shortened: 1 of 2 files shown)


eliax1996 (approved):
LGTM, nice great job!

nosahama:
> LGTM, nice great job!

Thanks man, I was thinking in another iteration, we can add labels to the shutdown metric, i.e. the reason for the shutdown, maybe set the error on the `app` object and reference the error when we fire the shutdown metric or leave it empty which will denote a normal expected shutdown, we can improve 👍

PR #946, opened 2024-09-12
feat: add systemd journal logging configuration

jjaakola-aiven:

Add configurability for logging handlers. The supported handlers are StreamHandler and systemd.journal.JournalHandler.
The log level is preserved for journal as log is written in structured format for journald.

Requires #947 first.

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/karapace_all.py
+++ b/src/karapace/karapace_all.py
@@ -2,10 +2,12 @@
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
+from __future__ import annotations
+
 from aiohttp.web_log import AccessLogger
 from contextlib import closing
 from karapace import version as karapace_version
-from karapace.config import read_config
+from karapace.config import Config, read_config
 from karapace.instrumentation.prometheus import PrometheusInstrumentation
 from karapace.kafka_rest_apis import KafkaRest
 from karapace.rapu import RestApp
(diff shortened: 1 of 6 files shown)


src/karapace/karapace_all.py
```diff
@@ -21,6 +23,37 @@ class KarapaceAll(KafkaRest, KarapaceSchemaRegistryController):
...
+    log_handler = config.get("log_handler", None)
+    if "systemd" == log_handler:
+        from systemd import journal
+
+        root_handler = journal.JournalHandler(SYSLOG_IDENTIFIER="karapace")
+    if "stdout" == log_handler or log_handler is None:
+        root_handler = logging.StreamHandler(stream=sys.stdout)
+    else:
+        logging.basicConfig(level=logging.INFO, format=log_format)
+        logging.getLogger().setLevel(log_level)
```
eliax1996, line 40: maybe add a warn log that the `log_handler` isn't recognized?

src/karapace/karapace_all.py
```diff
@@ -21,6 +23,37 @@ class KarapaceAll(KafkaRest, KarapaceSchemaRegistryController):
...
+    log_level = config.get("log_level", "DEBUG")
+    log_format = config.get("log_format", "%(name)-20s\t%(threadName)s\t%(levelname)-8s\t%(message)s")
+
+    root_handler: logging.Handler | None = None
+    log_handler = config.get("log_handler", None)
+    if "systemd" == log_handler:
+        from systemd import journal
+
+        root_handler = journal.JournalHandler(SYSLOG_IDENTIFIER="karapace")
+    if "stdout" == log_handler or log_handler is None:
```
eliax1996, line 36: nit: `elif` to highlight that this is an alternative option other than the first entry?

eliax1996 (approved):
Except the comments that are optional/non blocking LGTM

src/karapace/karapace_all.py
jjaakola-aiven: Done.

src/karapace/karapace_all.py
jjaakola-aiven: Done.

eliax1996 (approved):

PR #947, opened 2024-09-13
Migrate to pyproject toml

jjaakola-aiven:

First commit migrates to use `pyproject.toml`. Build backend is still the `setuptools`.
Karapace install: `pip install .`
Editable install for development: `pip install -e .[dev,typing]
Building the wheel: `python -m build`
Pinning of dependencies to: `make pin-requirements`. Pinning runs the `pip-compile` in a Python 3.8 container.
Version is now created by `setuptools-scm` from git state and is PEP compliant.

The second commit migrates to source layout.
The third commit cleans the Github links from `aiven` organization to `Aiven-Open`.
The fourth commit removes the `pyenv` requirement. Proposal is to implement and use `tox` later on for multiversion testing locally.

References: #xxxxx

# Why this way


.dockerignore
```diff
@@ -3,11 +3,11 @@
 *
 
 # Include source directories and files required for building.
+!.git
```
aiven-anton, line 6: issue: This is likely not a good idea, as it will include the full git history in built docker images. If this is to enable VCS versioning, I would instead highly encourage making the build job extract the version from VCS and pass it as a build parameter to docker.

GNUmakefile
```diff
@@ -84,12 +87,14 @@ cleanest: cleaner
 .PHONY: requirements
 requirements: export CUSTOM_COMPILE_COMMAND='make requirements'
 requirements:
-	pip install --upgrade pip setuptools pip-tools
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements.in -o requirements.txt
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-dev.in -o requirements-dev.txt
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-typing.in -o requirements-typing.txt
+	$(PIP) install --upgrade pip setuptools pip-tools
+	$(PIP) install . .[dev] .[typing]
```
aiven-anton, line 91:
```suggestion
	$(PIP) install .[dev,typing]
```

.dockerignore
jjaakola-aiven: The container build is multistage build. The actual Karapace container has only the copied virtual environment: https://github.com/Aiven-Open/karapace/blob/main/container/Dockerfile#L42

aiven-anton: Ah right, and `KARAPACE_VERSION` is already passed as a build parameter actually, perhaps that's not necessary any longer then.

GNUmakefile
jjaakola-aiven: Done.

CONTRIBUTING.md
```diff
@@ -12,15 +12,15 @@ Contributions are very welcome on Karapace. When contributing please keep this i
...
+  documented in the `project:requires-python`.
 - Create [a virtual environment](https://docs.python.org/3/tutorial/venv.html) and install the dev dependencies in it:
 
 ```python
 python -m venv <path_to_venv>
 source <path_to_venv>/bin/activate
-pip install -r ./requirements/requirements-dev.txt
-pip install -e .
+pip install .
+pip install .[dev,typing]
```
aiven-anton, line 23:
```suggestion
pip install -e .[dev,typing]
```

issue: Why is `-e` removed here? If it's removed, you would have to reinstall for any local changes to take effect for testing etc. I see there's no `src/` directory, which might mean that tests can see the non-installed local package instead, but that should really also be remediated to make sure tests run on the actually installed package.

.github/workflows/lint.yml
```diff
@@ -45,6 +43,6 @@ jobs:
         python-version: '3.12'
     - name: Install libsnappy-dev
       run: sudo apt install libsnappy-dev
-    - run: pip install -r requirements/requirements.txt -r requirements/requirements-typing.txt
-    - run: make karapace/version.py
+    - name: Install requirements and typing requirements
+      run: pip install . .[typing]
```
aiven-anton, line 47:
```suggestion
      run: pip install .[typing]
```

aiven-anton (dismissed):

CONTRIBUTING.md
jjaakola-aiven: The move to `/src´ is good, I'll find some time to do it subsequent PR.

.dockerignore
aiven-anton: Mentioned in 1-1, just copying here for completeness too: including the `.git` directory, even in the build stage, will cause all layers of the image to have to be rebuilt if there are changes in that directory -> which means for instance doing a `git fetch` can cause a complete rebuild.

jjaakola-aiven: Removed the requirement, added use of `SETUPTOOLS_SCM_PRETEND_VERSION` to `Dockerfile`.

container/Dockerfile
```diff
@@ -19,8 +19,14 @@ RUN --mount=type=cache,target=/root/.cache/pip \
...
 COPY . /build/karapace-repo
+WORKDIR /build/karapace-repo
 RUN --mount=type=cache,target=/root/.cache/pip \
-    python3 -m pip install /build/karapace-repo
+    if [ -z "${KARAPACE_VERSION}" ]; then \
+        PRETEND_VERSION="$(python -c 'from karapace import version; print(version.__version__)')"; \
+    else \
+        PRETEND_VERSION=$KARAPACE_VERSION; \
+    fi; \
+    SETUPTOOLS_SCM_PRETEND_VERSION=$PRETEND_VERSION python3 -m pip install --no-deps --no-cache-dir .
```
aiven-anton, line 29:
```suggestion
    SETUPTOOLS_SCM_PRETEND_VERSION=$PRETEND_VERSION python3 -m pip install --no-deps .
```

issue: The command runs with `--mount=type=cache,target=/root/.cache/pip` which sets up docker volume that is persistent between builds, so caching should work "as expected", without adding any additional weight to the resulting image.

GNUmakefile
```diff
@@ -84,12 +87,14 @@ cleanest: cleaner
 .PHONY: requirements
 requirements: export CUSTOM_COMPILE_COMMAND='make requirements'
 requirements:
-	pip install --upgrade pip setuptools pip-tools
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements.in -o requirements.txt
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-dev.in -o requirements-dev.txt
-	cd requirements && pip-compile --upgrade --resolver=backtracking requirements-typing.in -o requirements-typing.txt
+	$(PIP) install --upgrade pip setuptools pip-tools
+	$(PIP) install . .[dev,typing]
```
aiven-anton, line 91:
```suggestion
	$(PIP) install .[dev,typing]
```

Claudenw:
The JSON serialization tests failed.  This seems, to me,  like an odd one to fail.  Does this change impact JSON serialization or is the JSON serialization test flaky?

jjaakola-aiven:
@Claudenw config file reading and parsing failed. The file (fixture) is created here: https://github.com/Aiven-Open/karapace/blob/main/tests/conftest.py#L163

nosahama:
If we now use `setuptools-scm`, we might need to add it to the [spec file](https://github.com/aiven/aiven-core/blob/main/deps/karapace/karapace.spec#L19) when we bump the karapace version on aiven-core, i.e. `BuildRequires:  python3dist(setuptools-scm)` and maybe `BuildRequires:  python3dist(setuptools-scm[toml])`, as these tools are no longer global on the RPMs build.

GNUmakefile
```diff
@@ -84,12 +87,14 @@ cleanest: cleaner
 .PHONY: requirements
 requirements: export CUSTOM_COMPILE_COMMAND='make requirements'
```
aiven-anton, line 88: issue: It no longer makes sense to set this environment variable here. If it is to be kept, it should be moved and updated to the `pin-requirements` target and propagated inside the container run there.

tests/integration/test_master_coordinator.py
```diff
@@ -210,7 +210,9 @@ async def test_schema_request_forwarding(registry_async_pair):
             path = "config"
         for compat in ["FULL", "BACKWARD", "FORWARD", "NONE"]:
             resp = requests.put(f"{slave_url}/{path}", json={"compatibility": compat})
-            assert resp.ok
+            if not resp.ok:
+                # Allow retry if response is not ok.
+                continue
```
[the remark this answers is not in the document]
aiven-anton: Outdated.

.github/workflows/container-smoke-test.yml
```diff
@@ -13,12 +13,27 @@ jobs:
...
+        run: |
+          source ./venv/bin/activate
+          KARAPACE_VERSION=$(python -c "from karapace import version; print(version.__version__)")
+          echo KARAPACE_VERSION=$KARAPACE_VERSION >> $GITHUB_ENV
+
+      - name: Build container
+        run: docker build --build-arg KARAPACE_VERSION=${{ env.KARAPACE_VERSION }} --file=container/Dockerfile .
+
+      - name: Run container
+        run: docker compose --file=container/compose.yml up --wait --detach
```
aiven-anton, line 36: issue: I think removing `--build` here is going to result in compose instead downloading `ghcr.io/aiven-open/karapace:develop`, rather than using a build with evaluated version of the code-base.

There are some controls for the behavior: https://docs.docker.com/reference/compose-file/build/#using-build-and-image. But I think we would also want to the above build command to rather invoke `docker compose build` than `docker build`.

GNUmakefile
```diff
@@ -40,31 +48,26 @@ venv/.deps: requirements/requirements-dev.txt requirements/requirements.txt | ve
...
 	$(PIP) check
 	touch '$(@)'
 
-
-karapace/version.py:
-	$(PYTHON) version.py
-
 .PHONY: version
-version: venv/.make | karapace/version.py
+version: venv/.make
```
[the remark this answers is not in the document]
jjaakola-aiven: Done.

GNUmakefile
jjaakola-aiven: Done.

requirements/requirements-dev.txt
```diff
@@ -2,44 +2,40 @@
 # This file is autogenerated by pip-compile with Python 3.8
 # by the following command:
 #
-#    'make requirements'
+#    pip-compile --extra=dev --output-file=/karapace/requirements/requirements-dev.txt /karapace/pyproject.toml
```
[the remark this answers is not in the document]
jjaakola-aiven: Done.

src/karapace/backup/backends/v3/schema_tool.py
```diff
@@ -58,6 +58,17 @@ def relative_path(path: pathlib.Path) -> pathlib.Path:
     return pathlib.Path(str_path[len(cwd) + 1 :]) if str_path.startswith(cwd) else path
 
 
+def target_has_source_layout(git_target: str) -> bool:
```
aiven-anton, line 61: thought (non-blocking): This is OK, but I think would have allowed this check to fail for this PR instead. The trade-off is that the PR is now not purely a clean move into `src/`.

aiven-anton (dismissed):
LGTM, nice work! 👍

CONTRIBUTING.md
```diff
@@ -12,15 +12,15 @@ Contributions are very welcome on Karapace. When contributing please keep this i
...
+  documented in the `project:requires-python`.
 - Create [a virtual environment](https://docs.python.org/3/tutorial/venv.html) and install the dev dependencies in it:
 
 ```python
 python -m venv <path_to_venv>
 source <path_to_venv>/bin/activate
-pip install -r ./requirements/requirements-dev.txt
-pip install -e .
+pip install .
+pip install -e .[dev,typing]
```
davide-armand, line 23: In my zsh terminal I had to add quotes because of the squared brackets:
```bash
pip install -e ".[dev,typing]"
```

README.rst
```diff
@@ -651,7 +651,7 @@ If you don't need or want to have the Karapace images around you can now proceed
 Installed from Sources
 ----------------------
 
-If you installed Karapace from the sources via ``python setup.py install``, it can be uninstalled with the following ``pip`` command::
+Karapace is installed ``pip install .``, it can be uninstalled with the following ``pip`` command::
```
davide-armand, line 654:
```suggestion
If Karapace is installed with ``pip install .``, then it can be uninstalled with the following ``pip`` command::
```

README.rst
```diff
@@ -695,11 +689,8 @@ targets that correctly clean the ``runtime`` directory without deleting it, but
 keep this in mind whenever you are not using ``make`` (e.g. running tests from
 your IDE).
 
-Note that the pre-commit checks are currently not working with the default
-Python version. This is because isort dropped Python 3.7 support. You have to
-use at least Python 3.8 for the pre-commit checks. Use ``pipx`` or ``brew`` or
-… to install pre-commit and use the global installation, there is also no
-dependency on it.
+Use ``pipx`` or ``brew`` or to install pre-commit and use the global installation,
```
davide-armand, line 692:
```suggestion
Use ``pipx`` or ``brew`` to install ``pre-commit`` and use the global installation,
```

keejon (approved):
LGTM 👍

README.rst
keejon: @jjaakola-aiven do you want to fix this before merging?

jjaakola-aiven: Done.

keejon (approved):

PR #949, opened 2024-09-17
chore: add retry to tests requiring forwarding

jjaakola-aiven:

References: #xxxxx

# Why this way

### DIFF
--- a/tests/integration/test_schema_registry_auth.py
+++ b/tests/integration/test_schema_registry_auth.py
@@ -4,9 +4,9 @@ karapace - schema registry authentication and authorization tests
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
-from karapace.client import Client
 from karapace.kafka.admin import KafkaAdminClient
 from karapace.schema_models import SchemaType, ValidatedTypedSchema
+from tests.integration.utils.rest_client import RetryRestClient
 from tests.utils import (
     new_random_name,
     new_topic,
@@ -20,7 +20,6 @@ from urllib.parse import quote
 
 import aiohttp
 import asyncio
-import requests
 
 NEW_TOPIC_TIMEOUT = 10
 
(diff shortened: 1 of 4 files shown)


tests/integration/utils/rest_client.py
```diff
@@ -0,0 +1,78 @@
...
+
+class UnexpectedResponseStatus(Exception):
+    pass
+
+
+class RetryRestClient:
+    def __init__(self, client: Client):
+        self._karapace_client = client
+
+    @retry(stop=stop_after_attempt(5))
```
keejon, line 24: Should we maybe wait a bit between retries?

tests/integration/utils/rest_client.py
```diff
@@ -0,0 +1,78 @@
...
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+from __future__ import annotations
+
+from aiohttp import BasicAuth
+from collections.abc import Mapping
+from karapace.client import Client, Headers, Path, Result
+from karapace.typing import JsonData
+from tenacity import retry, stop_after_attempt
```
nosahama, line 13: Noice 💯

tests/integration/test_schema_registry_auth.py
```diff
@@ -29,139 +29,157 @@
...
     )
     assert res.status_code == 200
     sc_id = res.json()["id"]
     assert sc_id >= 0
 
-    res = await registry_async_client_auth.get(f"subjects/{quote(subject)}/versions/latest")
+    res = await registry_async_retry_client_auth.get(
+        f"subjects/{quote(subject)}/versions/latest", expected_response_code=401
+    )
     assert res.status_code == 401
```
nosahama, line 50: **[non-blocking]**
I know it'd be a harder refactor, but we also somehow already assert the response code:
```
if response.status_code != expected_response_code:
    raise UnexpectedResponseStatus(f"Unexpected status code: {response!r}")
```
we can omit the assertion `assert res.status_code == 401` in all places where we use the `retry_client` or even move the assertion into the client stub.

tests/integration/utils/rest_client.py
```diff
@@ -0,0 +1,81 @@
...
+
+from aiohttp import BasicAuth
+from collections.abc import Mapping
+from karapace.client import Client, Headers, Path, Result
+from karapace.typing import JsonData
+from tenacity import retry, stop_after_attempt, wait_fixed
+from typing import Final
+
+
+RETRY_WAIT: Final = 0.5
```
nosahama, line 17: nanoseconds 😄

nosahama (dismissed):

tests/integration/test_schema_registry_auth.py
```diff
@@ -198,8 +218,7 @@ async def test_sr_auth_forwarding(registry_async_auth_pair: List[str]) -> None:
         while True:
             assert counter < max_tries, "Compat update not propagated"
             resp = requests.get(f"{primary_url}/config", auth=auth)
-            if not resp.ok:
-                continue
+            assert resp.ok
```
keejon, line 221: Is this change intentional? It will fail now instead of retrying.

nosahama: As we have a retry alreadt, maybe no need to continue, but then we have this check above `assert counter < max_tries, "Compat update not propagated"`, not sure if it's used for the old retry style.

jjaakola-aiven: No, good catch.

tests/integration/utils/rest_client.py
jjaakola-aiven: Done.

tests/integration/utils/rest_client.py
jjaakola-aiven: Renamed.

tests/integration/test_schema_registry_auth.py
jjaakola-aiven: Done.

tests/integration/test_schema_registry_auth.py
jjaakola-aiven: Requires bit more work. I'd like to have this in first as these do reduce the flakiness.

keejon (dismissed):
LGTM 👍

tests/integration/test_schema_registry_auth.py
```diff
@@ -170,34 +188,36 @@ async def test_sr_ids(registry_async_client_auth: Client) -> None:
...
+async def test_sr_auth_forwarding(
+    registry_async_auth_pair: List[str], registry_async_retry_client_auth: RetryRestClient
+) -> None:
     # Test primary/replica forwarding with global config setting
     primary_url, replica_url = registry_async_auth_pair
     max_tries, counter = 5, 0
     wait_time = 0.5
     for compat in ["FULL", "BACKWARD", "FORWARD", "NONE"]:
-        resp = requests.put(f"{replica_url}/config", json={"compatibility": compat}, auth=auth)
+        resp = await registry_async_retry_client_auth.put(f"{replica_url}/config", json={"compatibility": compat}, auth=admin)
```
nosahama, line 216: This was failing and we had to split the used `auth`, `aiohttp` requires the `BasicAuth` and `requests` requires something different, but we were first passing the requests auth to the aiohttp client and it was failing.

AnatolyPopov (commented):
LGTM

PR #950, opened 2024-09-17
config: Dropped `kafka` prefix from shutdown flags

nosahama:
We remove the `kafka_` prefix from the config values: `kafka_schema_reader_strict_mode` and `kafka_retriable_errors_silenced`.


aiven-anton:
I think it's worth noting that this is a backwards incompatible change.

nosahama:
This will require a new release on aiven-core, i'd make the fix directly there.

eliax1996 (commented):
Maybe lets accept also the previous configuration as valid, otherwise we have a breaking change

PR #953, opened 2024-09-18
fix: align transitive compatibility checks

davide-armand:

Make so that the transitive compatibility checks that are done in the
schema creation endpoint are also done in the schema validation
endpoint.

In the creation endpoint (`/subjects/<subject-key>/versions`), if the
compatibility mode is transient then the new schema is checked against
all schemas.

In the validation endpoint (`/compatibility/subjects/<subject-key>/versions/<schema-version>`):
 - Before this fix, only the latest schema is checked against (even in
   case of transitive mode)
 - After this fix, in case of transitive mode then all schema are
   checked against.
   Note that in this case the version provided in the POST request
   (`<schema-version>`) is ignored.

Also includes some preparation refactorings and some additional tests in separate commits. 

# Why this way

The refactorings are needed to avoid code duplication.

An alternative solution would be to allow querying the validation endpoint with no version (`/compatibility/subjects/<subject-key>`), and apply the fix only in that case.


website/source/quickstart.rst
```diff
@@ -60,6 +60,10 @@ Test the compatibility of a schema with the latest schema under subject "test-ke
     $KARAPACE_REGISTRY_URI/compatibility/subjects/test-key/versions/latest
   {"is_compatible":true}
 
+NOTE: if the subject's compatibility mode is transitive (BACKWARD_TRANSITIVE, FORWARD_TRANSITIVE or FULL_TRANSITIVE) then the
```
davide-armand, line 63: Is there any other documentation that needs to be updated?

karapace/schema_registry.py
```diff
@@ -351,32 +349,15 @@ async def write_new_schema_local(
...
-                    check_against = [schema_versions[-1]]
-
-                for old_version in check_against:
-                    parsed_old_schema = self._resolve_and_parse(all_schema_versions[old_version].schema)
-                    result = check_compatibility(
-                        old_schema=parsed_old_schema,
-                        new_schema=new_schema,
-                        compatibility_mode=compatibility_mode,
+                if is_incompatible(result):
+                    message = set(result.messages).pop() if result.messages else ""
```
eliax1996, line 355: mmm isn't better to return all the errors? Maybe a string interpolation with all the errors its better than returning just the first

karapace/schema_registry.py
```diff
@@ -465,3 +446,48 @@ def send_delete_subject_message(self, subject: Subject, version: Version) -> Non
...
+            old_parsed_schema = self.resolve_and_parse(all_schema_versions[old_version].schema)
+
+            result = SchemaCompatibility.check_compatibility(
+                old_schema=old_parsed_schema,
+                new_schema=new_schema,
+                compatibility_mode=compatibility_mode,
+            )
+
+            if is_incompatible(result):
+                break
```
eliax1996, line 482: maybe:
```
            if is_incompatible(result):
                return result
```
more direct with the intent and less error prone in refactoring

karapace/schema_registry_apis.py
```diff
@@ -1362,3 +1317,57 @@ def no_master_error(self, content_type: str) -> None:
             content_type=content_type,
             status=HTTPStatus.INTERNAL_SERVER_ERROR,
         )
+
+    def get_new_schema(self, body: JsonObject, content_type: str) -> ValidatedTypedSchema:
```
eliax1996, line 1322: maybe private? It make sense to expose this outside of this file? I think its strictly bound to the webpage logic

tests/unit/test_schema_registry.py
```diff
@@ -0,0 +1,257 @@
...
+from karapace.typing import Subject, Version
+from unittest import mock
+from unittest.mock import MagicMock, Mock
+
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+class MockedSchemaRegistry:
```
eliax1996, line 22: Can you extract an interface and try to use DI, not blocking on this since its matter of taste, but I find hard to understand what this object its trying to do and what is the expected behaviour

tests/unit/test_schema_registry.py
```diff
@@ -0,0 +1,257 @@
...
+    result.compatibility = SchemaCompatibilityType.compatible
+    SchemaCompatibility.check_compatibility = Mock(return_value=result)
+
+    # When
+    schema_registry.check_schema_compatibility(new_schema, subject)
+
+    # Then
+    assert result.compatibility is SchemaCompatibilityType.compatible
+
+    # All 3 schemas are checked against
```
eliax1996, line 93: got it, but why not creating a test where you know that the schemas are compatible 1 by 1 but not fully transitive? Isn't better to test the outcome instead of the actions? Again, matter of taste. Maybe better to ask a second opinion on this before doing work :D Maybe its me that isn't used to test like this and isn't anything bad

tests/unit/test_schema_registry.py
```diff
@@ -0,0 +1,257 @@
...
+    # Only one schema is checked against (first fail stops all checks)
+    SchemaCompatibility.check_compatibility.assert_has_calls(
+        [
+            mock.call(
+                old_schema=schema_registry.parsed_schema1,
+                new_schema=new_schema,
+                compatibility_mode=schema_registry.compatibility_mode,
+            ),
+        ]
+    )
```
eliax1996, line 141: kinda the same comment, I would like to see two test. 1 where the schemas are compatible and the api reply green, one where the schemas are 2 by 2 compatible but not fully transitive compatible and the api reply red

eliax1996 (commented):
I think the code its nice, the feature its well tested.
I'm still not fully convinced by the use of mocks, I think they can couple the way of a certain result its computed.
I would much more like a test where you give the input and the real implementation does the logic and you check the outcome, its a type of testing that makes the refactor much less painful since you don't need to change the assertions not being coupled by the names of the functions and the order in which they are called

karapace/schema_registry_apis.py
davide-armand: Made it private, although it is used outside the class in a test though (`tests/unit/test_schema_registry_api.py`), does that count? :)
My Pyright complains about it, but maybe I should just remove that check in my setup

karapace/schema_registry.py
jjaakola-aiven: Yes, it would be better to hand out all errors instead of first.

karapace/schema_registry_apis.py
```diff
@@ -1362,3 +1318,56 @@ def no_master_error(self, content_type: str) -> None:
...
+            self.r(
+                body={
+                    "error_code": SchemaErrorCodes.INVALID_SCHEMA.value,
+                    "message": f"Invalid {schema_type} schema",
+                },
+                content_type=content_type,
+                status=HTTPStatus.UNPROCESSABLE_ENTITY,
+            )
+
+    def _get_old_schema(self, subject: Subject, version: Version, content_type: str) -> ValidatedTypedSchema:
```
jjaakola-aiven, line 1343:
```suggestion
    def _get_old_schema(self, subject: Subject, version: Version, content_type: str) -> ParsedTypedSchema:
```
The old schemas are not fully validated, there can be such old schemas in the Karapace database that cannot pass validation anymore with newer and current Avro library version.

tests/unit/test_schema_registry.py
jjaakola-aiven, line 1: Tests here could be integration tests and might be clearer there.

karapace/schema_registry.py
davide-armand: @jjaakola-aiven Good point, I have fixed it in a separate commit since not directly in the scope of this ticket.
I made a simple comma separated list, not sure if good enough.

I have also fixed it for the `/compatibility` endpoint here (it's json in this case):
https://github.com/Aiven-Open/karapace/pull/953/files#diff-a79ae23c164ad9d4d7ba51504e6d20f25dbbdc2b59f4d91bc2f105e3da9a6e63R410

tests/unit/test_schema_registry.py
davide-armand: @jjaakola-aiven I have move all test to `tests/integration/test_schema_compatibility.py`.
No mocks now and hopefully clearer than before :)

tests/unit/test_schema_registry.py
davide-armand: Moved to `tests/integration/test_schema_compatibility.py`

tests/unit/test_schema_registry.py
davide-armand: Moved to `tests/integration/test_schema_compatibility.py`

tests/unit/test_schema_registry.py
davide-armand: Moved to `tests/integration/test_schema_compatibility.py`

davide-armand:
Rebased and fixed conflicts

website/source/quickstart.rst
eliax1996: I don't think so. The documentation should be revised completely IMO but for that PR I think its fine as it is. PS good addition

src/karapace/compatibility/schema_compatibility.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if old_schema.schema_type is SchemaType.AVRO:
+            assert isinstance(old_schema.schema, AvroSchema)
+            assert isinstance(new_schema.schema, AvroSchema)
+            if compatibility_mode in {CompatibilityModes.BACKWARD, CompatibilityModes.BACKWARD_TRANSITIVE}:
+                result = SchemaCompatibility.check_avro_compatibility(
+                    reader_schema=new_schema.schema,
+                    writer_schema=old_schema.schema,
+                )
+
```
eliax1996, line 55: nit: why those empty lines between the `if` and `elif`?

src/karapace/schema_registry.py
```diff
@@ -465,3 +448,48 @@ def send_delete_subject_message(self, subject: Subject, version: Version) -> Non
...
+        new_schema: ValidatedTypedSchema,
+        subject: Subject,
+    ) -> SchemaCompatibilityResult:
+        result = SchemaCompatibilityResult(SchemaCompatibilityType.compatible)
+
+        compatibility_mode = self.get_compatibility_mode(subject=subject)
+        all_schema_versions: dict[Version, SchemaVersion] = self.database.find_subject_schemas(
+            subject=subject, include_deleted=True
+        )
+        live_versions: list[Version] = self.get_live_versions_sorted(all_schema_versions)
```
eliax1996, line 463: nit: do we need the explicit type declaration here? 🤔
It should inferred by the return type of the function

src/karapace/schema_registry_apis.py
```diff
@@ -446,13 +396,18 @@ async def compatibility_check(
...
+            # Ignore the schema version provided in the rest api call (`version`)
+            # Instead check against all previous versions (including `version` if existing)
+            result = self.schema_registry.check_schema_compatibility(new_schema, subject)
+        else:
+            # Check against the schema version provided in the rest api call (`version`)
+            result = SchemaCompatibility.check_compatibility(old_schema, new_schema, compatibility_mode)
+
         if is_incompatible(result):
-            self.r({"is_compatible": False}, content_type)
+            self.r({"is_compatible": False, "incompatibilities": f"{', '.join(result.messages)[:300]}"}, content_type)
```
eliax1996, line 410: again super nit, I would declare the error message as a separate variable like:
```python3
maybe_truncated_error = ', '.join(result.messages)[:300]
self.r({"is_compatible": False, "incompatibilities": maybe_truncated_error}, content_type)
```

tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -183,7 +183,7 @@ async def test_protobuf_schema_compatibility_dependencies(registry_async_client:
         json={"schemaType": "PROTOBUF", "schema": evolved_schema, "references": evolved_references},
     )
     assert res.status_code == 200
-    assert res.json() == {"is_compatible": False}
+    assert not res.json().get("is_compatible")
```
eliax1996, line 186: issue: not its true also for `None` that its the default value of the `.get` when its empty

tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -271,7 +271,7 @@ async def test_protobuf_schema_compatibility_dependencies1(registry_async_client
...
```
eliax1996, line 274: same, I would change it tto: `assert res.json().get("is_compatible") is False`

tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -339,7 +339,7 @@ async def test_protobuf_schema_compatibility_dependencies1g(registry_async_clien
         json={"schemaType": "PROTOBUF", "schema": evolved_schema},
     )
     assert res.status_code == 200
-    assert res.json() == {"is_compatible": False}
+    assert not res.json().get("is_compatible")
```
eliax1996, line 342: same

tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -407,7 +407,7 @@ async def test_protobuf_schema_compatibility_dependencies1g_otherway(registry_as
         json={"schemaType": "PROTOBUF", "schema": evolved_schema, "references": container_references},
     )
     assert res.status_code == 200
-    assert res.json() == {"is_compatible": False}
+    assert not res.json().get("is_compatible")
```
eliax1996, line 410: and here

tests/integration/test_dependencies_compatibility_protobuf.py
```diff
@@ -491,7 +491,7 @@ async def test_protobuf_schema_compatibility_dependencies2(registry_async_client
         json={"schemaType": "PROTOBUF", "schema": evolved_schema, "references": evolved_references},
     )
     assert res.status_code == 200
-    assert res.json() == {"is_compatible": False}
+    assert not res.json().get("is_compatible")
```
eliax1996, line 494: here

tests/integration/test_schema_compatibility.py
```diff
@@ -0,0 +1,219 @@
...
+import json
+import logging
+import pytest
+
+LOG = logging.getLogger(__name__)
+
+schema_int = {"type": "record", "name": "schema_name", "fields": [{"type": "int", "name": "field_name"}]}
+schema_long = {"type": "record", "name": "schema_name", "fields": [{"type": "long", "name": "field_name"}]}
+schema_string = {"type": "record", "name": "schema_name", "fields": [{"type": "string", "name": "field_name"}]}
+schema_double = {"type": "record", "name": "schema_name", "fields": [{"type": "double", "name": "field_name"}]}
```
eliax1996, line 21: nit: add `Final` as type

tests/integration/test_schema_compatibility.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+@dataclass
+class SchemaCompatibilityTestCase(BaseTestCase):
+    new_schema: str
+    compatibility_mode: str
+    register_baseline_schemas: Callable[[Client, Subject], Coroutine[Any, Any, None]]
```
eliax1996, line 28: optional: wdyt of creating a type alias for this? Not required, just asking if make sense in your opinion

tests/integration/test_schema_compatibility.py
```diff
@@ -0,0 +1,219 @@
...
+    expected_is_compatible: Union[bool, None]
```
eliax1996, line 29: maybe `bool | None`?
If you add at the beginning: `from __future__ import annotations` so you can use also the type annotations of the most recent python in the old versions

tests/integration/test_schema_compatibility.py
```diff
@@ -0,0 +1,219 @@
...
+    expected_status_code: int
+    expected_incompatibilities: Union[str, None]
```
eliax1996, line 31: same here, but should be fixed by the `pre-commit` as soon as you import the `annotations`

tests/integration/test_schema_compatibility.py
```diff
@@ -0,0 +1,219 @@
...
+            register_baseline_schemas=_register_no_baseline_schemas,
+            new_schema=json.dumps(schema_int),
+            expected_is_compatible=None,
+            expected_status_code=404,
+            expected_incompatibilities=None,
+        ),
+        # Case 7
+        # Same as previous case, but in non-transitive mode
+        #   --> No incompatibilities are found
+        #   --> Status code is 404 because `latest` version to check against does not exists
```
eliax1996, line 194: awesome! Thanks, love this style 🥳 🚀

eliax1996 (commented):
Almost there, except for a couple of changes great great job!
I like much more those tests, I can read them and understand them really well, thanks for the hard work 🚀

src/karapace/compatibility/schema_compatibility.py
davide-armand: Not sure, I think it was pre-existing code that I moved around.
I have now removed the empty lines so it's more compact :)

src/karapace/schema_registry_apis.py
davide-armand: Agree, looks more readable now!

tests/integration/test_dependencies_compatibility_protobuf.py
davide-armand: Good catch, fixed

tests/integration/test_schema_compatibility.py
davide-armand: Agree, more readable

tests/integration/test_schema_compatibility.py
davide-armand: Didn't know that :D
Fixed

karapace/schema_registry_apis.py
davide-armand: Good catch, fixed

davide-armand:
@eliax1996 I should I have address all comments, thanks for your detailed review!

tests/integration/test_schema.py
```diff
@@ -332,7 +332,8 @@ async def test_compatibility_endpoint(registry_async_client: Client, trail: str)
         json={"schema": json.dumps(schema)},
     )
     assert res.status_code == 200
-    assert res.json() == {"is_compatible": False}
+    assert not res.json().get("is_compatible")
```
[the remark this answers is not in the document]
eliax1996: mmm maybe you forgot this

eliax1996 (approved):

PR #955, opened 2024-09-18
test: fix backup test flakiness

jjaakola-aiven:

References: #xxxxx

# Why this way


tests/integration/backup/test_v3_backup.py
```diff
@@ -89,6 +89,23 @@ def _raise(exception: Exception) -> NoReturn:
...
+def _delete_topic(admin_client: KafkaAdminClient, topic_name: str) -> None:
+    try:
+        admin_client.delete_topic(topic_name)
+    except UnknownTopicOrPartitionError:
+        logger.info("No previously existing topic.")
+    else:
+        logger.info("Deleted topic from previous run.")
+
+    start_time = time.monotonic()
+    while True:
```
juha-aiven, line 101: Use `Timer`.

eliax1996 (approved):
even if we can improve we need to get rid asap of the flakiness. Merging

PR #956, opened 2024-09-19
chore: include hidden files when uploading coverage

jjaakola-aiven:
See: https://github.com/py-cov-action/python-coverage-comment-action/issues/470


AnatolyPopov (approved):
LGTM

PR #957, opened 2024-09-19
feat: improve health check to fail if schema_reader raises exceptions

keejon:

This PR improves the health check endpoint by returning 503 if the schema_reader throws  >= 3 unexpected consecutive exceptions in 10 seconds.
In addition it will check if topic is available on the kafka cluster.  This is needed because consumer will only return error once in this case.

# Why this way

In its current state health check would always report healthy service even if connection to kafka was broken.

### DIFF
--- a/src/karapace/schema_reader.py
+++ b/src/karapace/schema_reader.py
@@ -22,7 +22,7 @@ from aiokafka.errors import (
 )
 from avro.schema import Schema as AvroSchema
 from collections.abc import Mapping, Sequence
-from confluent_kafka import Message, TopicPartition
+from confluent_kafka import Message, TopicCollection, TopicPartition
 from contextlib import closing, ExitStack
 from enum import Enum
 from jsonschema.validators import Draft7Validator
@@ -48,6 +48,7 @@ from karapace.utils import json_decode, JSONDecodeError, shutdown
 from threading import Event, Thread
 from typing import Final
 
+import asyncio
 import json
 import logging
 import time
@@ -61,6 +62,11 @@ OFFSET_EMPTY: Final = -1
 
 KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS: Final = 2.0
 SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS: Final = 5.0
+
+# If handle_messages throws at least UNHEALTHY_CONSECUTIVE_ERRORS
+# for UNHEALTHY_TIMEOUT_SECONDS the SchemaReader will be reported unhealthy
+UNHEALTHY_TIMEOUT_SECONDS: Final = 10.0
+UNHEALTHY_CONSECUTIVE_ERRORS: Final = 3
 
 # For good startup performance the consumption of multiple
 # records for each consume round is essential.
@@ -176,6 +182,9 @@ class KafkaSchemaReader(Thread):
         self.start_time = time.monotonic()
         self.startup_previous_processed_offset = 0
 
+        self.consecutive_unexpected_errors: int = 0
+        self.consecutive_unexpected_errors_start: float = 0
+
     def close(self) -> None:
         LOG.info("Closing schema_reader")
         self._stop_schema_reader.set()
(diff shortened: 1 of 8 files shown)


karapace/schema_reader.py
```diff
@@ -247,11 +248,13 @@ def run(self) -> None:
                     self.offset = self._get_beginning_offset()
                 try:
                     self.handle_messages()
+                    self.consecutive_unexpected_errors = 0
                 except ShutdownException:
                     self._stop_schema_reader.set()
                     shutdown()
                 except Exception as e:  # pylint: disable=broad-except
                     self.stats.unexpected_exception(ex=e, where="schema_reader_loop")
+                    self.consecutive_unexpected_errors += 1
```
nosahama, line 257: We only care about the message handling errors? This won't consider other possible kafka errors.

karapace/schema_reader.py
```diff
@@ -173,6 +173,7 @@ def __init__(
         self.last_check = time.monotonic()
         self.start_time = time.monotonic()
         self.startup_previous_processed_offset = 0
+        self.consecutive_unexpected_errors = 0
```
Claudenw, line 176: There was a comment that this would fail if it failed 3x.  While I see the variable `consecutive_unexpected_errors` and I see it being incremented.  There does not seem to be a check of it.   Either remove it or make it used in something.

karapace/schema_registry_apis.py
```diff
@@ -115,7 +115,14 @@ async def schema_registry_health(self) -> JsonObject:
...
         resp["schema_registry_coordinator_running"] = cs.is_running
         resp["schema_registry_coordinator_generation_id"] = cs.group_generation_id
-        return resp
+
+        healthy = True
+        if self.schema_registry.schema_reader.consecutive_unexpected_errors > 3:
+            # Check if there are unexpected errors in schema_reader.
+            # If there are more that 3 consecutive errors render service unhealthy.
+            healthy = False
+
```
Claudenw, line 124: I found the usage that I missed in my earlier comment.

Q: should the schema_reader  determine if it is healthy or should that be done here?  If, in future, there are more checks that determine if schema_reader is healthy will they all go here or does it make sense to put it in schema_reader as a method?

This does not pass the smell test for me, though I will admit that I am not 100% on python idioms and usage.

Claudenw:
Where are the test cases to show this works and to verify that it will not be broken later?

karapace/schema_reader.py
keejon: I would assume if there are kafka problems they would show up in consume call

keejon:
> Where are the test cases to show this works and to verify that it will not be broken later?

@Claudenw i added tests now, pls have another look

src/karapace/schema_reader.py
```diff
@@ -62,6 +63,11 @@
 KAFKA_CLIENT_CREATION_TIMEOUT_SECONDS: Final = 2.0
 SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS: Final = 5.0
 
+# If handle_messages throws at least UNHEALTHY_CONSECUTIVE_ERRORS
+# for UNHEALTHY_TIMEOUT_SECONDS the SchemaReader will be reported unhealthy
+UNHEALTHY_TIMEOUT_SECONDS: Final = 10.0
+UNHEALTHY_CONSECUTIVE_ERRORS: Final = 3
```
eliax1996, line 69: not mandatory for this pr, would be nice to have those configurable. Maybe depending on the load and the cluster size those numbers needs to be adjusted.

src/karapace/schema_reader.py
```diff
@@ -249,15 +258,44 @@ def run(self) -> None:
                     self.offset = self._get_beginning_offset()
                 try:
                     self.handle_messages()
+                    self.consecutive_unexpected_errors = 0
                 except ShutdownException:
                     self._stop_schema_reader.set()
                     shutdown()
                 except KafkaUnavailableError:
+                    self.consecutive_unexpected_errors += 1
```
eliax1996, line 266: would be nice to be able to differentiate the cause, like if kafka its down obviously karapace isn't going to work. But again not something I wanna see in this pr, as it is its a good improvement already

eliax1996 (approved):
LGTM

PR #959, opened 2024-09-23
tests, protobuf: add test on invalid reference schema

davide-armand:
Given a valid schema, if it references an invalid (corrupted) schema then:
- Exception CorruptKafkaRecordException  should be raised
- Warning Invalid Protobuf references should be logged

[EC-459]


davide-armand:
Branch renamed, replaced by https://github.com/Aiven-Open/karapace/pull/962

PR #960, opened 2024-09-24
test: select random free server ports with 0

jjaakola-aiven:

The previous port selection algorithm relied on port range. The range
was from 48700 to 49000. This is 300 ports of which some may be already
reserved in the system running tests. The free ports are allocated between
the pytest workers. With 4 workers there is 75 ports for each. This causes
flakiness in the tests as ports may not be available or are still in
wait after previous tests. In the test environment it is not necessary
to select ports from IANA dynamic port assignment range. Any free port
is ok and port 0 can be used for dynamically allocated free port.

### DIFF
--- a/tests/integration/test_karapace.py
+++ b/tests/integration/test_karapace.py
@@ -2,11 +2,10 @@
 Copyright (c) 2023 Aiven Ltd
 See LICENSE for details
 """
-from contextlib import ExitStack
+from contextlib import closing, contextmanager, ExitStack
 from karapace.config import set_config_defaults
 from pathlib import Path
 from tests.integration.utils.kafka_server import KafkaServers
-from tests.integration.utils.network import PortRangeInclusive
 from tests.integration.utils.process import stop_process
 from tests.utils import popen_karapace_all
 from typing import Iterator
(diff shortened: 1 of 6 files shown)


tests/integration/test_karapace.py
```diff
@@ -15,8 +14,15 @@
 import socket
 
 
+@contextmanager
+def allocate_port_no_reuse() -> Iterator[int]:
```
nosahama, line 18: **[nit-pick]**
Maybe `Generator[int]` as the return type, since you already did earlier, but doesn't hurt as generators as also iterators.

tests/integration/test_karapace.py
```diff
@@ -15,8 +14,15 @@
...
+    """Allocate random free port and do not allow reuse."""
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
+        sock.bind(("127.0.0.1", 0))
```
nosahama, line 21: > If host or port are ‘’ or 0 respectively the OS default behavior will be used.
Simpler and nice 👍

tests/integration/utils/network.py
```diff
@@ -2,82 +2,19 @@
...
-        finally:
-            # Remove the port at the end, this is a hack to give extra time for a TIME_WAIT socket to
-            # close, but it is not sufficient.
-            self._allocated.remove(port)
+@contextmanager
+def allocate_port() -> Iterator[int]:
+    """Allocate random free port."""
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
+        sock.bind(("127.0.0.1", 0))
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
nosahama, line 16: > The `SO_REUSEADDR` flag tells the kernel to reuse a local socket in `TIME_WAIT` state, without waiting for its natural timeout to expire.

nosahama (approved):

PR #961, opened 2024-09-24
test: add retries to schema coordinator workflow test

jjaakola-aiven:

References: #xxxxx

# Why this way

### DIFF
--- a/tests/integration/test_schema_coordinator.py
+++ b/tests/integration/test_schema_coordinator.py
@@ -21,8 +21,10 @@ from aiokafka.util import create_future, create_task
 from karapace.coordinator.schema_coordinator import Assignment, SchemaCoordinator, SchemaCoordinatorGroupRebalance
 from karapace.utils import json_encode
 from karapace.version import __version__
+from tenacity import retry, stop_after_delay, TryAgain, wait_fixed
 from tests.integration.utils.kafka_server import KafkaServers
-from typing import AsyncGenerator, Iterator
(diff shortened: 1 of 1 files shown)


tests/integration/test_schema_coordinator.py
```diff
@@ -82,12 +87,26 @@ async def get_client(
...
 
 
+@retry(stop=stop_after_delay(RETRY_TIME), wait=wait_fixed(RETRY_WAIT_SECONDS))
+async def wait_for_ready(coordinator: SchemaCoordinator) -> None:
+    if not coordinator.ready and coordinator.coordinator_id is None:
+        raise TryAgain()
+
+
+@retry(stop=stop_after_delay(RETRY_TIME), wait=wait_fixed(RETRY_WAIT_SECONDS))
+async def wait_for_primary_state(coordinator: SchemaCoordinator) -> None:
```
eliax1996, line 97: we should implement also an `async` version of the `retry`. Not sure if its smart enough to implement the `asyncio.wait` variant in case its decorating an async function.
In any case its an improvement for the flakiness that its the highest priority as now. Merging

eliax1996 (approved):
LGTM

tests/integration/test_schema_coordinator.py
eliax1996: https://tenacity.readthedocs.io/en/latest/#async-and-retry

Jarkko provided me this doc, already solved.

PR #962, opened 2024-09-25
tests, protobuf: add test on invalid reference schema

davide-armand:
Given a valid schema, if it references an invalid (corrupted) schema then:
- Exception CorruptKafkaRecordException should be raised
- Warning Invalid Protobuf references should be logged


nosahama (approved):
LGTM 🚀

PR #963, opened 2024-09-25
schema-reader: Log the erroring kafka message key

nosahama:
- we add the kafka message key to the log in the case of an error, this makes it easier to debug issues
- we also add a check for kafka messages without a key

### DIFF
--- a/karapace/schema_reader.py
+++ b/karapace/schema_reader.py
@@ -363,8 +363,14 @@ class KafkaSchemaReader(Thread):
 
                 assert message_key is not None
                 key = json_decode(message_key)
+            except AssertionError as exc:
+                LOG.warning("Empty msg.key() at offset %s", msg.offset())
+                self.offset = msg.offset()  # Invalid entry shall also move the offset so Karapace makes progress.
+                self.kafka_error_handler.handle_error(location=KafkaErrorLocation.SCHEMA_READER, error=exc)
+                continue  # [non-strict mode]
             except JSONDecodeError as exc:
-                LOG.warning("Invalid JSON in msg.key() at offset %s", msg.offset())
+                non_bytes_key = msg.key().decode()  # type: ignore[union-attr]
+                LOG.warning("Invalid JSON in msg.key(): %s at offset %s", non_bytes_key, msg.offset())
                 self.offset = msg.offset()  # Invalid entry shall also move the offset so Karapace makes progress.
                 self.kafka_error_handler.handle_error(location=KafkaErrorLocation.SCHEMA_READER, error=exc)
                 continue  # [non-strict mode]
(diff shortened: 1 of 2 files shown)


karapace/schema_reader.py
```diff
@@ -364,7 +364,13 @@ def consume_messages(self, msgs: list[Message], watch_offsets: bool) -> None:
                 assert message_key is not None
                 key = json_decode(message_key)
             except JSONDecodeError as exc:
-                LOG.warning("Invalid JSON in msg.key() at offset %s", msg.offset())
+                non_bytes_key = msg.key().decode()  # type: ignore[union-attr]
+                LOG.warning("Invalid JSON in msg.key(): %s at offset %s", non_bytes_key, msg.offset())
+                self.offset = msg.offset()  # Invalid entry shall also move the offset so Karapace makes progress.
+                self.kafka_error_handler.handle_error(location=KafkaErrorLocation.SCHEMA_READER, error=exc)
+                continue  # [non-strict mode]
+            except AssertionError as exc:
```
davide-armand, line 372: I wonder if it's possible to be more precise in catching exactly the assert on the `message_key` and not maybe others potentially raised in `json_decode()`

davide-armand:
Left a comment but LGTM

karapace/schema_reader.py
nosahama: Yes, I can move this catch above the json error catch.

davide-armand: Maybe with an if statement instead of assert/catch:
```python
        if message_key is None:
                LOG.warning("Empty msg.key() at offset %s", msg.offset())
                self.offset = msg.offset()  # Invalid entry shall also move the offset so Karapace makes progress.
                self.kafka_error_handler.handle_error(location=KafkaErrorLocation.SCHEMA_READER, error=exc)
                continue  # [non-strict mode]

```

nosahama: I get the idea, but I do not want to change the body of the try at this point, I am also not sure why `json.loads()` might raise an assertion error, I mean it's possible, but ideally a component should catch and parse all external facing errors.

eliax1996 (approved):
LGTM

karapace/schema_reader.py
```diff
@@ -363,8 +363,14 @@ def consume_messages(self, msgs: list[Message], watch_offsets: bool) -> None:
 
                 assert message_key is not None
                 key = json_decode(message_key)
+            except AssertionError as exc:
+                LOG.warning("Empty msg.key() at offset %s", msg.offset())
+                self.offset = msg.offset()  # Invalid entry shall also move the offset so Karapace makes progress.
+                self.kafka_error_handler.handle_error(location=KafkaErrorLocation.SCHEMA_READER, error=exc)
+                continue  # [non-strict mode]
```
aiven-anton, line 370: Hmm, this looks like a bit of a sketchy idea, an `AssertionError` could come from deeper in the code, and this essentially abuses the `assert` to function as a `goto`. `AssertionError` should really never be caught, bar some special cases in test harnessing code.

aiven-anton: If an `AssertionError` is raised from `msg.key()` or `msg.error()`, the correct behavior is for the code to fall over, and be restarted by the scheduler.

nosahama: Yes this had existed prior to this change, we can take a look later.

PR #964, opened 2024-09-25
chore: update tests to use Kafka 3.4.1

jjaakola-aiven:

References: #xxxxx

# Why this way


eliax1996 (approved):
LGTM

PR #971, opened 2024-10-02
fix: add Protopace go source to be included in MANIFEST.in

jjaakola-aiven:

References: #xxxxx

# Why this way


keejon (approved):
👍

PR #973, opened 2024-10-08
feat: add rest endpoint to display current config

eliax1996:
This endpoint could be used for testing purposes or system monitoring, e.g. to verify when a certain configuration as been correctly parsed and considered at runtime by karapace

References: #xxxxx

# Why this way


eliax1996:
too risky at exposing. We don't have a good acl to represent this type of requirement. Lets close it

PR #976, opened 2024-10-16
fix: Avro dataclass introspect typing

jjaakola-aiven:

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/avro_dataclasses/introspect.py
+++ b/src/karapace/avro_dataclasses/introspect.py
@@ -42,10 +42,17 @@ def _field_type_array(field: Field, origin: type, type_: object) -> AvroType:
     else:
         (inner_type,) = get_args(type_)
 
+    items: AvroType
+    if is_dataclass(inner_type):
+        assert isinstance(inner_type, type)
+        items = record_schema(inner_type)
+    else:
+        items = _field_type(field, inner_type)
+
     return {
         "name": f"one_of_{field.name}",
         "type": "array",
-        "items": (record_schema(inner_type) if is_dataclass(inner_type) else _field_type(field, inner_type)),
+        "items": items,
     }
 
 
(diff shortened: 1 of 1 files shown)


src/karapace/avro_dataclasses/introspect.py
```diff
@@ -128,7 +135,7 @@ def _field_type(field: Field, type_: object) -> AvroType:  # pylint: disable=too
 T = TypeVar("T", str, int, bool, Enum, None)
 
 
-def transform_default(type_: type[T], default: T) -> str | int | bool | None:
+def transform_default(type_: type[T] | str, default: T) -> str | int | bool | None:
```
Claudenw, line 138: What is the failure case that necessitates the addition of `str` here?  Just curious.

jjaakola-aiven: This is pure typing correction. It looks that primitive types use string instead of `type[T]`.

keejon (approved):
LGTM 👍

src/karapace/avro_dataclasses/introspect.py
aiven-anton: Tracking this to here:

https://github.com/python/typeshed/blob/6feca188689e14cdacd8dbf8bc9cf0635993027a/stdlib/dataclasses.pyi#L111

When originally authored, this type was `type`, only changed recently and I guess propagated in most recent mypy bump:

https://github.com/python/typeshed/commit/6b2033a783ad4f963979c761f32b14d1d9374bc5

aiven-anton: `_field_type` doesn't give a type error because it accepts `object` and makes no assumptions. It's probably somewhat likely to break when the new deferred annotations arrive, but I wouldn't deal with that until it's possible to test with.

PR #977, opened 2024-10-16
chore: fix schema tool source layout check return code handling

jjaakola-aiven:

References: #xxxxx

# Why this way


aiven-anton:
Shouldn't this be possible to remove now that the change is made? We can always assume src layout now?

aiven-anton (approved):

jjaakola-aiven:
Closing, #978 removes the source layout checking. After 4.0.0 the assumption is the source layout. Last release with non-source layout is 3.16.0.

PR #980, opened 2024-10-17
chore: pip compile to upgrade requirements

jjaakola-aiven:

References: #xxxxx

# Why this way


eliax1996 (approved):
LGTM, lets wait the CI before merging

PR #982, opened 2024-10-24
tests,backup: fix flakiness by using current timestamps

davide-armand:
Using some (old) fixed timestamps when producing messages was causing the
messages to be automatically deleted because past message retention.

I think most of the time the test was managing to complete successfully
before the retention kicked in, making it fail only sometimes.

### DIFF
--- a/tests/integration/backup/test_v3_backup.py
+++ b/tests/integration/backup/test_v3_backup.py
@@ -119,14 +119,13 @@ def test_roundtrip_from_kafka_state(
     admin_client.update_topic_config(new_topic.topic, {"max.message.bytes": "999"})
 
     # Populate topic.
-    producer.send(
+    first_record_fut = producer.send(
         new_topic.topic,
         key=b"bar",
         value=b"foo",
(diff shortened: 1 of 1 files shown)


aiven-anton:
Nice finding, I guess alternative fix could be to disable retention? (Asking because we most likely have the exact same issue here: https://github.com/aiven/backups-for-apache-kafka/issues/26)

eliax1996 (approved):
LGTM, great finding. How did you realized it?

davide-armand:
> Nice finding, I guess alternative fix could be to disable retention? (Asking because we most likely have the exact same issue in an internal project)

Ah yes I guess that would work too, did no think about that

davide-armand:
> LGTM, great finding. How did you realized it?

@eliax1996 The beginning offset of the topic was mysteriously changing from 0 to the 2 (end offset) after a few seconds.
Doing `consumer.seek` to offset 0 was failing.
Then I simplified the test until I found out about the timestamps..

PR #986, opened 2024-10-30
fix: proto3 optionals from base64 encoded proto descriptors

keejon:
This fixes  deserialization of optionals in proto3 descriptors as they are represented by synthetic oneofs.

### DIFF
--- a/src/karapace/protobuf/serialization.py
+++ b/src/karapace/protobuf/serialization.py
@@ -93,17 +93,31 @@ def _deserialize_msg(msgtype: Any) -> MessageElement:
     for nested_enum in msgtype.enum_type:
         nested_types.append(_deserialize_enum(nested_enum))
 
-    one_ofs: list[OneOfElement] = [OneOfElement(oneof.name) for oneof in msgtype.oneof_decl]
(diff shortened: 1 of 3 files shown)


src/karapace/protobuf/serialization.py
```diff
@@ -93,17 +93,31 @@ def _deserialize_msg(msgtype: Any) -> MessageElement:
...
         sf = _deserialize_field(f)
-        if f.HasField("oneof_index"):
+        is_oneof = f.HasField("oneof_index")
+        is_proto3_optional = f.HasField("oneof_index") and f.HasField("proto3_optional") and f.proto3_optional
+        if is_proto3_optional:
+            # Every proto3 optional field is placed into a one-field oneof, called a "synthetic" oneof,
+            # as it was not present in the source .proto file.
+            # This will make sure that we don't interpret those optionals as oneof.
+            one_ofs[f.oneof_index] = None
+            fields.append(sf)
```
eliax1996, line 107: to be honest I don't really like the idea of marking the syntetic ones as `None` but I have no better ideas on how to keep the information in the type system 😄
Nice work! Thanks for the fix

eliax1996 (approved):
LGTM

PR #987, opened 2024-10-31
Avro references poc implementation using avro lib

davide-armand:
This is a proof of concept for implementing Avro reference using Avro python library.
I used the following PR (implementation using unions) as the base: https://github.com/Aiven-Open/karapace/pull/926

What works:
- Direct reference (schema1 --> schema2)
- Recursive reference (schema1 --> schema1)

What does not work:
- Indirect references (schema1 --> schema2 --> schema3) - maybe not needed, not sure Schema Registry supports this
- ... probably other things as well

### DIFF
--- a/tests/integration/test_schema_avro_references.py
+++ b/tests/integration/test_schema_avro_references.py
@@ -54,6 +54,30 @@ SCHEMA_PERSON = {
     ],
 }
 
+SCHEMA_PERSON_RECURSIVE = {
+    "type": "record",
+    "name": "PersonRecursive",
+    "namespace": "com.netapp",
+    "fields": [
+        {"name": "name", "type": "string"},
+        {"name": "age", "type": "int"},
+        {"name": "job", "type": "Job"},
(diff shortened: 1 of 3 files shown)


tests/integration/test_schema_avro_references.py
```diff
@@ -49,19 +49,43 @@
...
+
 SCHEMA_PERSON_AGE_INT_LONG = {
     "type": "record",
     "name": "Person",
     "namespace": "com.netapp",
     "fields": [
         {"name": "name", "type": "string"},
         {"name": "age", "type": "long"},
-        {"name": "address", "type": "Address"},
+        # {"name": "address", "type": "Address"},
```
davide-armand, line 88: Commented out indirect refs since they don't work, so that other tests can be executed

davide-armand: Maybe we don't need to support indirect refs, it's unclear if Schema Registry supports them

davide-armand:
Closing this as it was just a proof of concept

PR #990, opened 2024-11-07
fix: don't send all aiokafka log.error to sentry

kmichel-aiven:
These log events are numerous and costly. When these `log.error` events are related to real error, they cause exceptions which are still properly handled and recorded by Sentry, no need to also send individual events for all the `log.error` in addition to the exceptions. 

References: #xxxxx

# Why this way
The same logic was already applied for the `kafka` library.


jjaakola-aiven (approved):
LGTM

PR #993, opened 2024-11-19
rest-proxy: split out into standalone module

nosahama:
Make `rest-proxy` component standalone module still running the `rapu` based webserver, leaving only `schema_registry` and shared components in `src/karapace`.

References: #xxxxx

# Why this way
We can make future changes that aid in providing the karapace components as separate services.


eliax1996 (approved):
Except the minor note LGTM

bin/smoke-test-rest-proxy.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+retries=5
+
+for ((i = 0; i <= retries; i++)); do
+    response=$(curl --silent --verbose --fail http://localhost:8083/topics)
+
+    if [[ $response == '["_schemas","__consumer_offsets"]' ]]; then
+        echo "Ok!"
+        break
```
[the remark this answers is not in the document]
nosahama: I didn't plan to change anything in this file except the port.

src/rest_proxy/__main__.py
```diff
@@ -0,0 +1,93 @@
...
+    parser = argparse.ArgumentParser(
+        prog=PROGRAM_NAME,
+        description="Karapace rest proxy: exposes an API over common Kafka operations, your Kafka essentials in one tool",
+    )
+    parser.add_argument("--version", action="version", help="show program version", version=karapace_version.__version__)
+    parser.add_argument("config_file", help="configuration file path", type=argparse.FileType())
+    arg = parser.parse_args()
+
+    with closing(arg.config_file):
+        config = read_config(arg.config_file)
```
eliax1996, line 65: if you aren't using the closing arg I don't think you are closing it correctly.
Shouldn't be:
```python3
with closing(foo) as bar:
   config = read_config(bar)
```
?

bin/smoke-test-rest-proxy.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+retries=5
+
+for ((i = 0; i <= retries; i++)); do
+    response=$(curl --silent --verbose --fail http://localhost:8083/topics)
+
+    if [[ $response == '["_schemas","__consumer_offsets"]' ]]; then
```
jjaakola-aiven, line 8: Would `__consumer_offsets` be enough here, the `_schemas` is schema registry special.

container/start.sh
```diff
@@ -27,6 +27,22 @@ rest)
     echo "Starting Karapace REST API"
     exec python3 -m karapace.karapace_all /opt/karapace/rest.config.json
     ;;
+rest_proxy)
```
jjaakola-aiven, line 30: The case `rest)` is not required anymore.

src/rest_proxy/__main__.py
```diff
@@ -0,0 +1,93 @@
...
+    _configure_logging(config=config)
+
+    app = KafkaRest(config=config)
+
+    logging.log(logging.INFO, "\n%s\nStarting %s\n%s", ("=" * 50), PROGRAM_NAME, ("=" * 50))
+
+    config_without_secrets = {}
+    for key, value in config.items():
+        if "password" in key:
+            value = "****"
```
eliax1996, line 78: this is dangerous, you should keep the info about the secrecy of the field in the context of where its created. Otherwise if tomorrow we add another field like e.g. `sentry_api_token` this isn't filtered and there isn't any hint about this indirect dependency.
I suggest to add the filtering in the object itself, you can add to the `Config` object a method like:
```

SENSITIVE_FIELDS: Final = ("registry_password", "ssl_password", "sasl_plain_password", "sasl_oauth_token")

@staticmethod
def is_field_a_secret(self, field_name: str) -> bool:
   if field_name not in Config:
      raise ValueError(f"{field_name} is not known")
   return field_name not in SENSITIVE_FIELDS
```

In that way the filtering logic its "near" to where the addition/removal/renaming of the property is, its a property of the data structure rather than a piece of coupled code put randomly in another file and that the developer should know to look before adding/removing something far away.

PS I think the logic wasn't covering the `sasl_oauth_token`

eliax1996 (changes requested):
A couple of notes, again near to the merge 😄

src/rest_proxy/__main__.py
nosahama: I can take a look, once again, this is existing code 😉

src/rest_proxy/__main__.py
nosahama: Once again, this is existing code 😉 
I'd reason around your suggestions and see to add/fix or create follow up items 👍

nosahama:
Setting as draft PR, so I can run some system test against this hash.

bin/smoke-test-rest-proxy.sh
nosahama: Also this is existing code/logic.

PR #994, opened 2024-11-22
schema_registry: standalone module for SR with basic dependency injection

nosahama:
We move schema registry new FastAPI based implementation into it's own module `schema_registry` and add dependency usage with an example of the application config injection, we can build up from here.


src/karapace/logging.py
```diff
@@ -0,0 +1,45 @@
...
+See LICENSE for details
+"""
+
+from karapace.config import Config
+
+import logging
+import sys
+
+
+def configure_logging(*, config: Config) -> None:
```
nosahama, line 12: Even here, we can already use the DI, but let's move a bit and we build on this.

src/schema_registry/__main__.py
```diff
@@ -0,0 +1,77 @@
...
+async def lifespan(_: FastAPI) -> AsyncGenerator[None, None]:
+    schema_registry: KarapaceSchemaRegistry | None = None
+    authorizer: AuthenticatorAndAuthorizer | None = None
+    try:
+        schema_registry = await SchemaRegistryDependencyManager.get_schema_registry()
+        await schema_registry.start()
+        await schema_registry.get_master()
+        authorizer = AuthorizationDependencyManager.get_authorizer()
+        if authorizer is not None:
+            await authorizer.start(StatsDependencyManager.get_stats())
```
nosahama, line 41: All these will be moved to the DI `container`.

src/schema_registry/dependencies/stats_dependeny.py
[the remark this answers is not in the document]
nosahama: I will then kill the file as a whole and move the statsd to the DI

PR #995, opened 2024-11-27
feat: fastapi `schema-registry` module with DI and E2E tests

nosahama:

Some changes based off the work from the [jjaakola-aiven-fastapi](https://github.com/Aiven-Open/karapace/tree/jjaakola-aiven-fastapi) branch, majorly:
- integrations tests which are failing need to be checked and resolved
  - some of the tests can be moved to the `e2e` folder
  - it'd be nice to stop spinning up all these kafka processes
  - it'd be nice to move this into the docker tests environment
  - i am not making any changes to the integration tests setup, so it's clear that we aren't changing logic that should verify the operability of our services
  - the integration tests fail now with issues spinning up kafka, so we can check that now or after we merge this work
- unit tests now run within a docker compose environment
- `e2e` tests were introduced, which also run within a docker compose environment
  - the plan is to extract some test out of the `integration` tests folder slowly into the `e2e` tests folder
  - `e2e` tests run against the karapace docker compose services: i.e. `schema-registry/rest` rather than local processes
- karapace schema registry is now it's own module which houses all the FastAPI related things
- karapace rest proxy still runs on `rapu`
- we add dependency injection to the project
  - wiring the dependencies for both sr and rest
  - config is loaded on app startup
  - dependencies are injected into route handlers and other modules
  - modules with dependencies are wired on app startup
- config is now based on pydantic-settings
  - settings are loaded via an `.env` file
  - base config is added as a yaml and packaged with the built python library
  - base config loads the reference to the `.env` file needed to populate the rest of the config
  - we can override the settings values directly via environment variables or via a helper method on the settings object itself
  - previously some other values were set, i.e. `sentry` config was added, **we need to revisit this.**
- commits show progress
- initial draft PR #994 was rushed and contained more changes, this is a split out of that PR which will be closed at some point
- we would need to check the coverage mechanism now with the new module and split


bin/smoke-test-registry.sh
```diff
@@ -6,6 +6,7 @@ for ((i = 0; i <= retries; i++)); do
     response=$(
         curl --silent --verbose --fail --request POST \
             --header 'Content-Type: application/vnd.schemaregistry.v1+json' \
+            --header 'Authorization: Basic Og==' \
```
nosahama, line 9: This are empty credentials, i.e. `username="" | password=""`, this is needed as now on FastAPI we require the credentials to be at least present, even if it's empty, we can revise this later as it'd be now a breaking change, but shouldn't hurt too much.

.github/workflows/tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.9', '3.10', '3.11', '3.12' ]
```
nosahama, line 23: Dropping 3.9 from the tests matrix due to this issue - https://github.com/pydantic/pydantic-settings/issues/298
We'd need to revisit this at a later time.

src/schema_registry/user.py
```diff
@@ -0,0 +1,36 @@
...
+from schema_registry.container import SchemaRegistryContainer
+from typing import Annotated
+
+
+@inject
+async def get_current_user(
+    credentials: Annotated[HTTPBasicCredentials, Depends(HTTPBasic(auto_error=False))],
+    authorizer: AuthenticatorAndAuthorizer = Depends(Provide[SchemaRegistryContainer.karapace_container.authorizer]),
+) -> User:
+    if authorizer and not credentials:
```
nosahama, line 19: I am thinking to flip this logic, as now there will always be an authorizer, even if it's the no auth authorizer, so I will switch the operands.

.github/workflows/tests.yml
```diff
@@ -40,22 +40,43 @@ jobs:
       with:
         go-version: '1.21.0'
 
-    - run: make install-dev
-    - run: make unit-tests
+    - name: Resolve Karapace version
+      run: echo KARAPACE_VERSION=4.1.1.dev44+gac20eeed.d20241205 >> $GITHUB_ENV
```
nosahama, line 44: Need a better way to resolve the version here, cannot do `make install` as now the `src` is copied into the dev image.

tests/integration/test_schema_protobuf.py
```diff
@@ -1326,30 +1330,42 @@ async def test_protobuf_update_ordering(registry_async_client: Client) -> None:
 """
 
 
-@pytest.mark.parametrize(
-    "registry_cluster, status",
-    [({"config": {}}, 404), ({"config": {"use_protobuf_formatter": True}}, 200)],
-    indirect=["registry_cluster"],
```
nosahama, line 1332: Not sure how this test was able to assert the behaviour, not it is really doing what it should do. 
Test against 2 SR, 1 without protobuf formatter and the other with the formatter.

jjaakola-aiven (approved):
LGTM, getting rather large and preferable to merge. Most integration tests pass, forwarding tests have some issues. Linting and other issues can be resolved in subsequent PRs.

PR #998, opened 2024-11-29
fix: add the context output back to container GH Action workflow

jjaakola-aiven:

References: #xxxxx

# Why this way


eliax1996 (approved):
Let’s hope this finally fixes it 😁

PR #1004, opened 2024-12-09
Linting fixes

jjaakola-aiven:

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/forward_client.py
+++ b/src/karapace/forward_client.py
@@ -46,9 +46,8 @@ class ForwardClient:
                         status_code=response.status,
                         headers=response.headers,
                     )
-                else:
-                    return PlainTextResponse(
-                        content=await response.text(),
-                        status_code=response.status,
-                        headers=response.headers,
-                    )
+                return PlainTextResponse(
+                    content=await response.text(),
+                    status_code=response.status,
+                    headers=response.headers,
+                )
--- a/src/karapace/forward_client.py
+++ b/src/karapace/forward_client.py
@@ -3,9 +3,11 @@ Copyright (c) 2024 Aiven Ltd
 See LICENSE for details
 """
 
-from fastapi import Request, Response
-from fastapi.responses import JSONResponse, PlainTextResponse
+from fastapi import HTTPException, Request, status
+from karapace.utils import json_decode
 from karapace.version import __version__
+from pydantic import BaseModel
+from typing import overload, TypeVar, Union
 
 import aiohttp
 import async_timeout
@@ -14,16 +16,30 @@ import logging
 LOG = logging.getLogger(__name__)
 
 
+BaseModelResponse = TypeVar("BaseModelResponse", bound=BaseModel)
+SimpleTypeResponse = TypeVar("SimpleTypeResponse", bound=Union[int, list[int]])
+
+
 class ForwardClient:
     USER_AGENT = f"Karapace/{__version__}"
 
-    def __init__(self):
+    def __init__(self) -> None:
         self._forward_client: aiohttp.ClientSession | None = None
 
     def _get_forward_client(self) -> aiohttp.ClientSession:
         return aiohttp.ClientSession(headers={"User-Agent": ForwardClient.USER_AGENT})
 
-    async def forward_request_remote(self, *, request: Request, primary_url: str) -> Response:
+    def _acceptable_response_content_type(self, *, content_type: str) -> bool:
+        return (
+            content_type.startswith("application/") and content_type.endswith("json")
+        ) or content_type == "application/octet-stream"
+
+    async def _forward_request_remote(
+        self,
+        *,
+        request: Request,
+        primary_url: str,
+    ) -> bytes:
         LOG.info("Forwarding %s request to remote url: %r since we're not the master", request.method, request.url)
         timeout = 60.0
         headers = request.headers.mutablecopy()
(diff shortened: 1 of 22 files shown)


src/karapace/forward_client.py
```diff
@@ -14,16 +16,30 @@
...
+
+    async def _forward_request_remote(
+        self,
+        *,
+        request: Request,
+        primary_url: str,
+    ) -> bytes:
         LOG.info("Forwarding %s request to remote url: %r since we're not the master", request.method, request.url)
         timeout = 60.0
         headers = request.headers.mutablecopy()
```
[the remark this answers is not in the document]
nosahama: There's also this piece of code: `func = getattr(self._get_forward_client(), request.method.lower())`, this would mean that we generate a new client every time we want to forward a request. Maybe we move the client initialization to the `__init__()` function, assign directly to `self._forward_client` and that will get reused for the whole app runtime and we can even add a `close()` function that cleans up the client session.

src/karapace/forward_client.py
```diff
@@ -37,18 +53,34 @@ async def forward_request_remote(self, *, request: Request, primary_url: str) ->
...
-                        headers=response.headers,
-                    )
-                else:
-                    return PlainTextResponse(
-                        content=await response.text(),
-                        status_code=response.status,
-                        headers=response.headers,
-                    )
+                if self._acceptable_response_content_type(content_type=response.headers.get("Content-Type")):
+                    return await response.text()
```
nosahama, line 60: Would there be anytime when we expect `text/plain`?

src/karapace/forward_client.py
```diff
@@ -37,18 +53,34 @@ async def forward_request_remote(self, *, request: Request, primary_url: str) ->
...
+    async def forward_request_remote(self, *, request: Request, primary_url: str, response_type: type[T]) -> T:
+        ...
+
+    @overload
+    async def forward_request_remote(self, *, request: Request, primary_url: str, response_type: type[P]) -> P:
+        ...
+
+    async def forward_request_remote(self, *, request: Request, primary_url: str, response_type: type[T] | type[P]) -> T | P:
+        body = await self._forward_request_remote(request=request, primary_url=primary_url)
+        if response_type == int:
```
nosahama, line 80: Why not using `isinstance()`?

src/karapace/forward_client.py
```diff
@@ -37,18 +53,34 @@ async def forward_request_remote(self, *, request: Request, primary_url: str) ->
...
+            return int(body)  # type: ignore[return-value]
+        if response_type == list[int]:
+            return json_decode(body, assume_type=list[int])  # type: ignore[return-value]
+        if issubclass(response_type, BaseModel):
+            return response_type.model_validate_json(body)  # type: ignore[return-value]
```
nosahama, line 85: If it's a `BaseModel`, it should already have been validated from the source response, unless there's no other way to load the text into the object.

src/karapace/forward_client.py
```diff
@@ -37,18 +53,34 @@ async def forward_request_remote(self, *, request: Request, primary_url: str) ->
...
-                        status_code=response.status,
-                        headers=response.headers,
-                    )
-                else:
-                    return PlainTextResponse(
-                        content=await response.text(),
-                        status_code=response.status,
-                        headers=response.headers,
-                    )
+                if self._acceptable_response_content_type(content_type=response.headers.get("Content-Type")):
```
nosahama, line 59: What if the server cannot be reached, i.e. there's even no response, etc, or 5xx status.

src/karapace/forward_client.py
```diff
@@ -14,16 +16,30 @@
 LOG = logging.getLogger(__name__)
 
 
+T = TypeVar("T", bound=BaseModel)
```
nosahama, line 19: Should we reason around some better naming.

src/karapace/forward_client.py
jjaakola-aiven: Error cases, I can't think of anything else.

src/karapace/forward_client.py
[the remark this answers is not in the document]
jjaakola-aiven: Yes, probably not in this PR though.

src/karapace/forward_client.py
jjaakola-aiven: The `response_type` is a type not an `int` object.

src/karapace/forward_client.py
jjaakola-aiven: Uh, Pydantic 1.x has `parse_raw` which is deprecated in 2.x. I think we got to support 1.x so I'll change to `parse_raw`.

src/karapace/forward_client.py
jjaakola-aiven: Handling improvement to do later.

src/karapace/forward_client.py
jjaakola-aiven: Done.

src/karapace/forward_client.py
jjaakola-aiven: Done.

src/karapace/forward_client.py
nosahama: Ah yh makes sense, hence the type vars 👍

src/karapace/forward_client.py
```diff
@@ -37,18 +53,52 @@ async def forward_request_remote(self, *, request: Request, primary_url: str) ->
...
+    async def forward_request_remote(
+        self,
+        *,
+        request: Request,
+        primary_url: str,
+        response_type: type[BaseModelResponse] | type[SimpleTypeResponse],
+    ) -> BaseModelResponse | SimpleTypeResponse:
+        body = await self._forward_request_remote(request=request, primary_url=primary_url)
+        if response_type == int:
+            return int(body)  # type: ignore[return-value]
```
nosahama, line 99: Maybe a cast will work to avoid the `# type: ignore[return-value]`, we can cast to the `SimpleTypeResponse`, but not sure if this works, will test in another PR.

src/karapace/forward_client.py
```diff
@@ -14,16 +16,30 @@
 LOG = logging.getLogger(__name__)
 
 
+BaseModelResponse = TypeVar("BaseModelResponse", bound=BaseModel)
+SimpleTypeResponse = TypeVar("SimpleTypeResponse", bound=Union[int, list[int]])
```
nosahama, line 20: Do we need the `Union`, maybe `|` works, will test it later.

tests/unit/test_forwarding_client.py
```diff
@@ -0,0 +1,153 @@
...
+            test_name="application/vnd.schemaregistry.v1+json", content_type="application/vnd.schemaregistry.v1+json"
+        ),
+        ContentTypeTestCase(
+            test_name="application/vnd.schemaregistry+json", content_type="application/vnd.schemaregistry+json"
+        ),
+        ContentTypeTestCase(test_name="application/octet-stream", content_type="application/octet-stream"),
+    ],
+    ids=str,
+)
+async def test_forward_request_with_basemodel_response(testcase: ContentTypeTestCase) -> None:
```
nosahama, line 45: Thank you.

nosahama (approved):

PR #1005, opened 2024-12-10
Drop python39 support

jjaakola-aiven:

References: #xxxxx

# Why this way


nosahama (approved):
Thank you 🚀

PR #1006, opened 2024-12-10
forwarding_client: use single session and close on shutdown

nosahama:
We use a single client session for the forwarded requests and further close the session on application shutdown.

### DIFF
--- a/tests/unit/test_forwarding_client.py
+++ b/tests/unit/test_forwarding_client.py
@@ -1,7 +1,7 @@
 """
-karapace - schema registry authentication and authorization tests
+karapace - master forwarding tests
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_forwarding_client.py
```diff
@@ -42,112 +57,100 @@ class ContentTypeTestCase(BaseTestCase):
     ],
     ids=str,
 )
-async def test_forward_request_with_basemodel_response(testcase: ContentTypeTestCase) -> None:
```
nosahama, line 45: Tests remain the same, `forward_client` is now a fixture and the context manager scope is removed.

jjaakola-aiven (approved):

PR #1007, opened 2024-12-11
instrumentation: fastapi prometheus middleware

nosahama:
Added a middleware to instrument the fastapi requests.

Screenshot 2024-12-11 at 14 13 37


src/schema_registry/middlewares/prometheus.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@inject
+async def prometheus_middleware(
+    request: Request,
+    call_next: Callable[[Request], Awaitable[Response]],
+    prometheus: PrometheusInstrumentation = Depends(Provide[SchemaRegistryContainer.karapace_container.prometheus]),
+) -> Response:
+    # Set start time for request
+    setattr(request.state, prometheus.START_TIME_REQUEST_KEY, time.time())
```
jjaakola-aiven, line 25: Use `time.monotonic` instead.

src/schema_registry/middlewares/prometheus.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Extract request labels
+    path = request.url.path
```
jjaakola-aiven, line 28: This can cause quite high cardinality, can we align this with the path template. E.g. `/subjects/<subject>/versions`?

src/schema_registry/middlewares/prometheus.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Increment requests in progress before response handler
+    prometheus.karapace_http_requests_in_progress.labels(method=method, path=path).inc()
+
+    # Call request handler
+    response: Response = await call_next(request)
+
+    # Instrument request duration
+    prometheus.karapace_http_requests_duration_seconds.labels(method=method, path=path).observe(
+        time.time() - getattr(request.state, prometheus.START_TIME_REQUEST_KEY)
```
jjaakola-aiven, line 39: `time.monotonic`.

jjaakola-aiven (changes requested):

src/schema_registry/middlewares/prometheus.py
nosahama: 👍 very much agree, just skipped my mind.

src/schema_registry/middlewares/prometheus.py
nosahama: Yes I can think around it, it was similar thought I had on the tracing PR #1009, where I extract the resource from the path as I didn't find a way to get the template.

In prometheus tho, I do not think this should be a problem, major cardinality issues arise with too many labels. To instrument, might be necessary to see exactly the path that's being checked, but I also get your point. Let me reason around it and get back.

nosahama: Looking at this discussion - https://github.com/fastapi/fastapi/discussions/12975, the route template is only accessed after a successful response, I do need the path already to instrument the requests in progress. 
Unless we want to fallback and extract the base resource, i.e. `resource = request.url.path.split("/")[1]`. From a template like `/subjects/{subject}/versions` and a request for `/subjects/test-topic-value/versions`, this will yield `subjects`.

nosahama:
I will close this in favour of the OTel metrics adding in #1011

PR #1011, opened 2024-12-19
feat: OTel metrics & metering

nosahama:
- we add OTel metering and HTTP metrics to the service.
- the metrics are exported to `prometheus` via the `opentelemetry-collector` as shown below

Screenshot 2024-12-19 at 10 47 58

- no changes to  the `grafana` dashboard shows that the `prometheus` source still works as expected

Screenshot 2024-12-19 at 10 49 00

# Follow up
- update the `schema_reader` metrics to use the OTel meter
- maybe deprecate `statsd`
- close the prometheus PR #1007 as we now use the OTel metering and exporting to prometheus, thus native prometheus instrumentation is not needed

### DIFF
--- a/src/schema_registry/telemetry/meter.py
+++ b/src/schema_registry/telemetry/meter.py
@@ -8,7 +8,12 @@ from karapace.config import Config
 from karapace.container import KarapaceContainer
 from opentelemetry import metrics
 from opentelemetry.exporter.otlp.proto.grpc.metric_exporter import OTLPMetricExporter
-from opentelemetry.sdk.metrics.export import ConsoleMetricExporter, MetricReader, PeriodicExportingMetricReader
+from opentelemetry.sdk.metrics.export import (
+    ConsoleMetricExporter,
+    MetricExporter,
+    MetricReader,
+    PeriodicExportingMetricReader,
+)
 from typing import Final
 
 
@@ -23,7 +28,7 @@ class Meter:
     @staticmethod
     @inject
     def get_metric_reader(config: Config = Provide[KarapaceContainer.config]) -> MetricReader:
-        exporter = ConsoleMetricExporter()
+        exporter: MetricExporter = ConsoleMetricExporter()
         if config.telemetry.otel_endpoint_url:
             exporter = OTLPMetricExporter(endpoint=config.telemetry.otel_endpoint_url)
         return PeriodicExportingMetricReader(
--- a/src/schema_registry/telemetry/meter.py
+++ b/src/schema_registry/telemetry/meter.py
@@ -4,7 +4,7 @@ See LICENSE for details
 """
 
 from dependency_injector.wiring import inject, Provide
-from karapace.config import Config
+from karapace.config import Config, KarapaceTelemetryOTelExporter
 from karapace.container import KarapaceContainer
 from opentelemetry import metrics
 from opentelemetry.exporter.otlp.proto.grpc.metric_exporter import OTLPMetricExporter
@@ -13,7 +13,32 @@ from opentelemetry.sdk.metrics.export import (
     MetricExporter,
     MetricReader,
     PeriodicExportingMetricReader,
+    MetricsData,
+    MetricExportResult,
 )
+from typing import Any
+
+
+class NOOPMetricExporter(MetricExporter):
+    """Implementation of :class:`MetricExporter` that does nothing.
+
+    This class is intended to be used when metrics exporting to an OTel backend is disabled
+    and the ConsoleExporter is too verbose to be used.
+    """
+
+    def export(
+        self,
+        metrics_data: MetricsData,
+        timeout_millis: float = 0,
+        **kwargs: Any,
+    ) -> MetricExportResult:
+        return MetricExportResult.SUCCESS
+
+    def shutdown(self, timeout_millis: float = 0, **kwargs: Any) -> None:
+        pass
+
+    def force_flush(self, _: float = 0) -> bool:
+        return True
 
 
 class Meter:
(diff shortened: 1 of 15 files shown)


src/schema_registry/telemetry/middleware.py
```diff
@@ -20,12 +23,59 @@ async def telemetry_middleware(
     request: Request,
     call_next: Callable[[Request], Awaitable[Response]],
     tracer: Tracer = Provide[TelemetryContainer.tracer],
+    meter: Meter = Provide[TelemetryContainer.meter],
 ) -> Response:
     resource = request.url.path.split("/")[1]
     with tracer.get_tracer().start_as_current_span(name=f"{request.method}: /{resource}", kind=SpanKind.SERVER) as span:
+        span.add_event("Creating metering resources")
+        karapace_http_requests_in_progress: UpDownCounter = meter.get_meter().create_up_down_counter(
```
jjaakola-aiven, line 31: Is this creating the counter or reusing the existing one? Similar question for the duration histogram and requests counter.

src/schema_registry/telemetry/middleware.py
```diff
@@ -20,12 +23,59 @@ async def telemetry_middleware(
...
+        method = request.method
+
+        # Increment requests in progress before response handler
+        span.add_event("Metering requests in progress (increase)")
+        karapace_http_requests_in_progress.add(amount=1, attributes={"method": method, "path": path})
+
+        # Call request handler
         tracer.update_span_with_request(request=request, span=span)
+        span.add_event("Calling request handler")
         response: Response = await call_next(request)
```
jjaakola-aiven, line 59: If call to next fails, are the subsequent calls to tracer and meters called?

src/schema_registry/telemetry/middleware.py
```diff
@@ -20,12 +23,59 @@ async def telemetry_middleware(
...
+        karapace_http_requests_total: Counter = meter.get_meter().create_counter(
+            name="karapace_http_requests_total",
+            description="Total Request Count for HTTP/TCP Protocol",
+        )
+
+        # Set start time for request
+        setattr(request.state, meter.START_TIME_KEY, time.monotonic())
+
+        # Extract request labels
+        path = request.url.path
```
jjaakola-aiven, line 49: This will create quite large cardinality when Karapace handles large number of topics and versions.

src/schema_registry/telemetry/middleware.py
nosahama: Good point, I looked up the docs for [get_meter()](https://opentelemetry-python.readthedocs.io/en/latest/api/metrics.html#opentelemetry.metrics.MeterProvider.get_meter) and they also mention that it is uncertain if subsequent calls to this function return the same instrument.

I will move them out and define them concretely in another module, so we are not trying to access `get_meter()` on every API request.

src/schema_registry/telemetry/middleware.py
nosahama: I will use the base resource name, i.e. `POST: /subjects/xyz` will instrument `/subjects`, this is the only way I can think of now, aside the custom `APIRouter`.

src/schema_registry/telemetry/middleware.py
nosahama: Really good catch, I will wrap in a `try/except/finally` block, run the instrumentation in the `except` if any issues and then reraise so it pro
pagates to FastAPI.

src/schema_registry/telemetry/metrics.py
```diff
@@ -0,0 +1,27 @@
...
+from opentelemetry.metrics import Counter, Histogram, UpDownCounter
+from schema_registry.telemetry.meter import Meter
+from typing import Final
+
+
+class Metrics:
+    START_TIME_KEY: Final = "start_time"
+
+    def __init__(self, meter: Meter):
+        self.karapace_http_requests_in_progress: UpDownCounter = meter.get_meter().create_up_down_counter(
```
jjaakola-aiven, line 15:
```suggestion
        self.karapace_http_requests_in_progress: Final[UpDownCounter] = meter.get_meter().create_up_down_counter(
```

src/schema_registry/telemetry/metrics.py
```diff
@@ -0,0 +1,27 @@
...
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+
+from opentelemetry.metrics import Counter, Histogram, UpDownCounter
+from schema_registry.telemetry.meter import Meter
+from typing import Final
+
+
+class Metrics:
```
jjaakola-aiven, line 11:
```suggestion
class HttpRequestMetrics:
...
    def new_request(self): # increment in progress and also the total requests.
    def finish_request(self, duration: float): # reduce in progress, record request duration.
```
What do you think?

nosahama: I like the idea, provide future scoping of metrics 👍

src/schema_registry/telemetry/meter.py
```diff
@@ -0,0 +1,34 @@
...
+class Meter:
+    @staticmethod
+    @inject
+    def get_meter(config: Config = Provide[KarapaceContainer.config]) -> metrics.Meter:
+        return metrics.get_meter_provider().get_meter(f"{config.tags.app}.meter")
+
+    @staticmethod
+    @inject
+    def get_metric_reader(config: Config = Provide[KarapaceContainer.config]) -> MetricReader:
+        exporter: MetricExporter = ConsoleMetricExporter()
```
nosahama, line 28: @jjaakola-aiven if you remember the recent issue with the systemd logging, maybe we should also implement a `noop` exporter for the metrics, I can check to see if we can reuse the one added for the traces.

The rest of the code can be reviewed.

jjaakola-aiven (approved):
LGTM

PR #1021, opened 2025-01-09
fix: max http request size for Karapace REST

jjaakola-aiven:

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/config.py
+++ b/src/karapace/config.py
@@ -146,6 +146,29 @@ class Config(BaseSettings):
             or f"{self.advertised_protocol}://{self.get_advertised_hostname()}:{self.get_advertised_port()}"
         )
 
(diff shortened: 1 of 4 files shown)


src/karapace/config.py
```diff
@@ -154,6 +177,7 @@ def to_env_str(self) -> str:
         return "\n".join(env_lines)
 
     def set_config_defaults(self, new_config: Mapping[str, str] | None = None) -> Config:
+        """TODO: Used from tests, this requires removal."""
```
keejon, line 180: why does this require removal?

jjaakola-aiven: This is not really used anymore, the config should have sane defaults without having code to mutate it.

keejon (approved):
LGTM 👍

PR #1027, opened 2025-01-21
fix: delete references if schema is deleted

keejon:

References: #xxxxx

# Why this way


src/karapace/in_memory_database.py
```diff
@@ -251,6 +251,10 @@ def insert_schema_version(
                 references=references,
             )
 
+            if references:
+                for ref in references:
+                    self.insert_referenced_by(subject=ref.subject, version=ref.version, schema_id=schema_id)
```
jjaakola-aiven, line 256: The `insert_referenced_by` locks `self.schema_lock_thread`, lock is re-entrant but maybe this function should be private and skip locking as this function does it already?

Similar locking happens with removal.

keejon: I made insert and remove referenced_by private and removed it from API level because messaging should be blocking until schema_ready has read and handled message with the latest offset. (have left locking in though because it does not hurt and we can be safe that nobody misses it in the future) Strange thing is that this is not working as expected though and i get 422 in integration tests. 🤔

keejon: @jjaakola-aiven tests are green again

jjaakola-aiven (approved):

PR #1028, opened 2025-01-22
feat: handle urlencoded forward slash in subject

jjaakola-aiven:

Adds support for url encoded forward slashes in subject names.
See: https://github.com/encode/starlette/issues/826
See: https://github.com/fastapi/fastapi/discussions/7328#discussioncomment-8443865

References: #xxxxx

# Why this way

### DIFF
--- a/src/schema_registry/routers/mode.py
+++ b/src/schema_registry/routers/mode.py
@@ -14,6 +14,7 @@ from schema_registry.routers.raw_path_router import RawPathRoute
 from schema_registry.routers.requests import ModeResponse
 from schema_registry.user import get_current_user
 from typing import Annotated
+from urllib.parse import unquote_plus
 
 mode_router = APIRouter(
     prefix="/mode",
@@ -44,6 +45,7 @@ async def mode_get_subject(
     authorizer: AuthenticatorAndAuthorizer = Depends(Provide[SchemaRegistryContainer.karapace_container.authorizer]),
     controller: KarapaceSchemaRegistryController = Depends(Provide[SchemaRegistryContainer.schema_registry_controller]),
 ) -> ModeResponse:
+    subject = Subject(unquote_plus(subject))
     if authorizer and not authorizer.check_authorization(user, Operation.Read, f"Subject:{subject}"):
         raise unauthorized()
 
--- a/src/schema_registry/routers/mode.py
+++ b/src/schema_registry/routers/mode.py
@@ -10,6 +10,7 @@ from karapace.typing import Subject
 from schema_registry.container import SchemaRegistryContainer
 from schema_registry.controller import KarapaceSchemaRegistryController
 from schema_registry.routers.errors import unauthorized
+from schema_registry.routers.raw_path_router import RawPathRoute
 from schema_registry.routers.requests import ModeResponse
 from schema_registry.user import get_current_user
 from typing import Annotated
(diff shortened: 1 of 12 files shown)


src/schema_registry/routers/raw_path_router.py
```diff
@@ -0,0 +1,39 @@
...
+
+Copyright (c) 2018 Sebastián Ramírez
+"""
+
+from fastapi import HTTPException  # noqa: E402
+from fastapi.routing import APIRoute  # noqa: E402
+from starlette.routing import Match  # noqa: E402
+from starlette.types import Scope  # noqa: E402
+
+import re  # noqa: E402
```
jjaakola-aiven, line 17: The `# noqa` comments are due to double header. Ruff gets confused.

nosahama: Maybe we add to the flake8 config to ignore specific codes per-file, i.e. https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores

src/schema_registry/routers/raw_path_router.py
```diff
@@ -0,0 +1,44 @@
+"""
+Copyright (c) 2025 Aiven Ltd
+See LICENSE for details
+"""
+
+"""
+The MIT License (MIT)
+
+Copyright (c) 2018 Sebastián Ramírez
```
nosahama, line 9: Sebyyy 😄

src/schema_registry/routers/subjects.py
```diff
@@ -56,6 +57,7 @@ async def subjects_subject_post(
     authorizer: AuthenticatorAndAuthorizer = Depends(Provide[SchemaRegistryContainer.karapace_container.authorizer]),
     controller: KarapaceSchemaRegistryController = Depends(Provide[SchemaRegistryContainer.schema_registry_controller]),
 ) -> SchemaResponse:
+    subject = Subject(unquote_plus(subject))
```
nosahama, line 60: **[nit-pick | non-blocking]**
right now, we have - `Subject = NewType("Subject", str)`, at a later time, we can change to a class so we can intercept the `__init__()` and do the `unquote_plus()` in there.

jjaakola-aiven: Or in the APIRouter, the `Subject` should be changed to actual Pydantic model.

src/schema_registry/routers/raw_path_router.py
jjaakola-aiven: No, this needs to be here.

src/schema_registry/routers/mode.py
```diff
@@ -42,6 +45,11 @@ async def mode_get_subject(
     authorizer: AuthenticatorAndAuthorizer = Depends(Provide[SchemaRegistryContainer.karapace_container.authorizer]),
     controller: KarapaceSchemaRegistryController = Depends(Provide[SchemaRegistryContainer.schema_registry_controller]),
 ) -> ModeResponse:
+    import logging
```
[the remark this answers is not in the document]
jjaakola-aiven: Done.

src/schema_registry/routers/raw_path_router.py
jjaakola-aiven: As long as the change is this small I am ok with this, I would not bother to add per file configs.

PR #1030, opened 2025-01-24
module: single karapace package

nosahama:
We move modules around to form a single karapace package


src/karapace/core/content_type.py
jjaakola-aiven, line 1: This has `fastapi` import. Move to `api`.

jjaakola-aiven (commented):
The `ForwardClient` also has `fastapi` import, but is a bit trickier to move. Can be left as is and solve this later.

src/karapace/core/content_type.py
nosahama: I've made both changes.

davide-armand (approved):
LGTM

PR #1031, opened 2025-01-24
rest-proxy: allow reading messages with avro-value and string-key

jclarysse:
It is not uncommon that Kafka users produce messages with String-key and Avro-value. In this case Karapace REST typically fails to read the message since it only supports one format for deserializing both the key and the value. 

For more convenience, I suggest to fallback to `binary` deserializer for the key (only) when reading it using `avro` fails.

--

For the smoke and integration tests we actually don't care about the value. We just assume that it exists in Avro-format, important is that we produce a record with Base64 encoded String-key:
```
$ echo "testkey" | base64
dGVzdGtleQo=
$ curl -X POST -H "Content-Type: application/vnd.kafka.v2+json" \
  -d '{"records": [{"key": "dGVzdGtleQ=="}]}' \
  http://localhost:8082/topics/my_topic
{"key_schema_id":null,"offsets":[{"offset":0,"partition":0}],"value_schema_id":null}
```
Now, we want to consume the record by knowing that the value is in Avro format. So we create the consumer accordingly, and subscribe to the topic:
```
$ curl -X POST -H "Content-Type: application/vnd.kafka.v2+json" -H "Accept: application/vnd.kafka.v2+json" \
  --data '{"name": "my_consumer", "format": "avro", "auto.offset.reset": "earliest"}' \
  http://localhost:8082/consumers/avro_consumers
$ curl -X POST -H "Content-Type: application/vnd.kafka.v2+json" --data '{"topics":["my_topic"]}' \
  http://localhost:8082/consumers/avro_consumers/instances/my_consumer/subscription
```

Reading the record BEFORE this PR change:
```
$ curl -X GET -H "Accept: application/vnd.kafka.avro.v2+json" \
  http://localhost:8082/consumers/avro_consumers/instances/my_consumer/records\?timeout\=1000
{"error_code":422,"message":"key deserialization error for format avro: unpack requires a buffer of 5 bytes"}
```

Reading the record AFTER this PR change:
```
$ curl -X GET -H "Accept: application/vnd.kafka.avro.v2+json" \
  http://localhost:8081/consumers/avro_consumers/instances/my_consumer/records\?timeout\=1000
[{"key":"dGVzdGtleQ==","offset":0,"partition":0,"timestamp":1738667083672,"topic":"test-topic","value":null}]
```

I also tested with a record with a String-key and a non-empty Avro-value that was produced in Java.


jclarysse:
@muralibasani conflicts and failing test are now fixed.

src/karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -562,11 +562,15 @@ async def fetch(self, internal_name: tuple[str, str], content_type: str, formats
                 try:
                     key = await self.deserialize(msg.key(), request_format) if msg.key() else None
                 except DeserializationError as e:
-                    KarapaceBase.unprocessable_entity(
-                        message=f"key deserialization error for format {request_format}: {e}",
-                        sub_code=RESTErrorCodes.HTTP_UNPROCESSABLE_ENTITY.value,
-                        content_type=content_type,
-                    )
+                    if request_format == "avro":
+                        LOG.warning("Cannot process non-empty key using avro deserializer, falling back to binary.")
```
muralibasani, line 566: What do you think about capturing this warning the test too ?

muralibasani (approved):
Looks good, thanks for the changes.

PR #1033, opened 2025-01-30
chore: run schema registry from karapace module

jjaakola-aiven:

References: #xxxxx

# Why this way


davide-armand (approved):
LGTM

PR #1035, opened 2025-02-03
feat: use otel in core metrics

jjaakola-aiven:

References: #xxxxx

# Why this way


nosahama (approved):
I like the implementation, really nice, also shows how the DI containers simplify things. It's nice that you added dedicated containers. 
**[nit]**
- since we now have 2 containers in the core package, maybe we want to further group the container modules (`auth_container` and `metrics_container`) under a single package.
- maybe we need to document somewhere that the config tags are added to metric labels and issue tracking.
- would be nice to see the actual calls in the tests for the `stats` client.
- there were also some commented code.

PR #1040, opened 2025-02-12
otel: split exporters

nosahama:

We would like to have an OTel exporter per component, i.e. what if we want to export the traces to the console, but export the metrics to an OTel endpoint, right now that will not be possible as all OTel components share the same exporter, this PR splits the exporters.

- we also squeeze in a fix for the `grafana` dashboards
- moved the tracer component to `core`


nosahama:
closed/reopened to trigger the CI.

PR #1044, opened 2025-02-20
Support mTLS

nosahama:
We add support for mTLS to the service.

- we remove the pytest skip set for the TLS integration test in `test_client.py`.
- we add integration test for mTLS forwarding.
- `forward_client` is updated to use mTLS.
- `healthcheck` is also updated to use mTLS.
- certificates are generated using `mkcert`. 
- `mkcert` is added to the `karapace_cli` docker file stage.
- the CAROOT is installed.
- a make target `generate-sr-https-certs` is added to automate the certs generation process.
- a make target `curl-sr-https` is added to aid DX.
- certificates are for testing & development purposes only and thus are added to version control.
- config is added to set the server ca certs, docs are also updated to reflect this.
- certificates are mounted as docker volumes.
- container smoke tests are now run in docker.
- we add a replica schema registry docker-compose service (`karapace-schema-registry-follower`) so we see the master election behaviour.

## Caveats
- mTLS is not added to the `rest-proxy`:
  - this should now be straightforward to add when needed, the rest proxy docker container host should be added to the list of domains while creating the certificates and then the `rapu.py` components handling the entrypoint should be updated with the certificate locations, tests, any redirects, etc should be revised.

## Disabling mTLS
- unset env vars `KARAPACE_SERVER_TLS_CAFILE & KARAPACE_SERVER_TLS_CERTFILE & KARAPACE_SERVER_TLS_KEYFILE`
- set `KARAPACE_ADVERTISED_PROTOCOL` to `http`
- on `rest-proxy`, set `KARAPACE_REGISTRY_SCHEME` to `http`


tests/integration/conftest.py
```diff
@@ -614,8 +656,6 @@ async def fixture_registry_async_client_tls(
     registry_https_endpoint: str,
     server_ca: str,
 ) -> AsyncIterator[Client]:
-    pytest.skip("Test certification is not properly set")
```
nosahama, line 617: 🚀

website/source/install.rst
```diff
@@ -18,9 +18,18 @@ An example setup including configuration and Kafka connection is available as co
 
 Then you should be able to reach two sets of endpoints:
 
-* Karapace schema registry on http://localhost:8081
+* Karapace schema registry master on http://localhost:8081
```
muralibasani, line 21: Do you mean https here for intercommunication between master/slaves ?

nosahama: > Do you mean https here for intercommunication between master/slaves ?

you mean master/follower 😅 😉 ?
Yes exactly that, with mTLS support, we need now to use the proper scheme on the URL.

nosahama: I've been thinking to also use the [HTTPSRedirect](https://fastapi.tiangolo.com/advanced/middleware/#httpsredirectmiddleware) middleware from FastAPI to redirect all `http` requests. 
This might be nice to have, but we need to verify the behaviour of the redirect as we already have some request forwarding in place.

muralibasani:
```suggestion
* Karapace schema registry master on https://localhost:8081
```

Oh, my question was about protocol. Should this be https ?

nosahama: That is left as is, because there is still the option to have the service running without mTLS, thus then the scheme/protocol will be `http`, so to help, I added the make target `curl-sr-https` which is used to interface with the API on mTLS or noTLS and I then added an instruction for this in the docs.

muralibasani (approved):
LGTM.

PR #1045, opened 2025-02-24
Performance test improvements - EC-740

muralibasani:
Currently performance tests have a few potential issues before running.

Script does not run, due to locust file renaming

issues with LOCUST_GUI, LOCUST_FILE 

remove redundant var TOPICS

Currently performance tests have a few potential issues before running.

References: #xxxxx

# Why this way


nosahama (approved):
Thanks for adding the screenshot, I was also thinking to raise the point of if we still wanted to continue using locust, i.e. maybe consider jMeter. I also

PR #1052, opened 2025-03-05
Updating session timeout tests

muralibasani:

There is a test which uses producer instead of consumer to test session.timeout.ms (consumer config).
This PR updates it

References: #xxxxx

# Why this way

### DIFF
--- a/tests/integration/backup/test_session_timeout.py
+++ b/tests/integration/backup/test_session_timeout.py
@@ -6,18 +6,22 @@ See LICENSE for details
 from pathlib import Path
 
 import pytest
-from aiokafka.errors import NoBrokersAvailable
+from aiokafka.errors import InvalidSessionTimeoutError
+from confluent_kafka import TopicPartition
 from confluent_kafka.admin import NewTopic
 
-from karapace.backup.api import BackupVersion, create_backup
 from karapace.core.config import Config
 from karapace.core.kafka.admin import KafkaAdminClient
-from karapace.core.kafka_utils import kafka_producer_from_config
+
+from src.karapace.backup.api import _consume_records
+from src.karapace.backup.poll_timeout import PollTimeout
+from src.karapace.core.kafka_utils import kafka_producer_from_config, kafka_consumer_from_config
 from tests.integration.conftest import create_kafka_server
 from tests.integration.utils.config import KafkaDescription
 from tests.integration.utils.kafka_server import KafkaServers
 
 SESSION_TIMEOUT_MS = 65000
+INVALID_SESSION_TIMEOUT_MS = 5000
 GROUP_MIN_SESSION_TIMEOUT_MS = 60000
 GROUP_MAX_SESSION_TIMEOUT_MS = 70000
 
@@ -44,13 +48,13 @@ def fixture_kafka_server(
     )
 
 
-def test_producer_with_custom_kafka_properties_does_not_fail(
+def test_consumer_with_custom_kafka_properties_does_not_fail(
     kafka_server_session_timeout: KafkaServers,
     new_topic: NewTopic,
     tmp_path: Path,
 ) -> None:
     """
-    This test checks wether the custom properties are accepted by kafka.
+    This test checks weather the custom properties are accepted by kafka.
     We know by the implementation of the consumer startup code that if
     `group.session.min.timeout.ms` > `session.timeout.ms` the consumer
     will raise an exception during the startup.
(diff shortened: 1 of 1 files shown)


tests/integration/backup/test_session_timeout.py
```diff
@@ -88,27 +80,54 @@ def test_producer_with_custom_kafka_properties_does_not_fail(
...
+            key=b"foo",
+            value=b"bar",
+            partition=0,
+            headers=[
+                ("some-header", b"some header value"),
+                ("other-header", b"some other header value"),
+            ],
+            timestamp=1683474657,
+        )
+    producer.flush()
```
keejon, line 115: touching producer outside of context. Also i think the context manager is already flushing

tests/integration/backup/test_session_timeout.py
```diff
@@ -88,27 +80,54 @@ def test_producer_with_custom_kafka_properties_does_not_fail(
...
+    # configure session timeout less than min session time
+    config.session_timeout_ms = INVALID_SESSION_TIMEOUT_MS
+
+    print("Starting consumer with misconfigured session timeout...")
+    with kafka_consumer_from_config(config, new_topic.topic) as consumer:
+        consumer.subscribe([new_topic.topic])
+        consumer.poll(timeout=1000)
+
+    # without performing the backup the exception isn't raised.
+    with pytest.raises(InvalidSessionTimeoutError):
```
keejon, line 126: Don't really get this. Why is the error only raised on backup? Why are we creating the consumer in this test at all then? Same for the other test

tests/integration/backup/test_session_timeout.py
```diff
@@ -59,24 +62,13 @@ def test_producer_with_custom_kafka_properties_does_not_fail(
     """
     config = Config()
     config.bootstrap_uri = kafka_server_session_timeout.bootstrap_servers[0]
-    config.session_timeout_ms = SESSION_TIMEOUT_MS
+    config.session_timeout_ms = 65000
```
keejon, line 65: why don't we use `SESSION_TIMEOUT_MS` here?

tests/integration/backup/test_session_timeout.py
muralibasani: Yes, flush is not required here, removed.

tests/integration/backup/test_session_timeout.py
muralibasani: Removed the consumer block, and after looking at backup def, it does raise exception if any issues in consumption.

tests/integration/backup/test_session_timeout.py
muralibasani: Ah yes, updated.

tests/integration/backup/test_session_timeout.py
keejon: Do we know at which point the consumer will actually throw the error?

muralibasani: I guess it is here https://github.com/Aiven-Open/karapace/blob/62095264a86930baf2f3e749ff1431ece8e8e608/src/karapace/backup/api.py#L559

keejon: but thats the thing, we were also creating a consumer in the test but that did not fail right

muralibasani: Updated consumer and while starting to read the records it failed.

tests/integration/backup/test_session_timeout.py
```diff
@@ -6,18 +6,22 @@
...
+from aiokafka.errors import InvalidSessionTimeoutError
+from confluent_kafka import TopicPartition
 from confluent_kafka.admin import NewTopic
 
-from karapace.backup.api import BackupVersion, create_backup
 from karapace.core.config import Config
 from karapace.core.kafka.admin import KafkaAdminClient
-from karapace.core.kafka_utils import kafka_producer_from_config
+
+from src.karapace.backup.api import _consumer, _consume_records
```
keejon, line 16: You are using private stuff from backup package here. Would be better to just use kafka_consumer_from_config similar to kafka_producer_from_config

muralibasani: Oh yes, updated. Didn't notice it.

tests/integration/backup/test_session_timeout.py
```diff
@@ -44,13 +48,13 @@ def fixture_kafka_server(
...
 
-def test_producer_with_custom_kafka_properties_does_not_fail(
+def test_consumer_with_custom_kafka_properties_does_not_fail(
     kafka_server_session_timeout: KafkaServers,
     new_topic: NewTopic,
     tmp_path: Path,
 ) -> None:
     """
-    This test checks wether the custom properties are accepted by kafka.
+    This test checks weather the custom properties are accepted by kafka.
```
nosahama, line 57:
```suggestion
    This test checks whether the custom properties are accepted by kafka.
```

tests/integration/backup/test_session_timeout.py
```diff
@@ -64,51 +68,60 @@ def test_producer_with_custom_kafka_properties_does_not_fail(
...
 
-def test_producer_with_custom_kafka_properties_fail(
+def test_consumer_with_custom_kafka_properties_fail(
     kafka_server_session_timeout: KafkaServers,
     new_topic: NewTopic,
+    tmp_path: Path,
 ) -> None:
     """
-    This test checks wether the custom properties are accepted by kafka.
+    This test checks weather the custom properties are accepted by kafka.
```
nosahama, line 80:
```suggestion
    This test checks whether the custom properties are accepted by kafka.
```

tests/integration/backup/test_session_timeout.py
```diff
@@ -64,51 +68,60 @@ def test_producer_with_custom_kafka_properties_does_not_fail(
...
     We know by the implementation of the consumer startup code that if
     `group.session.min.timeout.ms` > `session.timeout.ms` the consumer
     will raise an exception during the startup.
     This test ensures that the `session.timeout.ms` can be injected in
-    the kafka config so that the exception isn't raised
+    the kafka config so that the exception is raised
     """
     admin_client = KafkaAdminClient(bootstrap_servers=kafka_server_session_timeout.bootstrap_servers)
```
nosahama, line 87: **[nit]**
the `admin_client` could be extracted into a fixture.

tests/integration/backup/test_session_timeout.py
```diff
@@ -64,51 +68,60 @@ def test_producer_with_custom_kafka_properties_does_not_fail(
...
+    else:
+        consume_messages(config, new_topic)
+
 
-    with pytest.raises(NoBrokersAvailable):
-        with kafka_producer_from_config(config) as producer:
-            _ = producer
+def consume_messages(config, new_topic):
+    with kafka_consumer_from_config(config, new_topic) as consumer:
+        (partition,) = consumer.partitions_for_topic(new_topic)
```
nosahama, line 121: **[nit]**
the `new_topic` fixture created just a single partition, so we can simplify the unpacking syntax as this is rather confusing.

PR #1054, opened 2025-03-06
Adding test for main

muralibasani:

Add tests to src/karapace/__main__.py, as currently there are no tests

References: #xxxxx

# Why this way

### DIFF
--- a/tests/unit/test_main.py
+++ b/tests/unit/test_main.py
@@ -5,13 +5,8 @@ Copyright (c) 2024 Aiven Ltd
 See LICENSE for details
 """
 
-import logging
 import runpy
 from unittest.mock import patch, MagicMock
-
-from _pytest.logging import LogCaptureFixture
-
-from karapace.core.container import KarapaceContainer
 
 from src.karapace import __main__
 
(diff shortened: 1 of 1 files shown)


tests/unit/test_main.py
```diff
@@ -0,0 +1,112 @@
...
+            __main__.karapace.api.routers.schemas,
+            __main__.karapace.api.routers.subjects,
+        ]
+        mock_schema_registry_container_instance.wire.assert_called_once_with(modules=schema_reg_container_expected_modules)
+
+        assert mock_create_app.call_count == 1
+
+
+def test_uvicorn_run(caplog: LogCaptureFixture, karapace_container: KarapaceContainer) -> None:
+    with caplog.at_level(logging.INFO, logger="karapace.api.factory"):
```
keejon, line 98: do we need the caplog here?

tests/unit/test_main.py
```diff
@@ -0,0 +1,112 @@
...
+            __main__.karapace.api.routers.mode,
+            __main__.karapace.api.routers.schemas,
+            __main__.karapace.api.routers.subjects,
+        ]
+        mock_schema_registry_container_instance.wire.assert_called_once_with(modules=schema_reg_container_expected_modules)
+
+        assert mock_create_app.call_count == 1
+
+
+def test_uvicorn_run(caplog: LogCaptureFixture, karapace_container: KarapaceContainer) -> None:
```
keejon, line 97: do we need `karapace_container` fixture?

tests/unit/test_main.py
muralibasani: Indeed, not required. Removed.

tests/unit/test_main.py
muralibasani: Removed from here too. Not required.

tests/unit/test_main.py
```diff
@@ -0,0 +1,104 @@
...
+Test config
+
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+
+import runpy
+from unittest.mock import patch, MagicMock
+
+from src.karapace import __main__
```
antonagestam, line 11: This is highly problematic. The `src` directory structure has the exact purpose of never executing code by local path in tests, but to enforce always executing what is _installed_. In other words, imports should never be prefixed with `src.*`.

antonagestam: See e.g. https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

tests/unit/test_main.py
```diff
@@ -0,0 +1,104 @@
...
+        patch("karapace.core.metrics_container.MetricsContainer") as mock_metrics_container,
+        patch("karapace.api.telemetry.container.TelemetryContainer") as mock_telemetry_container,
+        patch("karapace.api.container.SchemaRegistryContainer") as mock_schema_registry_container,
+        patch("uvicorn.run"),
+    ):
+        mock_app = MagicMock()
+        mock_create_app.return_value = mock_app
+
+        with patch("sys.argv", ["src.karapace.__main__.py"]):
+            runpy.run_path("src/karapace/__main__.py", run_name="__main__")
```
nosahama, line 28: **[curious?]**
https://docs.python.org/3/library/runpy.html

PR #1072, opened 2025-04-25
Upgrade h11 & httpcore

muralibasani:

Had to upgrade httpcore to 1.0.9 to upgrade h11 to 0.16.0

References: #xxxxx

# Why this way


roope-kar (approved):
lgtm

PR #1058, opened 2025-03-16
fix: Use installed imports in main test

antonagestam:
# About this change - What it does

References: #xxxxx

# Why this way

PR #1079, opened 2025-05-06
Replace src karapace path with module in main test py

muralibasani:

References: #xxxxx

# Why this way


nosahama (approved):

muralibasani:
https://github.com/Aiven-Open/karapace/pull/1058 fixed here in this pr.

PR #1100, opened 2025-06-26
docker: pin kafka versions in compose.yml

nosahama:
Using `:latest` started to cause issues as now this has the KRaft mode enforced. We will work on migrating to KRaft in another iteration, so for now we pin the docker kafka/zookeeper versions to ones without KRaft.


roope-kar (approved):
Tested locally and now works, thank you 👍🏼

PR #1105, opened 2025-07-07
Added kafka client id

markretallack:

This change adds in the Kafka client ID to the Kafka producer. It also adds it to the Kafka Admin. 

This is needed to allow the client id to be specified, allowing the difference instances to have different names etc...

This does not fix any issue. 

# Why this way

We needed to be able to specify the actual client id so that we can see the connection appearing from the logs. We also use Kafka rest with a non-standard Kafka, that uses the client ID to specific additional client specific handling.


muralibasani:
@markretallack Thanks for the pr. Will come back to it soon.

PR #1106, opened 2025-07-10
Adding docker configs for oidc, keycloak

muralibasani:

- Adds new keycloak image to docker compose
- Adds 4 configs to karapace configs

With these changes, we can get a token from keycloak and send it for schema registry requests on get or post calls.

References: #xxxxx

# Why this way


container/keycloak/compose.yml
```diff
@@ -0,0 +1,10 @@
+# docker-compose.keycloak.yml
+services:
```
nosahama: I do not think we need this file.

src/karapace/api/oidc/middleware.py
```diff
@@ -35,6 +35,8 @@ def __init__(self, app: FastAPI, config: Config) -> None:
                 raise ValueError(
                     "OIDC config error: 'issuer' and 'audience' must be set if 'jwks_endpoint_url' is provided."
                 )
+            log.info("OIDC middleware initialized — JWKS URL configured, bearer token validation enabled.")
+            log.debug("OIDC Params jwks_url=%s issuer=%s audience=%s", self.jwks_url, self.issuer, self.audience)
```
nosahama: Maybe revert this log line or unless none of those details would reveal sensitive information 👍

container/compose.yml
```diff
@@ -101,6 +101,10 @@ services:
       KARAPACE_SERVER_TLS_CAFILE: /opt/karapace/certs/ca/rootCA.pem
       KARAPACE_SERVER_TLS_CERTFILE: /opt/karapace/certs/cert.pem
       KARAPACE_SERVER_TLS_KEYFILE: /opt/karapace/certs/key.pem
+      KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
```
nosahama, line 104: Maybe we also want to add similar configuration to the `karapace-schema-registry-follower` service below, but can also wait to be done in another PR.

container/keycloak/compose.yml
muralibasani: Yes, this is not needed anymore. Will delete it.

src/karapace/api/oidc/middleware.py
muralibasani: So there is no sensitive info here. It is just the realm name and issuers. No token, secret etc are here. I think it is ok.

nosahama: If it is also a helpful log, I think we can have it as `INFO`, wdyt?

muralibasani: It is indeed helpful, I can change it to info. Good point.

nosahama (approved):

PR #1111, opened 2025-07-16
OIDC karapace authorization

muralibasani:
- Ability to authorize karapace schema registry requests based on oidc roles

New example configs in karapace config

```
    sasl_oauthbearer_jwks_endpoint_url: str = "http://localhost:8080/realms/klaw/protocol/openid-connect/certs"
    sasl_oauthbearer_expected_issuer: str = "http://localhost:8080/realms/klaw"
    sasl_oauthbearer_expected_audience: str = "account"
    sasl_oauthbearer_sub_claim_name: str | None = "sub"
    sasl_oauthbearer_authorization_enabled: bool = True
    sasl_oauthbearer_client_id: str = "klaw"
    sasl_oauthbearer_roles_claim_path: str = "resource_access.[client_id].roles"
    sasl_oauthbearer_method_roles: dict[str, list[str]] = {"GET": ["schema:read", "subject:read"],
                                                           "POST": [], "PUT": [], "DELETE": []}
```

Authorization for GET/POST/PUT/DELETE

### DIFF
--- a/src/karapace/api/middlewares/__init__.py
+++ b/src/karapace/api/middlewares/__init__.py
@@ -43,6 +43,7 @@ def setup_middlewares(app: FastAPI, config: Config) -> None:
         auth_header = request.headers.get("Authorization")
         if auth_header and auth_header.startswith("Bearer "):
             try:
+                # authentication
                 payload = oidc_middleware.validate_jwt(auth_header.split(" ", 1)[1])
                 request.state.user = payload
             except AuthenticationError:
@@ -50,6 +51,12 @@ def setup_middlewares(app: FastAPI, config: Config) -> None:
                     status_code=401,
                     content={"error": "Unauthorized", "reason": "Invalid token/payload"},
                 )
+
+            try:
+                # authorization
+                oidc_middleware.authorize_request(payload, request.method)
+            except HTTPException as e:
+                return JSONResponse({"error": "Authorization error", "reason": e.detail}, status_code=e.status_code)
 
         response = await call_next(request)
         response.headers["Content-Type"] = response_content_type
--- a/src/karapace/api/oidc/middleware.py
+++ b/src/karapace/api/oidc/middleware.py
@@ -3,9 +3,11 @@ Copyright (c) 2025 Aiven Ltd
 See LICENSE for details
 """
 
+from typing import Any
+
 import jwt
 import logging
-from fastapi import FastAPI
+from fastapi import FastAPI, HTTPException
 from jwt import PyJWKClient, InvalidTokenError
 from starlette.responses import JSONResponse
 from karapace.core.auth import AuthenticationError
@@ -25,6 +27,10 @@ class OIDCMiddleware:
         self.issuer = config.sasl_oauthbearer_expected_issuer
         self.audience = config.sasl_oauthbearer_expected_audience
         self.claim_name = config.sasl_oauthbearer_sub_claim_name
+        self.authorization_enabled = config.sasl_oauthbearer_authorization_enabled
+        self.client_id = config.sasl_oauthbearer_client_id
+        self.sasl_oauthbearer_method_roles: dict[str, list[str]] = config.sasl_oauthbearer_method_roles
+        self.sasl_oauthbearer_roles_claim_path = config.sasl_oauthbearer_roles_claim_path
 
         # Hardcoded default algorithms
         self.algorithms = ["RS256", "RS384", "RS512"]
@@ -43,6 +49,29 @@ class OIDCMiddleware:
             )
 
             self._jwks_client = PyJWKClient(self.jwks_url)
+
+            log.info("OIDC Authorization enabled: %s", self.authorization_enabled)
+            if self.authorization_enabled:
+                if self.client_id is None or self.sasl_oauthbearer_roles_claim_path is None:
+                    raise HTTPException(
+                        status_code=403,
+                        detail="Invalid configuration: client_id and roles_claim_path are required when authorization is enabled.",
+                    )
+
+                required_http_methods = set(self.sasl_oauthbearer_method_roles.keys())
+                # Validate required HTTP methods in method_roles
+                missing_methods = required_http_methods - self.sasl_oauthbearer_method_roles.keys()
+                if missing_methods:
+                    raise HTTPException(
+                        status_code=500,
+                        detail=f"Invalid configuration: method_roles is missing definitions for: {', '.join(sorted(missing_methods))}",
+                    )
+                log.info(
+                    "OIDC Authorization configured. — client_id: %s, method_roles: %s, roles_claim_path: %s",
+                    self.client_id,
+                    self.sasl_oauthbearer_method_roles,
+                    self.sasl_oauthbearer_roles_claim_path,
+                )
         else:
             self._jwks_client = None
             log.warning("OIDC middleware initialized without JWKS URL — token validation will be skipped.")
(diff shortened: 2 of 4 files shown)


src/karapace/core/config.py
```diff
@@ -108,6 +108,10 @@ class Config(BaseSettings):
     sasl_oauthbearer_expected_issuer: str | None = None
     sasl_oauthbearer_expected_audience: str | None = None
     sasl_oauthbearer_sub_claim_name: str | None = "sub"
+    sasl_oauthbearer_authorization_enabled: bool = False
```
nosahama, line 111: **[non-blocking]** 
**could be done in follow up PR**
I was thinking we can scope all the `sasl_oauthbearer_` configuration (which might also include some added earlier) into a model/class, similar to what was done for `KarapaceTelemetry`, then we can reference this whole class in the config as just an instantiation of itself and then all the values reside in the model, i.e. `sasl_oauthbearer: SASLOAuthBearer = SASLOAuthBearer()`. 

Mind you that this will cause a change to how we declare the environment variables.

src/karapace/api/oidc/middleware.py
```diff
@@ -25,6 +27,10 @@ def __init__(self, app: FastAPI, config: Config) -> None:
         self.issuer = config.sasl_oauthbearer_expected_issuer
         self.audience = config.sasl_oauthbearer_expected_audience
         self.claim_name = config.sasl_oauthbearer_sub_claim_name
+        self.authorization_enabled = config.sasl_oauthbearer_authorization_enabled
```
nosahama, line 30: **[non-blocking]**
**could be done in follow up PR**
If we manage to scope the config to a model, then we can just pass in only the model part of the config, i.e. `config: SASLOAuthBearer = config.sasl_oauthbearer`, this also would affect earlier definitions.

src/karapace/api/middlewares/__init__.py
```diff
@@ -43,13 +43,19 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
         auth_header = request.headers.get("Authorization")
         if auth_header and auth_header.startswith("Bearer "):
             try:
+                # authentication
                 payload = oidc_middleware.validate_jwt(auth_header.split(" ", 1)[1])
                 request.state.user = payload
+
+                # authorization
+                oidc_middleware.authorize_request(payload, request.method)
```
nosahama: This could go in a different `try/except` block scoped to the `HTTPException`, but it's still also fine as-is since it's just 2 statements and they ideally return fixed only one of the 2 exceptions.

src/karapace/api/oidc/middleware.py
```diff
@@ -43,6 +49,29 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
+
+            log.info("OIDC Authorization enabled: %s", self.authorization_enabled)
+            if self.authorization_enabled:
+                if self.client_id is None or self.sasl_oauthbearer_roles_claim_path is None:
+                    raise HTTPException(
+                        status_code=403,
+                        detail="Invalid configuration: client_id and roles_claim_path are required when authorization is enabled.",
+                    )
+
+                required_http_methods = {"GET", "POST", "PUT", "DELETE"}
```
nosahama: Maybe this should come from the keys of `self.sasl_oauthbearer_method_roles`, that way we are not forcing them to use auth on methods they haven't configured, unless we want to enforce auth on all these methods if OIDC is enabled.

src/karapace/api/oidc/middleware.py
```diff
@@ -43,6 +49,29 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
+                # Validate required HTTP methods in method_roles
+                missing_methods = required_http_methods - self.sasl_oauthbearer_method_roles.keys()
```
nosahama, line 63: nice to know that `dict_keys` supports set ops.

src/karapace/api/oidc/middleware.py
```diff
@@ -43,6 +49,29 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
+                if missing_methods:
+                    raise HTTPException(
+                        status_code=500,
```
nosahama, line 66: This is not really a `500` error, more like `4xx` since it's a user config error.

tests/unit/test_oidc.py
```diff
@@ -41,12 +47,20 @@ class DummyConfig:
         sasl_oauthbearer_expected_issuer="https://oidcprovider.com",
         sasl_oauthbearer_expected_audience="accounts-audience",
         sasl_oauthbearer_sub_claim_name="sub",
+        sasl_oauthbearer_authorization_enabled=False,
```
nosahama, line 51: I guess we still need to add unit tests for these. 
Also the `get_roles_from_claim_path()` and `authorize_request()`.

E2E tests would be nice, but can be a follow up.

src/karapace/api/oidc/middleware.py
```diff
@@ -71,6 +100,50 @@ def validate_jwt(self, token: str) -> dict:
...
+        request_method = request_method.upper()
+        allowed_roles = self.sasl_oauthbearer_method_roles.get(request_method, [])
+
+        # Dynamic role extraction based on configured path
+        if self.sasl_oauthbearer_roles_claim_path is not None and self.client_id is not None:
+            roles_claim_path = self.sasl_oauthbearer_roles_claim_path.replace("[client_id]", self.client_id)
+        else:
+            raise HTTPException(status_code=403, detail="Insufficient roles. Invalid roles_claim_path.")
+
+        log.info("------ payload %s", payload)
```
nosahama: Maybe we need to clean up such logs.

src/karapace/api/oidc/middleware.py
```diff
@@ -71,6 +100,50 @@ def validate_jwt(self, token: str) -> dict:
...
+                request_method,
+                user_roles,
+                allowed_roles,
+            )
+            raise HTTPException(status_code=403, detail="Insufficient roles")
+        log.info("------  Authorized ------ ")
+        return True
+
+    @staticmethod
+    def get_roles_from_claim_path(payload: dict[str, Any], path: str) -> list[str]:
```
nosahama, line 126: We can skip all this logic if we know the payload of the OIDC provider, I think we could also create models in code to reflect this, we can investigate the data the providers return and work with that, while this is more accepting, it is not certain that the providers will send a particular model for the roles and also what level in the payload is the object.

nosahama: We can gradually roll out supported providers for instance.

src/karapace/api/oidc/middleware.py
muralibasani: So this particular one is a config incorrectly defined on server. On server startup, this config is loaded and failed. So I thought its an internal server error. I looked at other 4xx err which are mostly for requests, if they are bad/conflict...etc.

src/karapace/api/middlewares/__init__.py
muralibasani: done.

src/karapace/api/oidc/middleware.py
muralibasani: Done. Good one.

src/karapace/api/oidc/middleware.py
nosahama: I still think it's a user error as they caused the anomaly.

muralibasani: Sure, let's come back to this again. 
But do you suggest any particular 4xx code

nosahama: `400` maybe as it's a bad user configuration.

PR #1112, opened 2025-07-21
Docker configs authorization

muralibasani:

References: #xxxxx

# Why this way

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -53,7 +53,7 @@ services:
       KAFKA_ZOOKEEPER_CONNECT: "zookeeper:2181"
 
   karapace-schema-registry:
-    image: ghcr.io/aiven-open/karapace:develop
+    image: karapacelocal:latest
     build:
       context: ..
       dockerfile: container/Dockerfile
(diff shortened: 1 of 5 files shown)


src/karapace/api/middlewares/__init__.py
```diff
@@ -46,6 +46,7 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
                 # authentication
                 payload = oidc_middleware.validate_jwt(auth_header.split(" ", 1)[1])
                 request.state.user = payload
+                log.debug("Authenticated")
```
nosahama, line 49: should we clean this?

src/karapace/api/oidc/middleware.py
```diff
@@ -120,6 +123,7 @@ def authorize_request(self, payload: dict, request_method: str) -> bool:
                 allowed_roles,
             )
             raise HTTPException(status_code=403, detail="Insufficient roles")
+        log.debug("Authorized")
```
nosahama, line 126: this also?

container/compose.yml
```diff
@@ -103,8 +103,12 @@ services:
       KARAPACE_SERVER_TLS_KEYFILE: /opt/karapace/certs/key.pem
       KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_ISSUER: http://keycloak:8080/realms/karapace
-      KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "account"
+      KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "karapace-audience"
       KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
+      KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: false
```
nosahama: I am in support of enabling this by default, did you run into any issues, maybe e2e tests might fail? 🤔

src/karapace/api/oidc/middleware.py
muralibasani: Is it ok if we remove these after aiven core integration, in the final steps ? These help in development.

container/compose.yml
```diff
@@ -157,8 +161,12 @@ services:
...
+      KARAPACE_SASL_OAUTHBEARER_CLIENT_ID: "karapace-client"
+      KARAPACE_SASL_OAUTHBEARER_ROLES_CLAIM_PATH: "resource_access.karapace-client.roles"
+      KARAPACE_SASL_OAUTHBEARER_METHOD_ROLES: '{"GET": ["schema:read", "subject:read"], "POST": ["schema:write", "subject:write"], "PUT": [], "DELETE": ["schema:delete", "subject:delete"]}'
```
nosahama: I see `PUT` has no roles, if we make a request using this method and imagine the method is not allowed, will we return `405` or `401`?

container/compose.yml
muralibasani: We can enable it. I shall add all post, delete and put as well in the allowed list and try in this pr.

container/compose.yml
muralibasani: We should see 401 error. I shall add PUT as well in the method list. And later we can add e2e tests

muralibasani: Also, we cannot update a schema as such. We can only add or delete versions on a subject.
We use PUT only to update configs(global or subject) like compatibility. Let me chk.

src/karapace/api/oidc/middleware.py
nosahama: Sure, we can also leave it, on core, the log level is higher than debug, maybe INFO IIRC 👍 
I just wasn't sure if you needed or left it there.

muralibasani: Ah yeah, will change these to info for now, as I need them.

nosahama: I think debug is fine as you only need them during development.

nosahama (approved):

PR #1113, opened 2025-07-21
Update to model

muralibasani:
Currently oidc configs are defined flatly in BaseModel. They are now moved under OidcKarapace in this pr.

### DIFF
--- a/src/karapace/api/oidc/middleware.py
+++ b/src/karapace/api/oidc/middleware.py
@@ -23,14 +23,14 @@ class OIDCMiddleware:
         self.config = config
 
         self._jwks_client: PyJWKClient | None
-        self.jwks_url = config.sasl_oauthbearer_jwks_endpoint_url
-        self.issuer = config.sasl_oauthbearer_expected_issuer
-        self.audience = config.sasl_oauthbearer_expected_audience
-        self.claim_name = config.sasl_oauthbearer_sub_claim_name
-        self.authorization_enabled = config.sasl_oauthbearer_authorization_enabled
-        self.client_id = config.sasl_oauthbearer_client_id
-        self.sasl_oauthbearer_method_roles: dict[str, list[str]] = config.sasl_oauthbearer_method_roles
-        self.sasl_oauthbearer_roles_claim_path = config.sasl_oauthbearer_roles_claim_path
+        self.jwks_url = config.oidc_karapace.sasl_oauthbearer_jwks_endpoint_url
+        self.issuer = config.oidc_karapace.sasl_oauthbearer_expected_issuer
+        self.audience = config.oidc_karapace.sasl_oauthbearer_expected_audience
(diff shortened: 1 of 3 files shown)


src/karapace/api/oidc/middleware.py
```diff
@@ -23,14 +23,14 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
         self._jwks_client: PyJWKClient | None
-        self.jwks_url = config.sasl_oauthbearer_jwks_endpoint_url
-        self.issuer = config.sasl_oauthbearer_expected_issuer
-        self.audience = config.sasl_oauthbearer_expected_audience
-        self.claim_name = config.sasl_oauthbearer_sub_claim_name
-        self.authorization_enabled = config.sasl_oauthbearer_authorization_enabled
-        self.client_id = config.sasl_oauthbearer_client_id
-        self.sasl_oauthbearer_method_roles: dict[str, list[str]] = config.sasl_oauthbearer_method_roles
-        self.sasl_oauthbearer_roles_claim_path = config.sasl_oauthbearer_roles_claim_path
+        self.jwks_url = config.oidc_karapace.sasl_oauthbearer_jwks_endpoint_url
```
nosahama, line 26: **[non-blocking]**
You actually do not need the whole config object, you could just pass in already, `config: OIDCKarapace`.

src/karapace/core/config.py
```diff
@@ -55,6 +55,17 @@ class KarapaceTelemetry(BaseModel):
     resource_telemetry_sdk_version: str = OTEL_VERSION
 
 
+class OidcKarapace(BaseModel):
```
nosahama, line 58: **[non-blocking]**
Maybe naming can be adjusted to `KarapaceOIDC` to fit the current way we do it, but then you'd need to replace where you've used it.

src/karapace/core/config.py
```diff
@@ -140,6 +143,7 @@ class Config(BaseSettings):
     sentry: Mapping[str, object] | None = None
     tags: KarapaceTags = KarapaceTags()
     telemetry: KarapaceTelemetry = KarapaceTelemetry()
+    oidc_karapace: OidcKarapace = OidcKarapace()
```
nosahama, line 146: This could just be `oidc` as it is already within scope of the Config.

tests/unit/test_oidc.py
```diff
@@ -30,61 +30,60 @@ def _assert_unauthorized_http_response(http_response: HTTPResponse) -> None:
 
 @dataclass
 class DummyConfig:
```
nosahama, line 32: You do not need this anymore, you could directly use the new object.

muralibasani:
@nosahama I only have one concern this with pr. If kafka connects to karapace, the parent config for sasl like kafka.sasl_.... will not work .. wdyt ?

PR #1116, opened 2025-07-23
docker: use kafka-3.9

nosahama:
Adds kafka-3.9 to docker compose:
- drop zookeeper for kafka running in KRaft mode.
- adds meaningful `hostname` and `container_name` to services rather than the auto-generated naming.
- use specific karapace network and bridge driver so we can connect to the compose resources via services maybe in another compose setup by referencing the same network.

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -1,59 +1,50 @@
 services:
-  zookeeper:
-    image: confluentinc/cp-zookeeper:7.2.2
-    ports:
-      - 2181:2181
-    environment:
-      ZOOKEEPER_CLIENT_PORT: 2181
-      ZOOKEEPER_TICK_TIME: 2000
 
   kafka:
-    image: confluentinc/cp-kafka:7.2.2
-    depends_on:
-      - zookeeper
+    networks:
+      - karapace
+    image: apache/kafka:3.9.0
+    hostname: kafka
+    container_name: kafka
     ports:
-      - 9101:9101  # JMX
-      - 9092:9092  # Kafka
+      - "9092:9092"
+      - "9101:9101"  # JMX
     environment:
-      # Listeners:
-      # PLAINTEXT_HOST -> Expose kafka to the host network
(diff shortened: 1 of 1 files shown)


container/compose.yml
```diff
@@ -1,59 +1,50 @@
...
-      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
-      KAFKA_SOCKET_RECEIVE_BUFFER_BYTES: 102400
-      KAFKA_SOCKET_REQUEST_MAX_BYTES: 104857600
-      KAFKA_SOCKET_SEND_BUFFER_BYTES: 102400
-      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
-      KAFKA_TRANSACTION_STATE_LOG_NUM_PARTITIONS: 16
-      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
-      KAFKA_ZOOKEEPER_CONNECTION_TIMEOUT_MS: 6000
-      KAFKA_ZOOKEEPER_CONNECT: "zookeeper:2181"
+      KAFKA_CONFLUENT_SCHEMA_REGISTRY_URL: http://karapace-schema-registry:8081
```
muralibasani, line 40: As we are referring to karapace, can we replace CONFLUENT with AIVEN ?

muralibasani: This impacts existing clients, but this change has to happen sometime. Similar change was already done in Aiven's kafka connectors

nosahama: Yes very true, I am a bit not interested in rerunning the CI just to change that, but it makes sense and I will need to check what environment variable Kafka expects for the schema registry.

nosahama: I was checking - https://doc.nais.io/persistence/kafka/reference/environment-variables and they mention Aiven which is good, I will take a look, but they also mention to export `KAFKA_SCHEMA_REGISTRY`, hopefully this isn't specific to their use case, I will still check around.

PR #1119, opened 2025-07-30
Update rapu, parse header with email message

muralibasani:

References: #xxxxx

# Why this way


src/karapace/rapu.py
```diff
@@ -285,7 +297,10 @@ async def _handle_request(
                 if not body:
                     raise HTTPResponse(body="Missing request JSON body", status=HTTPStatus.BAD_REQUEST)
                 try:
-                    _, options = cgi.parse_header(rapu_request.get_header("Content-Type"))
+                    content_msg = Message()
+                    content_msg["Content-Type"] = rapu_request.get_header("Content-Type", "")
+                    options = dict(content_msg.get_params()[1:])
```
nosahama: I see below, we only get the `charset` from the `options`, i.e. on the line below: `charset = options.get("charset", "utf-8")`, so maybe no need to slice and pack into a dict, I think we can use `charset = content_msg.get_param('charset')`

nosahama: See: [PEP-594 #cgi](https://peps.python.org/pep-0594/#cgi)

src/karapace/rapu.py
```diff
@@ -202,10 +202,18 @@ def check_rest_headers(self, request: HTTPRequest) -> dict:
         default_content = "application/vnd.kafka.json.v2+json"
         default_accept = "*/*"
         result = {"content_type": default_content}
-        content_matcher = REST_CONTENT_TYPE_RE.search(
-            cgi.parse_header(request.get_header("Content-Type", default_content))[0]
-        )
-        accept_matcher = REST_ACCEPT_RE.search(cgi.parse_header(request.get_header("Accept", default_accept))[0])
+
+        # Handle Content-Type
+        msg = Message()
```
nosahama, line 206: Would it make sense to use a single `Message()` class for all the parsing? 
Since we mostly only update the `Content-Type`, i.e. `msg["Content-Type"] = xyz`, but I am unsure how it'd affect the `msg.get_params()` call, ideally should be fine, but not sure.

nosahama (approved):
LGTM 👍 
Left some comments, but all non-blocking, hopefully we have tests for these lines.

src/karapace/rapu.py
muralibasani: Message class is mutable, and it might cause side effects during concurrent requests. I'd better leave it inside the function.

src/karapace/rapu.py
muralibasani: Good one. Updated.

PR #1127, opened 2025-08-08
rapu: drop usage of deprecated `cgi.parse_header`

nosahama:
Initial work was done in #1119, **we add logic to**: 
- reuse a single function for parsing.
- parse the extracted params, i.e. `charset` with `email.utils.collapse_rfc2231_value(<rawparam>)` because the param value can either be a string, or a 3-tuple if the parameter was RFC-2231 encoded, see [this comment](https://github.com/python/cpython/blob/362692852f13cdd1d33cc7ed35c0cbac7af1a785/Lib/email/message.py#L722) in the cpython source code.

### DIFF
--- a/src/karapace/rapu.py
+++ b/src/karapace/rapu.py
@@ -20,7 +20,7 @@ import aiohttp
 import aiohttp.web
 import aiohttp.web_exceptions
 import asyncio
-import cgi
+import email
 import hashlib
 import logging
 import re
@@ -57,6 +57,14 @@ REST_ACCEPT_RE = re.compile(
 def is_success(http_status: HTTPStatus) -> bool:
     """True if response has a 2xx status_code"""
     return http_status.value >= 200 and http_status.value < 300
+
+
+def parse_header(header_value: str) -> tuple[str, dict[str, str]]:
+    msg = email.message.Message()
+    msg["Content-Type"] = header_value
+    main_value = msg.get_content_type()
+    params = {k: v for k, v in msg.get_params()[1:]}
+    return main_value, params
 
 
 class HTTPRequest:
@@ -202,10 +210,8 @@ class RestApp:
         default_content = "application/vnd.kafka.json.v2+json"
         default_accept = "*/*"
         result = {"content_type": default_content}
-        content_matcher = REST_CONTENT_TYPE_RE.search(
-            cgi.parse_header(request.get_header("Content-Type", default_content))[0]
-        )
-        accept_matcher = REST_ACCEPT_RE.search(cgi.parse_header(request.get_header("Accept", default_accept))[0])
+        content_matcher = REST_CONTENT_TYPE_RE.search(parse_header(request.get_header("Content-Type", default_content))[0])
+        accept_matcher = REST_ACCEPT_RE.search(request.get_header("Accept", default_accept))
         if method in {"POST", "PUT"}:
             if not content_matcher:
                 http_error(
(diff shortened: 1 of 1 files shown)


src/karapace/rapu.py
```diff
@@ -202,10 +210,8 @@ def check_rest_headers(self, request: HTTPRequest) -> dict:
         default_content = "application/vnd.kafka.json.v2+json"
         default_accept = "*/*"
         result = {"content_type": default_content}
-        content_matcher = REST_CONTENT_TYPE_RE.search(
-            cgi.parse_header(request.get_header("Content-Type", default_content))[0]
-        )
-        accept_matcher = REST_ACCEPT_RE.search(cgi.parse_header(request.get_header("Accept", default_accept))[0])
+        content_matcher = REST_CONTENT_TYPE_RE.search(parse_header(request.get_header("Content-Type", default_content))[0])
+        accept_matcher = REST_ACCEPT_RE.search(request.get_header("Accept", default_accept))
```
jclarysse, line 214: Why does Content-Type header need to be parsed, while Accept header doesn't?

src/karapace/rapu.py
```diff
@@ -59,6 +59,14 @@ def is_success(http_status: HTTPStatus) -> bool:
     return http_status.value >= 200 and http_status.value < 300
 
 
+def parse_header(header_value: str) -> tuple[str, dict[str, str]]:
+    msg = email.message.Message()
+    msg["Content-Type"] = header_value
```
jclarysse, line 64: "Content-Type" is a bit hidden here, perhaps change method name to parse_header_content_type, just in case someone would be tempted to reuse it later?

src/karapace/rapu.py
nosahama: Yes makes sense, initially I tried to parse `Accept` and the tests were failing because `Accept` headers can contain  multiple values, and also `email.message.Message` is designed to work with single value `Content-Type`, so even if we try to use it as-is, it will only parse and return the first value and this drops the others which isn't also intended, we can test with the script you provided as indeed before we did `cgi.parse_header(request.get_header("Accept", default_accept)`.
I will try to see if the `Message` supports specifying the source of the header and try again.

src/karapace/rapu.py
nosahama: Yes makes sense, how fixed are you on updating the name? This is scoped only to this module, `rapu.py` will also soon be gone and I might not want to rerun the CI.

src/karapace/rapu.py
nosahama: After some tests with a script to parse Accept:

```
myHeaders: dict[str, str] = {
  "Accept": "application/json, text/json, application/octet-stream",
}
accept: str = get_header("Accept", 'header-default')
print(f"- get Accept as accept: {accept}")
print(f"- parse accept using cgi module: {cgi.parse_header(accept)[0]}")
print(f"- parse accept using email module: {parse_header(accept)[0]}")
```

Result:
```
- get Accept as accept: application/json, text/json, application/octet-stream
- parse accept using cgi module: application/json, text/json, application/octet-stream
- parse accept using email module: text/plain
```

The `cgi.parse_header` just returned the initial value, which is no different in just passing directly the header to `REST_ACCEPT_RE.search`. 

So I will leave as-is.

jclarysse: gotcha!

src/karapace/rapu.py
jclarysse: fine!

PR #1133, opened 2025-08-21
4.1.2 rapu: drop usage of deprecated cgi.parse_header

muralibasani:

Currently rapu.py has a few references to cgi.parse_header which is not efficient.
It is replaced with email.message.Message

More info : https://peps.python.org/pep-0594/#cgi

### DIFF
--- a/src/karapace/rapu.py
+++ b/src/karapace/rapu.py
@@ -18,7 +18,7 @@ import aiohttp
 import aiohttp.web
 import aiohttp.web_exceptions
 import asyncio
-import cgi  # pylint: disable=deprecated-module
+import email
 import hashlib
 import logging
 import re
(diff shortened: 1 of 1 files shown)


src/karapace/rapu.py
```diff
@@ -57,6 +57,14 @@ def is_success(http_status: HTTPStatus) -> bool:
     return http_status.value >= 200 and http_status.value < 300
 
 
+def parse_header(header_value: str) -> tuple[str, dict[str, str]]:
+    msg = email.message.Message()
+    msg["Content-Type"] = header_value
+    main_value = msg.get_content_type()
+    params = dict(msg.get_params()[1:])
```
muralibasani, line 64: pylint : had to use dict instead compared to main branch changes.

.github/workflows/container-smoke-test.yml
```diff
@@ -32,8 +32,14 @@ jobs:
       - name: Run container
         run: docker compose --file=container/compose.yml up --build --wait --detach
 
+      - name: Check container status
+        run: docker ps -a
+
+      - name: Show container logs (Kafka + others)
```
nosahama, line 38: This should come after the smoke test runs, so you do not need to repeat the commands directly within those jobs.

muralibasani: We can update this in another pr. We were trying to get some debug logs

PR #1135, opened 2025-09-01
Fix producer acks

muralibasani:
If producer_acks is set to all, sometimes it fails in integer parsing.

_producer_acks :   Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='all', input_type=str]_

It is now being fixed in config with this decorator.

### DIFF
--- a/src/karapace/core/config.py
+++ b/src/karapace/core/config.py
@@ -9,11 +9,13 @@ from __future__ import annotations
 
 from collections.abc import Mapping
 from copy import deepcopy
+from typing import Literal
+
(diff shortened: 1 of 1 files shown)


aiven-anton:
I have two questions:

- Why do we want to support both integer and string values here? Isn't it just a misconfiguration if a user is passing `"all"`?
- I think we would not want this union type as internal representation, as it means we need to spread the logic of normalizing throughout the application. In other words, I think the type of the field should remain `int`?

muralibasani:
> I have two questions:
> 
> * Why do we want to support both integer and string values here? Isn't it just a misconfiguration if a user is passing `"all"`?
> * I think we would not want this union type as internal representation, as it means we need to spread the logic of normalizing throughout the application. In other words, I think the type of the field should remain `int`?

I think we are just allowing a similar config we have in Apache kafka configs with acks as 'all' or -1

PR #1142, opened 2025-09-12
api: handle `InvalidReferences` error on compatibility check endpoint

nosahama:
The compatibility endpoint causes the application server to crash when it fails to handle the possibility of invalid references on both the old and new schema, this PR catches this error. 

The error was made evident to us by a recent issue where the Karapace state was corrupt, the API allowed the deletion of a schema that was still referenced. This should not be allowed as #1137 verifies, we still do not know why/how this happened and are actively investigating this. 
So because a schema existed with a reference that had been deleted, the request to the compatibility check endpoint failed with `InvalidReferences` and it was unhandled.

### DIFF
--- a/src/karapace/api/controller.py
+++ b/src/karapace/api/controller.py
@@ -917,6 +917,14 @@ class KarapaceSchemaRegistryController:
                     "message": f"Invalid {schema_request.schema_type} schema",
                 },
             ) from exc
+        except InvalidReferences as exc:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail={
+                    "error_code": SchemaErrorCodes.INVALID_SCHEMA.value,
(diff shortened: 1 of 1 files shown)


muralibasani:
@nosahama any relevant test or where it was failing ?

nosahama:
@muralibasani rightly asked as I had expected.
I have omitted tests for now because it wasn't straight-forward and I already lost some time trying to add it. I can add a simple unit test against the specific function to catch a side-effect raising the expected error, but I wanted to replicate exactly what happened; a schema existed in the topic which referenced a non-existing schema and this involved me to write the message directly and then try to read it, for some reason this was difficult to replicate, it shouldn't be, that is why I most likely will add tests in a follow up, but if this has waited this long, I can try to dedicate some time to it. 
Definitely a test will be nice. 
I will do the `v4` changes and then see how to replicate in a test.

PR #1143, opened 2025-09-22
Fix protobuf schema cache

muralibasani:
Protobuf schema cache issue

References: #xxxxx

# Why this way

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -92,14 +92,15 @@ services:
       KARAPACE_SERVER_TLS_CAFILE: /opt/karapace/certs/ca/rootCA.pem
       KARAPACE_SERVER_TLS_CERTFILE: /opt/karapace/certs/cert.pem
       KARAPACE_SERVER_TLS_KEYFILE: /opt/karapace/certs/key.pem
+      KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_ISSUER: http://keycloak:8080/realms/karapace
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "karapace-audience"
       KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
-      KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_CLIENT_ID: "karapace-client"
       KARAPACE_SASL_OAUTHBEARER_ROLES_CLAIM_PATH: "resource_access.karapace-client.roles"
       KARAPACE_SASL_OAUTHBEARER_METHOD_ROLES: '{"GET": ["schema:read", "subject:read"], "POST": ["schema:write", "subject:write"], "PUT": ["config_subject:update","config_global:update"], "DELETE": ["schema:delete", "subject:delete"]}'
+      KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
 
   karapace-schema-registry-follower:
     networks:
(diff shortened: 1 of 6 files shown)


src/karapace/api/middlewares/__init__.py
```diff
@@ -39,12 +39,26 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
             request._headers = new_headers
             request.scope.update(headers=request.headers.raw)
 
+        # Bypass authentication for health check and metrics endpoints
+        if request.url.path in {"/_health", "/metrics"}:
```
nosahama: **[suggestion: non-blocking]**
It'd be nice if this can come from config.

muralibasani: @nosahama we can add the config, but how feasible it could be for customers to add a header, in their metrics or health check portals. ?

muralibasani: Added a skip config.

container/compose.yml
```diff
@@ -92,14 +92,15 @@ services:
...
+      KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_ISSUER: http://keycloak:8080/realms/karapace
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "karapace-audience"
       KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
-      KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_CLIENT_ID: "karapace-client"
       KARAPACE_SASL_OAUTHBEARER_ROLES_CLAIM_PATH: "resource_access.karapace-client.roles"
       KARAPACE_SASL_OAUTHBEARER_METHOD_ROLES: '{"GET": ["schema:read", "subject:read"], "POST": ["schema:write", "subject:write"], "PUT": ["config_subject:update","config_global:update"], "DELETE": ["schema:delete", "subject:delete"]}'
+      KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
```
nosahama, line 103: Nice, thanks 🥇

tests/unit/test_oidc.py
```diff
@@ -194,6 +197,89 @@ def test_validate_token_invalid_token(mock_jwt_decode, mock_pyjwks_client, dummy
         assert payload == {}
 
 
+@pytest.mark.asyncio
```
nosahama, line 200: **[suggestion: non-blocking]**
You do not need this `pytest` marker.

tests/unit/test_oidc.py
```diff
@@ -194,6 +197,89 @@ def test_validate_token_invalid_token(mock_jwt_decode, mock_pyjwks_client, dummy
...
+async def test_oidc_enabled_but_no_header_skips_validation_async():
+    config = Config(
+        sasl_oauthbearer_authorization_enabled=False,
+    )
+
+    app = FastAPI()
```
nosahama, line 206: **[suggestion: non-blocking]**
It'd be nice to test against the app instance we use, this does not hurt, I guess an integration test would have been more suiting, but this can be a follow up.

tests/unit/test_oidc.py
```diff
@@ -194,6 +197,89 @@ def test_validate_token_invalid_token(mock_jwt_decode, mock_pyjwks_client, dummy
...
+    """
+    OIDC is enabled but no Authorization header is sent.
+    Middleware should return 401 Unauthorized.
+    """
+    config = Config(
+        sasl_oauthbearer_jwks_endpoint_url=None,
+        sasl_oauthbearer_expected_issuer=None,
+        sasl_oauthbearer_expected_audience=None,
+        sasl_oauthbearer_sub_claim_name=None,
+        sasl_oauthbearer_authorization_enabled=True,  # OIDC enabled
```
nosahama, line 231: **[suggestion: non-blocking]**
In the integration tests suite, you could spin up a karapace server/client for OIDC enabled/disabled and run the test, even checking that skipped url routes are actually skipped.

PR #1145, opened 2025-09-23
Continuation of PR #1143

muralibasani:

Based on few comments in https://github.com/Aiven-Open/karapace/pull/1143

References: #xxxxx

# Why this way

### DIFF
--- a/tests/e2e/schema_registry/test_oidc.py
+++ b/tests/e2e/schema_registry/test_oidc.py
@@ -22,7 +22,7 @@ async def test_schema_registry_oidc(
     subject_res = await registry_async_client_oidc.post(
         f"subjects/{subject}/versions",
         json={
-            "schema": json.dumps({"type": "object", "properties": {"name": {"type": "string"}}, "required": ["name"]}),
+            "schema": json.dumps({"type": "string"}),
             "schemaType": SchemaType.JSONSCHEMA.value,
         },
     )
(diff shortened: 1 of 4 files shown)


tests/e2e/schema_registry/test_oidc.py
```diff
@@ -39,3 +39,15 @@ async def test_schema_registry_oidc_invalid_token(
     assert subject_res.status_code == 401
     assert subject_res.json_result["error"] == "Unauthorized"
     assert subject_res.json_result["reason"] == "Invalid token/payload"
+
+
+async def test_integration_oidc_enabled_no_auth_header_fails(
```
nosahama, line 44: Nice 👏 
Just for sanity, can we also check that the skipped endpoints are allowed to be called by this `registry_async_client_oidc_no_auth_header` client.

tests/unit/test_oidc.py
```diff
@@ -197,7 +197,6 @@ def test_validate_token_invalid_token(mock_jwt_decode, mock_pyjwks_client, dummy
         assert payload == {}
 
 
-@pytest.mark.asyncio
 async def test_oidc_enabled_but_no_header_skips_validation_async():
```
nosahama: I'd be keen on removing these tests that use the bare `FastAPI` object and instead move to the `e2e` or `integration` suite like you've done. 
That's if it isn't much effort.

tests/e2e/schema_registry/test_oidc.py
muralibasani: Good point. Added.

tests/unit/test_oidc.py
muralibasani: Done. We already have them, with this newly added test.

muralibasani:
> @muralibasani can we improve the PR title 😃 , even just "Continuation of PR #1143" or something like that.

done. :)

PR #1150, opened 2025-10-08
Fix flakiness of tests

muralibasani:

For the below tests
test_regression_config_for_inexisting_object_should_not_throw
test_regression_soft_delete_schemas_should_be_registered

- After events are produced, a timout is necessary when getting the future object. timeout is added now.
- Also producer acks is set to all, so that we wait for event to be produced for sure.
- Added another assert to see if the offset is seen.

References: #xxxxx

# Why this way

### DIFF
--- a/tests/integration/test_schema_reader.py
+++ b/tests/integration/test_schema_reader.py
@@ -204,7 +204,8 @@ async def test_regression_config_for_inexisting_object_should_not_throw(
             producer.flush()
             msg = future.result(timeout=2)
 
-            schema_reader._offset_watcher.wait_for_offset(msg.offset(), timeout=5)
+            seen = schema_reader._offset_watcher.wait_for_offset(msg.offset(), timeout=5)
+            assert seen is True
 
             assert (
                 database.find_subject(subject=Subject(subject)) is not None


tests/integration/test_schema_reader.py
```diff
@@ -117,7 +118,7 @@ async def test_regression_soft_delete_schemas_should_be_registered(
 
             schema_reader._offset_watcher.wait_for_offset(msg.offset(), timeout=5)
```
jjaakola-aiven: This is a blocking call. If offset is marked as seen the schema must be in the in memory database.
Can you enable the `strict` mode in the config for the test and see if there is an exception catched when handing the message?
Option is also to see if the timeout is elapsed, the call returns the boolean if condition was met. And on timeout the return value is `False`. That can be asserted.

muralibasani: It was hard to get any logs. 
Enabling strict mode, did not trigger any exception.

I thought we needed some timeouts when getting future. Added. wdyt

tests/integration/test_schema_reader.py
```diff
@@ -113,10 +114,10 @@ async def test_regression_soft_delete_schemas_should_be_registered(
                 value=json_encode(value, binary=True),
             )
             producer.flush()
-            msg = future.result()
-
-            schema_reader._offset_watcher.wait_for_offset(msg.offset(), timeout=5)
+            msg = future.result(timeout=2)
```
aiven-anton, line 117: Please add some justification for this to the commit message, I have no idea why we would add a timeout here, as we are flushing right before.

tests/integration/test_schema_reader.py
```diff
@@ -141,7 +142,7 @@ async def test_regression_soft_delete_schemas_should_be_registered(
                 value=json_encode(value, binary=True),
             )
             producer.flush()
-            msg = future.result()
+            msg = future.result(timeout=2)
```
aiven-anton, line 145: Same here, we have flushed, waiting should not change anything. Have we actually observed something that would hint at this solving a problem we have? (I.e., I think this would raise if the future is not complete when calling `.result()`. Have we seen such exceptions?)

tests/integration/test_schema_reader.py
```diff
@@ -73,6 +73,7 @@ async def test_regression_soft_delete_schemas_should_be_registered(
     config.admin_metadata_max_age = 2
     config.group_id = group_id
     config.topic_name = topic_name
+    config.producer_acks = 1
```
aiven-anton, line 76: Please add reasoning in commit message, why this change?

tests/integration/test_schema_reader.py
```diff
@@ -200,10 +202,10 @@ async def test_regression_config_for_inexisting_object_should_not_throw(
                 value=json_encode(value, binary=True),
             )
             producer.flush()
-            msg = future.result()
+            msg = future.result(timeout=2)
```
aiven-anton, line 205: Same here.

tests/integration/test_schema_reader.py
```diff
@@ -200,10 +202,10 @@ async def test_regression_config_for_inexisting_object_should_not_throw(
...
+
+            schema_reader._offset_watcher.wait_for_offset(msg.offset(), timeout=5)
 
-            seen = schema_reader._offset_watcher.wait_for_offset(msg.offset(), timeout=5)
-            assert seen is True
```
aiven-anton, line 206: Why do we stop asserting here?

aiven-anton (changes requested):

tests/integration/test_schema_reader.py
muralibasani: Well, don't have a hard reason for this , but generally I think it is resource intensive sometimes.
And timeout on the future object is provided by the api to handle such cases where processes are slow.

tests/integration/test_schema_reader.py
muralibasani: Was thinking this will wait until the produced message is acked, adding some more wait time

tests/integration/test_schema_reader.py
muralibasani: Added it back.

PR #1151, opened 2025-10-10
Flaky test - Reset counter for every compatibility test

muralibasani:

counter reset for every compatibility test when config is put and get with retries

Failed build
https://github.com/Aiven-Open/karapace/actions/runs/18306671586/job/52125421718?pr=1149

References: #xxxxx

# Why this way

### DIFF
--- a/tests/integration/test_request_forwarding.py
+++ b/tests/integration/test_request_forwarding.py
@@ -81,7 +81,7 @@ async def test_schema_request_forwarding(
 ) -> None:
     master_url, follower_url = registry_async_pair
 
-    max_tries, counter = 5, 0
+    max_tries = 5
     wait_time = 60
     subject = create_subject_name_factory(subject)()
     schema = {"type": "string"}
@@ -93,6 +93,7 @@ async def test_schema_request_forwarding(
         else:
             path = "config"
         for compat in ["FULL", "BACKWARD", "FORWARD", "NONE"]:
+            counter = 0
             resp = await request_forwarding_retry_client.put(f"{follower_url}/{path}", json={"compatibility": compat})
             assert resp.ok
             while True:
@@ -185,7 +186,7 @@ async def test_schema_request_forwarding_tls(
 ) -> None:
     master_url, follower_url = registry_async_pair_tls
 
-    max_tries, counter = 5, 0
+    max_tries = 10
     wait_time = 0.5
     subject = create_subject_name_factory(subject)()
     schema = {"type": "string"}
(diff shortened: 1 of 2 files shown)


tests/integration/test_schema_registry_auth.py
```diff
@@ -212,9 +212,10 @@ async def test_sr_auth_forwarding(
 
     # Test primary/replica forwarding with global config setting
     primary_url, replica_url = registry_async_auth_pair
-    max_tries, counter = 5, 0
-    wait_time = 0.5
+    max_tries = 5
+    wait_time = 60
```
aiven-anton: issue: This is the wait time used in the sleep, waiting a minute here seems very long?

tests/integration/test_request_forwarding.py
```diff
@@ -185,8 +186,8 @@ async def test_schema_request_forwarding_tls(
 ) -> None:
     master_url, follower_url = registry_async_pair_tls
 
-    max_tries, counter = 5, 0
-    wait_time = 0.5
+    max_tries = 10
+    wait_time = 60
```
aiven-anton: Same here, seems like a pretty drastic extension of the sleep time. Maybe correcting of the counter reset is enough to make the tests stable?

tests/integration/test_schema_registry_auth.py
muralibasani: In other test there was 60 https://github.com/Aiven-Open/karapace/pull/1151/files#diff-5ae460886bfc7a462698e746412a31bcdc76c1c618ed7eb478f192f5c1e9c4a2R85 for same compat check. 
So thought of aligning with that.

tests/integration/test_request_forwarding.py
muralibasani: Only to align with other test, set it to 60 secs

tests/integration/test_schema_registry_auth.py
aiven-anton: That means this test could take up to 20 minutes to run, isn't that very very long?

aiven-anton: It would be nice to only do one of these changes, so that we can isolate what's needed. If the counter reset shows to not help when moving forward, it can be revisited. If, on the other hand it does help, maybe it's so that we can align the other value to some value a lot lower?

muralibasani: Agree, changed it to 10 secs.

aiven-anton: Hmm why? That doesn't achieve isolation then either, couldn't we try leaving it unchanged so that we truly identify the culprit?

muralibasani: :) 0.5 secs was like really low to get this from replica to primary for this test I felt. but lets isolate it.

muralibasani: done, let's see.

Issue #1154, opened 2025-10-24
BUG: Follower is not able to parse error responses from leader if an invalid request is sent.


harishva23:
# What happened?

If a user or client sends a mutation request (like adding new schema, change compatibility level etc) to Karapace which has a leader-follower deployment to a follower, that request is forwarded to the leader which is expected. But if the leader throws an error such as 409 or other error codes as defined, the follower is not parsing it correctly and instead returns "Internal Server Error" which is a plaintext response. This is causing Confluent clients to log errors as it expects a FastAPI response. 

Currently only valid requests which return 200 or some error responses are being handled correctly by the follower.

# What did you expect to happen?

We expect that the error message to be properly parsed by the follower and sent back to the client. 

# What else do we need to know?
This is printing stack trace in the follower pod logs which could expose sensitive info and accumulation of these logs. Also this causes the follower to respond with "Internal Server Error" even though the leader has handled the error.
More details to reproduce this error are given below:

We have currently deployed karapace with a leader and follower pod as a StatefulSet in k8s:

 kubectl get po -n schema-registry
NAME                         READY   STATUS    RESTARTS   AGE
karapace-schema-registry-0   1/1     Running   0          8d
karapace-schema-registry-1   1/1     Running   0          8d

//query health endpoint to get leader
GET http://karapace-schema-registry.schema-registry:8081/_health
Response:
{
    "karapace_version": "5.0.2",
    "status": {
        "schema_registry_ready": true,
        "schema_registry_startup_time_sec": 6.625380557961762,
        "schema_registry_reader_current_offset": 493,
        "schema_registry_reader_highest_offset": 493,
        "schema_registry_is_primary": true,
        "schema_registry_is_primary_eligible": true,
        "schema_registry_primary_url": "http://karapace-schema-registry-0.karapace-schema-registry.schema-registry:8081",
        "schema_registry_coordinator_running": true,
        "schema_registry_coordinator_generation_id": 6
    },
    "healthy": true
}

As per the above request, http://karapace-schema-registry-0.karapace-schema-registry.schema-registry:8081 is the leader's url. 
Now , we will try to change the compatibility mode which is invalid like below to the leader:
PUT http://karapace-schema-registry-0.karapace-schema-registry.schema-registry:8081/config

Request body:
{"compatibility":"RANDOM STRING"}

Response: 422 Unprocessable Entity
{
    "error_code": 42203,
[...]

PR #1155, opened 2025-10-24
Fix/follower error handling

harishva23:

Fixes the follower error handling to correctly resolve error messages from leader and corrected internal server error being raised if a non-registered user credentials in file-based authentication is passed in the request.

References: #1154

# Why this way

When I faced regular error 500 responses if users tried to insert incompatible or invalid schemas , I saw the logs of the leader pod were empty (set to ERROR) , but the follower pods had a full stack trace which meant that the error in the follower wasn't being handled properly. For requests returning 200, the follower was passing the responses from leader correctly. So I have added the correct error handling mechanisms to prevent these Internal Server Errors messages being thrown even though the karapace server is working as intended.

Similarly, the same error handling issue was present in basic authentication where if I a user sends a wrong username (which doesn't exist in the file), a generic ValueError was being raised which caused an Internal Server Error. But if the username was correct(existed in auth file) and the password was incorrect , then I used to get a correct Unauthorised error message. Therefore, I updated the error handling here to get Unauthorised Error for both the cases as the 2nd case was handled correctly.

### DIFF
--- a/src/karapace/api/forward_client.py
+++ b/src/karapace/api/forward_client.py
@@ -9,7 +9,7 @@ from karapace.core.config import Config
 from karapace.version import __version__
 from pydantic import BaseModel
 from typing import overload, TypeVar, Union
-
+from karapace.api.routers.requests import ErrorResponse
 import aiohttp
 import async_timeout
 import logging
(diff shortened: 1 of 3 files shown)


harishva23:
Closed due to improper commit history. Ignore this request

PR #1159, opened 2025-10-28
tests: add poll timeout to `test_roundtrip_from_file`

nosahama:
Attempt to improve the test and reduce the flakiness by increasing the poll timeout from default 1 minute to 5 minutes. 

# Why this way
As this is only mostly flaky on the CI, as it passes every time on local setup, we suspect that the tests resources are struggling. We've also scoped the fix to just this test, it may be necessary to add this increased timeout to other places that fail using the soon to be deprecated backup CLI. 

If this doesn't work, we can go ahead and `xfail` it (#1158) as the backup feature will be deprecated.

### DIFF
--- a/tests/integration/backup/test_v3_backup.py
+++ b/tests/integration/backup/test_v3_backup.py
@@ -406,6 +406,7 @@ def test_roundtrip_from_file(
             "--replication-factor=1",
             "--location",
             str(backup_location),
+            "--poll-timeout=PT5M",
         ],
         capture_output=True,
         check=True,


tests/integration/backup/test_v3_backup.py
```diff
@@ -406,6 +406,7 @@ def test_roundtrip_from_file(
             "--replication-factor=1",
             "--location",
             str(backup_location),
+            "--poll-timeout=PT5M",
```
muralibasani, line 409: I am not convinced about this being resource intensive, as this has been existing for long time

nosahama:
@muralibasani please feel free to `xfail` the test, I Cannot dedicate further time to this 😄

nosahama:
> > @muralibasani why are you eager to `xfail` a test that works, i really do not understand.
> 
> Not sure why you think it works. Here is a broken ci https://github.com/Aiven-Open/karapace/actions/runs/18491765623/job/52686864097 and have noticed it a few times.

Then why did it not break in this current run 😄 , anyways, we can move forward 👍

PR #1164, opened 2025-11-03
Fix get_schema coroutine issue

muralibasani:
**Error**
372, in deserialize schema, _ = await self.get_schema_for_id(schema_id) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/lib64/python3.12/site-packages/karapace/serialization.py", line 343, in get_schema_for_id schema_typed, subjects = await self.registry_client.get_schema_for_id(schema_id) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/lib64/python3.12/site-packages/karapace/serialization.py", line 235, in get_schema_for_id _, schema, version = await self.get_schema(reference.subject, reference.version) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ RuntimeError: cannot reuse already awaited coroutine

**Issue** : RuntimeError: cannot reuse already awaited coroutine occurred when deserializing protobuf messages with schema references. This happened when multiple references pointed to the same schema.

**Root Cause**

- In get_schema_for_id(), references were fetched sequentially in a loop:
for reference in parsed_references:    _, schema, version = await self.get_schema(reference.subject, reference.version)

- And get_schema() is decorated with @alru_cache.
If multiple references pointed to the same (subject, version), the cache returned the same coroutine object.
Awaiting the same coroutine multiple times raises RuntimeError: cannot reuse already awaited coroutine.

Fix : 
- Deduplicate references before fetching: group by (subject, version) so each unique schema is fetched once.
- Fetch in parallel: use asyncio.gather() to fetch all unique schemas concurrently.
- Map results back: build the dependencies dictionary using the fetched schemas.

### DIFF
--- a/src/karapace/core/serialization.py
+++ b/src/karapace/core/serialization.py
@@ -240,27 +240,27 @@ class SchemaRegistryClient:
                 dependencies = {}
 
                 # Deduplicate references by (subject, version)
-                unique_keys: dict[tuple[Subject, Version | None], list[Reference]] = {}
+                unique_keys_schema_references: dict[tuple[Subject, Version | None], list[Reference]] = {}
                 for reference in parsed_references:
                     key = (
                         (reference.subject, None)
                         if isinstance(reference, LatestVersionReference)
                         else (reference.subject, reference.version)
                     )
-                    unique_keys.setdefault(key, []).append(reference)
+                    unique_keys_schema_references.setdefault(key, []).append(reference)
 
                 # Fetch all unique schemas in parallel
                 coros = [
                     self.get_schema(subject, version) if version else self.get_schema(subject)
-                    for subject, version in unique_keys.keys()
+                    for subject, version in unique_keys_schema_references.keys()
                 ]
                 schema_results = await asyncio.gather(*coros, return_exceptions=False)
 
                 # Map fetched results to keys
-                schema_map = dict(zip(unique_keys.keys(), schema_results))
+                schema_map = dict(zip(unique_keys_schema_references.keys(), schema_results))
 
                 # Build dependencies for all references in a single pass
-                for key, references_for_key in unique_keys.items():
+                for key, references_for_key in unique_keys_schema_references.items():
                     _, schema, version = schema_map[key]
                     for reference in references_for_key:
                         dependencies[reference.name] = Dependency(reference.name, reference.subject, version, schema)


src/karapace/core/serialization.py
```diff
@@ -124,11 +124,13 @@ def __init__(
         self.base_url = schema_registry_url
 
     async def post_new_schema(
-        self, subject: str, schema: ValidatedTypedSchema, references: Reference | None = None
+        self, subject: str, schema: ValidatedTypedSchema, references: Sequence[Reference] | None = None
```
muralibasani, line 127: this is unrelated fix but without this, integration tests would fail to run with the coroutine fix.

src/karapace/core/serialization.py
```diff
@@ -237,12 +239,66 @@ async def get_schema_for_id(self, schema_id: SchemaId) -> tuple[TypedSchema, lis
...
+                for key in ordered_keys:
+                    subject, version = key  # Unpack the tuple key (subject, version)
+                    if version is not None:
+                        coro = self.get_schema(subject, version)
+                    else:
+                        coro = self.get_schema(subject)
+                    coros.append(coro)
+
+                # Await all coroutines in parallel - gather handles @alru_cache correctly
+                schema_results = await asyncio.gather(*coros, return_exceptions=False)
```
muralibasani, line 257: get results of all coroutines

src/karapace/core/serialization.py
```diff
@@ -237,13 +239,31 @@ async def get_schema_for_id(self, schema_id: SchemaId) -> tuple[TypedSchema, lis
...
+                # Deduplicate references by (subject, version)
+                unique_keys: dict[tuple[Subject, Version | None], list[Reference]] = {}
                 for reference in parsed_references:
-                    if isinstance(reference, LatestVersionReference):
-                        _, schema, version = await self.get_schema(reference.subject)
-                    else:
-                        _, schema, version = await self.get_schema(reference.subject, reference.version)
-
-                    dependencies[reference.name] = Dependency(reference.name, reference.subject, version, schema)
+                    key = (
```
muralibasani, line 245: get keys based on latest or a schema version

src/karapace/core/serialization.py
```diff
@@ -237,13 +239,31 @@ async def get_schema_for_id(self, schema_id: SchemaId) -> tuple[TypedSchema, lis
             if parsed_references:
                 dependencies = {}
 
+                # Deduplicate references by (subject, version)
+                unique_keys: dict[tuple[Subject, Version | None], list[Reference]] = {}
```
nosahama: **[nit]**
`unique_keys` naming is too vague.

muralibasani: yes, renamed.

src/karapace/core/serialization.py
nosahama: **[discussion]**
How about error handling?

src/karapace/core/serialization.py
```diff
@@ -124,11 +124,13 @@ def __init__(
...
 
     async def post_new_schema(
-        self, subject: str, schema: ValidatedTypedSchema, references: Reference | None = None
+        self, subject: str, schema: ValidatedTypedSchema, references: Sequence[Reference] | None = None
     ) -> SchemaId:
         if schema.schema_type is SchemaType.PROTOBUF:
             if references:
-                payload = {"schema": str(schema), "schemaType": schema.schema_type.value, "references": references.json()}
+                # Convert list of Reference objects to list of dicts for JSON serialization
+                references_json = [ref.to_dict() for ref in references]
```
nosahama, line 132: **[non-blocking]**
we could kill the `references_json` variable and put the list comprehension directly within the `payload` as it is not referenced anywhere else.

src/karapace/core/serialization.py
```diff
@@ -237,13 +239,31 @@ async def get_schema_for_id(self, schema_id: SchemaId) -> tuple[TypedSchema, lis
...
-                    else:
-                        _, schema, version = await self.get_schema(reference.subject, reference.version)
-
-                    dependencies[reference.name] = Dependency(reference.name, reference.subject, version, schema)
+                    key = (
+                        (reference.subject, None)
+                        if isinstance(reference, LatestVersionReference)
+                        else (reference.subject, reference.version)
+                    )
+                    unique_keys_schema_references.setdefault(key, []).append(reference)
```
nosahama, line 250: **[non-blocking]**
You could use a [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) for this.

src/karapace/core/serialization.py
davide-armand, line 1: @muralibasani Maybe I'm missing something, but shouldn't the `@alru_cache` annotation cover this case (calling the same decorated method multiple times with same args)?
The example in their readme is doing that: https://github.com/aio-libs/async-lru
I wonder is something else is going on here

davide-armand, line 1: An alternative solution could be caching results locally (although it kind of defeats the point of using `@alru_cache`):
```python
            if parsed_references:
                dependencies = {}

                schema_cache = {}
                for reference in parsed_references:
                    # Create cache key for this reference
                    cache_key = (
                        reference.subject,
                        reference.version if not isinstance(reference, LatestVersionReference) else None,
                    )

                    if cache_key in schema_cache:
                        # Reuse cached result
                        _, schema, version = schema_cache[cache_key]
                    else:
                        # Fetch schema and cache the result
                        if isinstance(reference, LatestVersionReference):
                            schema_result = await self.get_schema(reference.subject)
                        else:
                            schema_result = await self.get_schema(reference.subject, reference.version)
                        schema_cache[cache_key] = schema_result
                        _, schema, version = schema_result

                    dependencies[reference.name] = Dependency(reference.name, reference.subject, version, schema)
            else:
                dependencies = None

```

muralibasani: @davide-armand could be yes.

How about replacing alru with lru and return a async task like below.

```
@lru_cache(maxsize=100)
def _get_schema_task(self, subject: Subject, version: Version | None) -> asyncio.Task:
    return asyncio.create_task(self._get_schema_recursive(subject, set(), version))

async def get_schema(
    self,
    subject: Subject,
    version: Version | None = None,
) -> tuple[SchemaId, ValidatedTypedSchema, Version]:
    return await _get_schema_task(self, subject, version)
```

It is like one task per subject/version is cached.

muralibasani: And we don't have to change anything in parent call

davide-armand: I think that would not work, because `_get_schema_task`  would cache the coroutine (not the returned value) so a second call to `_get_schema_task` with same args would result in the same runtime error (the cache would return the same coroutine that already completed on first call, raise the error)

Issue #1140, opened 2025-09-09
BUG: Schema Metadata doesn't return if a schema is soft-deleted or not when querying with deleted=true


harishva23:
# What happened?

I am using Confluent Schema Registry API and when i query a soft deleted schema metadata using the lookupDeletedSchemas as true, I get the correct schema but the deleted field is null and there is no direct endpoint which returns it's correct status

# What did you expect to happen?
I expected the delete field to be populated or any other endpoint which offers this info in the response payload.

# What else do we need to know?
This can lead to confusion or we need to design big workarounds to determine if a particular schema is soft deleted or not.

PR #1166, opened 2025-11-05
Fix follower error handling when leader returns invalid compatibility…

harishva23:
About this change - What it does

This PR fixes 2 issues - error handling inside the follower and error handling for file-based basic authentication

Earlier, the follower could not parse error messages from leader if write requests were made to the follower and it failed due to some error like incompatible or invalid schema , trying to delete a non-existent schema etc. Now this is able to handle these error messages and pass it on to the user or client.

In the file-based authentication, if a non-existent user was sent in the Headers ( a username not in auth file), karapace threw an Internal Server Error instead of Unauthorized message. This did not happen when a correct username but wrong password was passed. So I have fixed this issue and it returns an Unauthorized error message for both the cases.

References: https://github.com/Aiven-Open/karapace/issues/1154
Why this way

Since Confluent client always expects a FastAPI response, I have changed the code to return proper format of error messages from follower. This leads to consistent error messages from both leader and follower so that the user can decide the next steps.

For the file-based authentication, raising a generic ValueError results in Internal Server Error being thrown and a full stacktrace appears in the logs which reveals that User is not found. Since karapace already throws a Unauthorised Error if wrong password is sent, I have removed the logic raising ValueError and instead returning None type which is then handled in the auth.py file in the authenticate function.

I have also included test cases for forwarding client as well as the basic authentication tests as well. For the follower client, I have patched the SSL context creation for the tests.

### DIFF
--- a/src/karapace/api/forward_client.py
+++ b/src/karapace/api/forward_client.py
@@ -45,7 +45,7 @@ class ForwardClient:
         *,
         request: Request,
         primary_url: str,
-    ) -> tuple[bytes, int]:  # Return both body and status code
+    ) -> tuple[str, int]:  # Return both body and status code
         LOG.info("Forwarding %s request to remote url: %r since we're not the master", request.method, request.url)
         timeout = 60.0
         func = getattr(self._forward_client, request.method.lower())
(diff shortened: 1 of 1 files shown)


harishva23:
> @harishva23 I see a conflict. Can you pls fix ?

Hi @muralibasani  , the commit for #1140 has been reverted after I raised the PR. Therefore I shouldn't include that changes in my PR right?

harishva23:
Hi @muralibasani  I have fixed the lint issues using the pre-commit tool and I have ran my tests locally with all cases passing.

PR #1176, opened 2025-11-28
Fix metrics issues

muralibasani:

Multiple issues fixed.
1. In prometheus.yml hosts were not defined correctly. And port 8889 should be usable.
2. In karapace.json, promql expr was doubling counts like subject totals due to 2 instances. Now picks up only from primary
3. In collector-config, earlier exporter was both otlp and prometheus. Now set to only prometheus to scrape open telemetry metrics.
4. Metrics scraping had an issue with octet-stream content-type. Fixed in prometheus.py
5. schema_reader now initializes counts of subjects, schemas and soft deleted schema versions.
6. Added some debug and info log statements

References: #xxxxx

# Why this way


container/grafana/dashboards/karapace.json
```diff
@@ -501,7 +501,7 @@
           },
           "disableTextWrap": false,
           "editorMode": "builder",
-          "expr": "karapace_schema_reader_subjects_total",
+          "expr": "max(karapace_schema_reader_subjects_total{app=\"karapace-schema-registry\"}) or vector(0)",
```
davide-armand: question: why is `max` needed here? Is a time window involved? Or are there multiple instances?

src/karapace/api/routers/metrics.py
```diff
@@ -20,4 +21,11 @@
 async def metrics(
     prometheus: PrometheusInstrumentation = Depends(Provide[SchemaRegistryContainer.karapace_container.prometheus]),
 ) -> Response:
-    return Response(content=await prometheus.serve_metrics(), media_type=prometheus.CONTENT_TYPE_LATEST)
+    try:
+        await prometheus.serve_metrics()
+    except HTTPResponse as ex:
+        # serve_metrics() raises HTTPResponse for RestApp compatibility
```
davide-armand: nit: maybe good to move this comment to the definition of `serve_metrics()`, for clarity.

src/karapace/core/schema_reader.py
```diff
@@ -510,20 +530,15 @@ def _report_schema_metrics(
...
-            )
+        # Always update gauges to reflect current database state
+        # This ensures metrics stay accurate even if no new records are processed
+        num_schemas = self.database.num_schemas()
+        num_subjects = self.database.num_subjects()
+        self.stats.set_schemas_num_total(value=num_schemas)
+        self.stats.set_subjects_num_total(value=num_subjects)
+
+        live_versions, soft_deleted_versions = self.database.num_schema_versions()
+        self.stats.set_schema_versions_num_total(live_versions=live_versions, soft_deleted_versions=soft_deleted_versions)
```
davide-armand: nit: this looks similar to what's done in `_initialize_metrics()`, could be refactored

container/grafana/dashboards/karapace.json
muralibasani: Yea max was not required and there are 2 instances. primary and follower. Mentioned the correct param now.

src/karapace/core/schema_reader.py
muralibasani: Yea there is a redundancy for params initialization. Done.

PR #1178, opened 2025-12-05
Performance test - adding delete consumers, instances, subscriptions

muralibasani:

In the performance test script, consumers are created but not deleted.

Adding consumer deletion (instances, group, subscriptions)
References: #xxxxx

# Why this way

### DIFF
--- a/performance-test/rest-proxy-produce-consume-test.py
+++ b/performance-test/rest-proxy-produce-consume-test.py
@@ -10,6 +10,7 @@ from locust.contrib.fasthttp import ResponseContextManager
 from locust.env import Environment
 
 import json
+import logging
 import os
 import uuid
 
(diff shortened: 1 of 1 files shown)


performance-test/rest-proxy-produce-consume-test.py
```diff
@@ -45,48 +45,71 @@
...
+            headers=HEADER_ACCEPT,
+            name="/consumers/[GROUP]/instances/[ID]/subscription DELETE",
         )
 
+        # 2. DELETE consumer instance
+        self.client.delete(
+            f"/consumers/{self.consumer_group}/instances/{self.consumer_instance_id}",
+            headers=HEADER_ACCEPT,
+            name="/consumers/[GROUP]/instances/[ID] DELETE",
+        )
```
davide-armand: not blocking: maybe overkill since this is just a cleanup for a test, but probably good to check the return code and log a warning/error accordingly.
Also if `delete()` raises exception, maybe good to handle them (logging them might be enough?).

muralibasani: Yea, some log statements will help. Added with try/catch.

performance-test/rest-proxy-produce-consume-test.py
```diff
@@ -41,52 +42,103 @@
...
+                name="/consumers/[GROUP]/instances/[ID]/subscription DELETE",
+                catch_response=True,
+            ) as resp:
+                if resp.status_code >= 400:
+                    msg = (
+                        f"Failed to delete subscription for {self.consumer_group}/{self.consumer_instance_id}: "
+                        f"{resp.status_code} {resp.text}"
+                    )
+                    LOG.warning(msg)
+                    resp.failure(msg)
```
davide-armand, line 97: This is not needed right? Since `resp` is not used afterwards.
Same in the other delete below

muralibasani: Actually not needed, but calling resp.failure makes the request show up as a failure in Locust stats instead of a success

Issue #1173, opened 2025-11-24
/openapi.json endpoint returns 500


dcolazin:
# What happened?

The GET `/openapi.json` endpoint returns 500. The stacktrace in the logs is as follows:

```
uvicorn.access      	MainThread	INFO    	172.18.0.1:51078 - "GET /openapi.json HTTP/1.1" 500
uvicorn.error       	MainThread	ERROR   	Exception in ASGI application
  + Exception Group Traceback (most recent call last):
  |   File "/venv/lib/python3.10/site-packages/starlette/_utils.py", line 76, in collapse_excgroups
  |     yield
  |   File "/venv/lib/python3.10/site-packages/starlette/middleware/base.py", line 186, in __call__
  |     async with anyio.create_task_group() as task_group:
  |   File "/venv/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 772, in __aexit__
  |     raise BaseExceptionGroup(
  | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "/venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 409, in run_asgi
    |     result = await app(  # type: ignore[func-returns-value]
    |   File "/venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    |     return await self.app(scope, receive, send)
    |   File "/venv/lib/python3.10/site-packages/fastapi/applications.py", line 1054, in __call__
    |     await super().__call__(scope, receive, send)
    |   File "/venv/lib/python3.10/site-packages/starlette/applications.py", line 113, in __call__
    |     await self.middleware_stack(scope, receive, send)
    |   File "/venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 187, in __call__
    |     raise exc
    |   File "/venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 165, in __call__
    |     await self.app(scope, receive, _send)
    |   File "/venv/lib/python3.10/site-packages/starlette/middleware/base.py", line 185, in __call__
    |     with collapse_excgroups():
    |   File "/usr/local/lib/python3.10/contextlib.py", line 153, in __exit__
    |     self.gen.throw(typ, value, traceback)
[...]

PR #1181, opened 2025-12-17
Fix pydantic v2 for openapi generation

muralibasani:

- Fix Pydantic v2 integration for Subject by adding explicit core/json schema hooks so GET /openapi.json generates without errors.
- Keep existing validation/serialization behavior as is to avoid breaking user models and clients.

#1173 

# Why this way

### DIFF
--- a/src/karapace/core/typing.py
+++ b/src/karapace/core/typing.py
@@ -48,12 +48,13 @@ class Subject(str):
         )
 
     @classmethod
-    def __get_pydantic_json_schema__(cls, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
+    def __get_pydantic_json_schema__(
+        cls, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+    ) -> JsonSchemaValue:
         return handler(core_schema)
 
     @classmethod
     def __get_validators__(cls) -> Generator[Callable[[str, ValidationInfo], str], None, None]:
-        # Kept for backward compatibility with pydantic v1 style validation.
         yield cls.validate
 
     @classmethod


muralibasani:
openapi-karapace

nosahama (approved):

src/karapace/core/typing.py
```diff
@@ -37,13 +39,26 @@
...
+        )
+
+    @classmethod
+    def __get_pydantic_json_schema__(
+        cls, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+    ) -> JsonSchemaValue:
+        return handler(core_schema)
+
+    @classmethod
     def __get_validators__(cls) -> Generator[Callable[[str, ValidationInfo], str], None, None]:
```
nosahama, line 57: **[non-blocking]**
I guess this is no longer needed, or maybe for v4 🤔

muralibasani: this could remain for backwards compatibility for v1

src/karapace/core/typing.py
```diff
@@ -37,13 +39,26 @@
 
 class Subject(str):
     @classmethod
-    # TODO[pydantic]: We couldn't refactor `__get_validators__`, please create the `__get_pydantic_core_schema__` manually.
-    # Check https://docs.pydantic.dev/latest/migration/#defining-custom-types for more information.
+    def __get_pydantic_core_schema__(cls, source: type[Any], handler: GetCoreSchemaHandler) -> core_schema.CoreSchema:
```
nosahama, line 42: **[non-blocking]**
The handler will be unused, since this is a `str` substype, you can change return statement to - `return core_schema.no_info_after_validator_function(cls.validate, handler(str)), serialization=core_schema.to_string_ser_schema()`. I think even with the handler, we do not need the `serialization`.

Issue #1175, opened 2025-11-25
mTLS is not enforcing the client certificate to be passed during the handshake


geekbala:
# What happened?

I wanted to run the karapace on mTLS mode. I'm using 5.0.3 docker and passing all the necessary configurations like server key, cert and ca file. Even then karapace didn't ask the client certificates to be passed during the SSL handshake. 

# What did you expect to happen?

curl --cacert certs/ca.crt --resolve karapace:8081:127.0.0.1 https://karapace:8081/subjects. -- Expected this to return fail. but this give the results.

curl --cacert certs/ca.crt --cert certs/client.crt --key certs/client.key \
     --resolve karapace:8081:127.0.0.1 https://karapace:8081/subjects. -- Expected to return the result

# What else do we need to know?

While investigating further, what I found out that uvicorn.Run() supports ssl_cert_reqs parameter and the possible values are ssl.CERT_NONE ( default ), ssl.CERT_OPTIONAL and ssl.CERT_REQUIRED. Since default is CERT_NONE it doesn't enforce the client certificate and its validation. setting ssl.CERT_REQUIRED works for me

PR #1182, opened 2025-12-17
Add configurable mTLS client auth via server_tls_client_auth

muralibasani:

Add configurable mTLS client auth via server_tls_client_auth and pass ssl_cert_reqs to uvicorn, enforcing client certs when set to required.

#1175 

# Why this way

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -491,6 +491,9 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
    * - ``server_tls_cafile``
      - ``/path/to/cafile``
      - Filename to the SSL CA certificate.
+  * - ``server_tls_client_auth``
+    - ``none``
+    - Client certificate requirement: ``none``, ``optional``, or ``required``.
    * - ``registry_scheme``
      - ``http``
      - Schema Registry scheme to use for rest-proxy, http | https (if certificates are provided).
--- a/src/karapace/__main__.py
+++ b/src/karapace/__main__.py
@@ -7,6 +7,7 @@ from karapace.api.container import SchemaRegistryContainer
 from karapace.api.factory import create_karapace_application, karapace_schema_registry_lifespan
 from karapace.api.telemetry.container import TelemetryContainer
 from karapace.core.auth_container import AuthContainer
+from karapace.core.config import ServerTLSClientAuth, get_client_auth_verify_mode
 from karapace.core.container import KarapaceContainer
 
 import karapace.api.controller
@@ -24,6 +25,7 @@ import karapace.api.telemetry.middleware
 import karapace.api.telemetry.setup
 import karapace.core.instrumentation.tracer
 import karapace.api.user
+import ssl
 import uvicorn
 
 from karapace.core.metrics_container import MetricsContainer
@@ -97,6 +99,17 @@ if __name__ == "__main__":
 
     config = karapace_container.config()
     app = create_karapace_application(config=config, lifespan=karapace_schema_registry_lifespan)
+
+    # Calculate SSL certificate requirements for uvicorn
+    ssl_cert_reqs: int = ssl.CERT_NONE
+    if config.server_tls_certfile and config.server_tls_keyfile:
+        client_auth_mode = ServerTLSClientAuth(config.server_tls_client_auth)
+        verify_mode = get_client_auth_verify_mode(
+            client_auth_mode,
+            has_cafile=bool(config.server_tls_cafile),
+        )
+        ssl_cert_reqs = int(verify_mode)
+
     uvicorn.run(
         app,
         host=config.host,
@@ -106,4 +119,5 @@ if __name__ == "__main__":
         ssl_keyfile=config.server_tls_keyfile,
         ssl_certfile=config.server_tls_certfile,
         ssl_ca_certs=config.server_tls_cafile,
+        ssl_cert_reqs=ssl_cert_reqs,
     )
--- a/src/karapace/core/config.py
+++ b/src/karapace/core/config.py
@@ -45,6 +45,29 @@ class KarapaceTelemetryOTelExporter(str, enum.Enum):
     NOOP = "NOOP"
 
 
+class ServerTLSClientAuth(str, enum.Enum):
+    NONE = "none"
+    OPTIONAL = "optional"
+    REQUIRED = "required"
+
+
+def get_client_auth_verify_mode(
+    client_auth_mode: ServerTLSClientAuth,
+    *,
+    has_cafile: bool,
+) -> ssl.VerifyMode:
+    match client_auth_mode:
+        case ServerTLSClientAuth.REQUIRED:
+            if not has_cafile:
+                raise InvalidConfiguration("`server_tls_client_auth` set to required but `server_tls_cafile` not provided")
+            return ssl.CERT_REQUIRED
+        case ServerTLSClientAuth.OPTIONAL:
+            return ssl.CERT_OPTIONAL
+        case ServerTLSClientAuth.NONE:
+            return ssl.CERT_NONE
+    raise InvalidConfiguration("`server_tls_client_auth` must be one of: none, optional, required")
+
+
 class KarapaceTelemetry(BaseModel):
     otel_endpoint_url: str | None = None
     otel_traces_exporter: KarapaceTelemetryOTelExporter = KarapaceTelemetryOTelExporter.NOOP
@@ -83,6 +106,7 @@ class Config(BaseSettings):
     server_tls_certfile: str | None = None
     server_tls_keyfile: str | None = None
     server_tls_cafile: str | None = None
+    server_tls_client_auth: ServerTLSClientAuth = ServerTLSClientAuth.NONE
     registry_scheme: str = "http"
     registry_host: str = "karapace-schema-registry"
     registry_port: int = 8081
@@ -196,7 +220,7 @@ class Config(BaseSettings):
 
     def to_env_str(self) -> str:
         env_lines: list[str] = []
-        for key, value in self.model_dump().items():
+        for key, value in self.model_dump(mode="json").items():
             if value is not None:
                 env_lines.append(f"{key.upper()}={value}")
         return "\n".join(env_lines)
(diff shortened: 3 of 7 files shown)


src/karapace/core/config.py
```diff
@@ -392,4 +393,20 @@ def create_server_ssl_context(config: Config) -> ssl.SSLContext | None:
...
     ssl_context.load_cert_chain(certfile=tls_certfile, keyfile=tls_keyfile)
+
+    client_auth_mode = config.server_tls_client_auth.lower()
+    if client_auth_mode not in ("none", "optional", "required"):
+        raise InvalidConfiguration("`server_tls_client_auth` must be one of: none, optional, required")
+
+    if config.server_tls_cafile:
+        ssl_context.load_verify_locations(config.server_tls_cafile)
+
+    if client_auth_mode == "required":
```
nosahama: **[nit]**
you could use `match/case`

src/karapace/core/config.py
```diff
@@ -83,6 +83,7 @@ class Config(BaseSettings):
     server_tls_certfile: str | None = None
     server_tls_keyfile: str | None = None
     server_tls_cafile: str | None = None
+    server_tls_client_auth: Literal["none", "optional", "required"] = "none"
```
nosahama: Can we extract `"none", "optional", "required"` into an enum, we can then use that for reference.

src/karapace/__main__.py
```diff
@@ -97,6 +98,14 @@
 
     config = karapace_container.config()
     app = create_karapace_application(config=config, lifespan=karapace_schema_registry_lifespan)
+    client_auth_mode = config.server_tls_client_auth.lower()
+    if client_auth_mode == "required":
```
nosahama: **[nit]**
maybe you can extract this logic into a function and then use that in `create_server_ssl_context()` and in this file.

tests/integration/test_mtls.py
```diff
@@ -0,0 +1,110 @@
...
+import aiohttp
+from aiohttp.client_exceptions import ServerDisconnectedError
+import pytest
+import ssl
+
+from karapace.core.config import Config
+from tests.integration.utils.cluster import start_schema_registry_cluster
+
+
+@pytest.mark.asyncio
```
nosahama, line 19: **[nit]**
you do not need the pytest asyncio marker.

tests/integration/test_mtls.py
```diff
@@ -0,0 +1,110 @@
...
+async def test_mtls_requires_client_certificate(
+    tmp_path,
+    kafka_servers,
+    server_ca: str,
+    server_cert: str,
+    server_key: str,
+) -> None:
+    # Skip cleanly if test credentials are not available in the environment.
```
nosahama: When is this the case?

tests/integration/test_mtls.py
```diff
@@ -0,0 +1,110 @@
...
+
+        async with aiohttp.ClientSession() as session:
+            failed = False
+            resp = None
+            try:
+                resp = await session.get(f"{endpoint}/subjects", ssl=ssl_ctx)
+                if resp.status >= 400:
+                    failed = True
+            except (
+                aiohttp.ClientConnectorCertificateError,
```
nosahama: Are we expecting such an exhaustive list of errors, just from this single API call?

tests/integration/test_mtls.py
muralibasani: Removed.

tests/integration/test_mtls.py
muralibasani: Actually not, replaced with only client and ssl error.

tests/integration/utils/cluster.py
```diff
@@ -66,6 +67,12 @@ async def start_schema_registry_cluster(
             logfile = stack.enter_context(open(log_path, "w"))
             errfile = stack.enter_context(open(error_path, "w"))
 
+            client_auth = (
+                config.server_tls_client_auth.value
+                if isinstance(config.server_tls_client_auth, enum.Enum)
```
nosahama, line 72: Is this really needed?

muralibasani: its good to keep it imo, as it normalizes the value just in case if it is not enum rather its a raw string. doesn't hurt

muralibasani: But it also looks redundant, I can remove it

muralibasani: Removed.

src/karapace/core/config.py
```diff
@@ -45,6 +45,29 @@ class KarapaceTelemetryOTelExporter(str, enum.Enum):
...
+) -> ssl.VerifyMode:
+    match client_auth_mode:
+        case ServerTLSClientAuth.REQUIRED:
+            if not has_cafile:
+                raise InvalidConfiguration("`server_tls_client_auth` set to required but `server_tls_cafile` not provided")
+            return ssl.CERT_REQUIRED
+        case ServerTLSClientAuth.OPTIONAL:
+            return ssl.CERT_OPTIONAL
+        case ServerTLSClientAuth.NONE:
+            return ssl.CERT_NONE
```
jclarysse, line 67:
```suggestion
            if has_cafile:
                raise InvalidConfiguration("`server_tls_client_auth` set to none but `server_tls_cafile` provided")
            return ssl.CERT_NONE
```

src/karapace/core/config.py
```diff
@@ -392,4 +416,18 @@ def create_server_ssl_context(config: Config) -> ssl.SSLContext | None:
...
+        client_auth_mode = ServerTLSClientAuth(config.server_tls_client_auth)
+    except ValueError as exc:
+        raise InvalidConfiguration("`server_tls_client_auth` must be one of: none, optional, required") from exc
+
+    if config.server_tls_cafile:
+        ssl_context.load_verify_locations(config.server_tls_cafile)
+
+    ssl_context.verify_mode = get_client_auth_verify_mode(
+        client_auth_mode,
+        has_cafile=bool(config.server_tls_cafile),
```
jclarysse, line 430: For completeness, I wonder if we should also pass `has_certfile=bool(config.server_tls_certfile)` and `has_keyfile=bool(config.server_tls_keyfile)` to the function?

src/karapace/__main__.py
```diff
@@ -97,6 +98,12 @@
 
     config = karapace_container.config()
     app = create_karapace_application(config=config, lifespan=karapace_schema_registry_lifespan)
+    client_auth_mode = ServerTLSClientAuth(config.server_tls_client_auth)
+    ssl_cert_reqs = get_client_auth_verify_mode(
+        client_auth_mode,
+        has_cafile=bool(config.server_tls_cafile),
```
jclarysse: For completeness, I wonder if we should also pass has_certfile=bool(config.server_tls_certfile) and has_keyfile=bool(config.server_tls_keyfile) to the function?

tests/integration/utils/cluster.py
```diff
@@ -85,6 +92,7 @@ async def start_schema_registry_cluster(
                 "KARAPACE_SERVER_TLS_CAFILE": config.server_tls_cafile if config.server_tls_cafile else "",
                 "KARAPACE_SERVER_TLS_CERTFILE": config.server_tls_certfile if config.server_tls_certfile else "",
                 "KARAPACE_SERVER_TLS_KEYFILE": config.server_tls_keyfile if config.server_tls_keyfile else "",
+                "KARAPACE_SERVER_TLS_CLIENT_AUTH": client_auth if client_auth else "",
```
jclarysse:
```suggestion
                "KARAPACE_SERVER_TLS_CLIENT_AUTH": client_auth if client_auth else "none",
```

src/karapace/core/config.py
```diff
@@ -392,4 +416,18 @@ def create_server_ssl_context(config: Config) -> ssl.SSLContext | None:
...
 
     ssl_context.load_cert_chain(certfile=tls_certfile, keyfile=tls_keyfile)
+
+    try:
+        client_auth_mode = ServerTLSClientAuth(config.server_tls_client_auth)
+    except ValueError as exc:
+        raise InvalidConfiguration("`server_tls_client_auth` must be one of: none, optional, required") from exc
+
+    if config.server_tls_cafile:
+        ssl_context.load_verify_locations(config.server_tls_cafile)
```
jclarysse, line 426: Why is this needed here?

README.rst
```diff
@@ -491,6 +491,9 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
    * - ``server_tls_cafile``
      - ``/path/to/cafile``
      - Filename to the SSL CA certificate.
+  * - ``server_tls_client_auth``
```
jclarysse, line 494: Do we have a good reason for calling the Karapace config `server_tls_client_auth` instead of `ssl_cert_reqs`?

jclarysse: I do see that the prefix `server_` aligns pretty well with the related `server_tls_cafile`,`server_tls_certfile`,`server_tls_keyfile`. But in this case we could also call it `server_ssl_cert_reqs`.

src/karapace/core/config.py
muralibasani: I see this as a breaking change. Can we be abit flexible and allow the users to keep the config and just with required or not required use the certs ?

src/karapace/core/config.py
muralibasani: this is required only for mtls when ca file is provided. basically it loads ca, to verify client certs

src/karapace/core/config.py
muralibasani: key file and certfile are already verified in earlier step create_server_ssl_context, so this is redundant imo

src/karapace/__main__.py
muralibasani: Adding a condition to check if key and cert file exists.

README.rst
muralibasani: Usually in applications, this is the standard convention if client auth is required.
For ex just found on internet : https://www.authlete.com/kb/oauth-and-openid-connect/client-authentication/v2/tls-client-auth/ it is tls-client-auth

PR #1187, opened 2026-01-05
Remove pinned versions from toml, and py requires 3.12, 3.13

muralibasani:

- Remove pinned versions from pyproject.toml
- aiokafka, confluent-kafka and opentelemetry-proto are pinned to original versions
- requirements file updated based on pin-requirements
- Fix a bug in make file related to current dir
- Update test suits to py versions 3.12 and 3.13
- Fix typing issues

TODO - follow up pr to remove pinned versions of aiokafka and confluent-kafka

References: #xxxxx

# Why this way


pyproject.toml
```diff
@@ -4,42 +4,42 @@ build-backend = "setuptools.build_meta"
...
-  "zstandard == 0.23.0",
-  "prometheus-client == 0.20.0",
-  "yarl == 1.20.0",
-  "opentelemetry-api == 1.27.0",
-  "opentelemetry-sdk == 1.27.0",
-  "opentelemetry-exporter-otlp == 1.27.0",
-  "dependency-injector == 4.43.0",
+  "accept-types",
+  "aiohttp",
+  "aiokafka==0.11.0",
```
aiven-anton:
```suggestion
  "aiokafka<0.12",
```

suggestion (non-blocking): Consider a softer spec, so that we would still receive patch upgrades for 0.11.x included when freezing.

runtime.txt
```diff
@@ -1 +1 @@
-python-3.10.16
+python-3.12.16
```
aiven-anton: issue (non-blocking): This version doesn't exist. The highest `3.12` version is `3.12.12`.

requirements/requirements.txt
```diff
@@ -27,245 +29,276 @@ async-timeout==5.0.1
...
+ujson==5.11.0
     # via karapace (/karapace/pyproject.toml)
-urllib3==2.5.0
-    # via requests
-uvicorn[standard]==0.35.0
+urllib3==2.6.2
+    # via
+    #   requests
+    #   sentry-sdk
+uvicorn[standard]==0.40.0
```
aiven-anton, line 268: note (non-blocking): An unrelated note to the change at hand, but we should specify `--strip-extras` [here](https://github.com/Aiven-Open/karapace/blob/1a2b23ab642934e964264e5ab194ed9b2cbac80c/GNUmakefile#L14-L16). The reason is because all extras should be fully "unfurled" into the corresponding transitive dependencies here anyway, and there shouldn't be any extra to resolve at the point of installing.

(And similarly, when installing, we should typically specify `--no-dependencies`, such that only exactly what's in the frozen file gets installed).

requirements/requirements.txt
```diff
@@ -27,245 +29,276 @@ async-timeout==5.0.1
...
     #   opentelemetry-exporter-otlp-proto-grpc
     #   opentelemetry-exporter-otlp-proto-http
-opentelemetry-sdk==1.27.0
+opentelemetry-sdk==1.39.1
     # via
     #   karapace (/karapace/pyproject.toml)
     #   opentelemetry-exporter-otlp-proto-grpc
     #   opentelemetry-exporter-otlp-proto-http
-opentelemetry-semantic-conventions==0.48b0
+opentelemetry-semantic-conventions==0.60b1
```
aiven-anton, line 156: note (non-blocking): It looks off that we are installing pre releases, but might be correct.

Update: yeah, it looks like there are only beta releases: https://pypi.org/project/opentelemetry-semantic-conventions/#history.

container/Dockerfile
```diff
@@ -83,6 +82,9 @@ RUN python3 -m pip install -r /opt/karapace/requirements/requirements-dev.txt -r
 COPY . .
 RUN SETUPTOOLS_SCM_PRETEND_VERSION=$KARAPACE_VERSION python3 -m pip install .
 ENV PYTHONPATH="/opt/karapace/src:$PYTHONPATH"
+# Set protobuf to use pure Python implementation for compatibility with protoc-generated code
+# This is required for protobuf 5.0+ which has breaking changes
+ENV PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
```
aiven-anton: issue (non-blocking): I think we likely need to hold this one back, IIRC it's not trivial to upgrade this. It's likely a better idea to do that only after we have fully phased out Karapace v4.

pyproject.toml
```diff
@@ -4,42 +4,43 @@ build-backend = "setuptools.build_meta"
...
-  "zstandard == 0.23.0",
-  "prometheus-client == 0.20.0",
-  "yarl == 1.20.0",
-  "opentelemetry-api == 1.27.0",
-  "opentelemetry-sdk == 1.27.0",
-  "opentelemetry-exporter-otlp == 1.27.0",
-  "dependency-injector == 4.43.0",
+  "accept-types",
+  "aiohttp",
+  "aiokafka==0.10.0",
```
muralibasani, line 14: keeping original versions. in follow up pr, this will be removed.

pyproject.toml
```diff
@@ -4,42 +4,43 @@ build-backend = "setuptools.build_meta"
...
+  "async_lru",
+  "cachetools",
+  "confluent-kafka==2.10.0",
```
muralibasani, line 17: keeping original version. in follow up pr, this will be removed.

pyproject.toml
```diff
@@ -4,42 +4,43 @@ build-backend = "setuptools.build_meta"
...
+  "ujson",
+  "watchfiles",
+  "xxhash",
+  "zstandard",
+  "prometheus-client",
+  "yarl",
+  "opentelemetry-api",
+  "opentelemetry-sdk",
+  "opentelemetry-exporter-otlp",
+  "opentelemetry-proto<1.28.0",
```
muralibasani, line 42: keeping it compatible with proto 3.20.3

pyproject.toml
```diff
@@ -4,42 +4,43 @@ build-backend = "setuptools.build_meta"
...
+  "async_lru",
+  "cachetools",
+  "confluent-kafka==2.10.0",
+  "cryptography",
+  "fastapi[standard]",
+  "isodate",
+  "jsonschema",
+  "lz4",
+  "networkx",
+  "protobuf==3.20.3",
```
muralibasani, line 24: in follow up pr, this will be removed

nosahama (approved):

PR #1191, opened 2026-01-08
Upgrade confluent kafka / librd

muralibasani:

- Remove pinned version confluent-kafka
- Upgrades confluent-kafka to 2.13.0
- Upgrades aiokafka to 0.12.0 (0.13.0 changes are breaking bad. Will be tried in follow up pr)
- Updated consumer and consumer_manager
- Updated schema reader - subscribing to a non-existent topic
- Fix typing and mypy issues
- Delete stubs dir and using the typings from https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/cimpl.pyi.

References: #xxxxx

# Why this way


src/karapace/backup/backends/v3/backend.py
```diff
@@ -337,7 +337,8 @@ def store_record(
         buffer: IO[bytes],
         record: Message,
     ) -> None:
-        stats: Final = self._partition_stats[record.partition()]
+        partition: int = record.partition()  # type: ignore[assignment]
```
muralibasani, line 340: changes due to upgraded confluent-kafka

src/karapace/core/kafka/admin.py
```diff
@@ -147,15 +150,37 @@ def get_topic_config(
         except KafkaException as exc:
             raise_from_kafkaexception(exc)
 
-        config: dict[str, str] = {}
+        # Use a dict to track config entries, prioritizing dynamic over static/default
+        # ConfigSource priority: DYNAMIC_TOPIC_CONFIG > DYNAMIC_BROKER_CONFIG > STATIC_BROKER_CONFIG > DEFAULT_CONFIG
+        config_entries: dict[str, tuple[ConfigSource, str]] = {}
```
muralibasani, line 155: With confluent-kafka 2.13+, when you call describe_configs(), it returns multiple entries

src/karapace/core/kafka/admin.py
```diff
@@ -147,15 +150,37 @@ def get_topic_config(
...
-                config_source_filter is None or ConfigSource(config_entry.source) in config_source_filter
-            )
+            config_source = ConfigSource(config_entry.source)
+            matches_source_filter: bool = config_source_filter is None or config_source in config_source_filter
 
             if matches_name_filter or matches_source_filter:
-                config[config_name] = config_entry.value
+                # If we already have an entry for this config name, check if we should replace it
+                if config_name in config_entries:
+                    existing_source, _ = config_entries[config_name]
```
muralibasani, line 164: Priority - 
DYNAMIC_TOPIC_CONFIG (highest priority - this is what we set via API)
DYNAMIC_BROKER_CONFIG
STATIC_BROKER_CONFIG
DEFAULT_CONFIG (lowest priority - Kafka defaults)

src/karapace/core/kafka/consumer.py
```diff
@@ -238,6 +238,10 @@ async def poll(self, timeout: float) -> Message | None:
         assert self.consumer is not None, self._START_ERROR
         return await self._run_in_executor(self.consumer.poll, timeout)
 
+    async def consume(self, num_messages: int = 1, timeout: float = -1) -> list[Message]:
```
muralibasani, line 241: adding this wrapper consume method to efficiently handle consuming messages. earlier the fetch() method was using poll() in a loop to fetch messages one at a time

src/karapace/core/schema_reader.py
```diff
@@ -252,6 +253,26 @@ def run(self) -> None:
                     LOG.exception("[Schema Topic] Failed to create %r, retrying", self.config.topic_name)
                     self._stop_schema_reader.wait(timeout=SCHEMA_TOPIC_CREATION_TIMEOUT_SECONDS)
 
+            # Subscribe to topic AFTER it has been created
+            # This is critical with confluent-kafka 2.13+ where subscribing to a non-existent topic
```
muralibasani, line 257: changes due to confluentkafka 2.13

pyproject.toml
```diff
@@ -11,10 +11,10 @@ license = {file = "LICENSE"}
 dependencies = [
   "accept-types",
   "aiohttp",
-  "aiokafka==0.10.0",
+  "aiokafka<0.13.0",
```
aiven-anton, line 14: How come we don't go to latest? Something breaking in `0.13.0`?

muralibasani: I was gradually making changes bumping versions and now that confluent-kafka is already on 2.13 latest, just pushed this suggestion.
Not many changes in requirements file. Think it should be good.

src/karapace/core/kafka/admin.py
```diff
@@ -187,7 +212,7 @@ def get_offsets(self, topic: str, partition_id: int) -> dict[str, int]:
             raise_from_kafkaexception(exc)
         return {"beginning_offset": startoffset.offset, "end_offset": endoffset.offset}
 
-    def describe_topics(self, topics: TopicCollection) -> dict[str, Future]:
+    def describe_topics(self, topics: TopicCollection) -> dict[str, Future]:  # type: ignore[override]
```
aiven-anton: What is the type annotation of `super().describe_topics()`? Since we return that, I think we should be able to have compatible typing here?

src/karapace/core/kafka/producer.py
```diff
@@ -54,7 +54,7 @@ def send(self, topic: str, **params: Unpack[ProducerSendParams]) -> Future[Messa
             self.produce(
                 topic,
                 on_delivery=partial(_on_delivery_callback, result),
-                **params,
+                **params,  # type: ignore[arg-type]
```
aiven-anton, line 57: issue: How does this start to fail? (What's the error?) Does `ProducerSendParams` need updating?

stubs/confluent_kafka/cimpl.pyi
aiven-anton, line 1: issue: It looks like this dependency now ships with type annotations. That means we should stop implementing them ourselves: we should delete `stubs/confluent_kafka/`, and rather use the upstream (builtin) types for type checking.

pyproject.toml
muralibasani: Remember now. 
aiokafka 0.13 is breaking bad with never ending errors. Will try in follow up pr.

starts like this, then changes to consumer manager and so on. 
/karapace/core/coordinator/master_coordinator.py", line 156, in get_coordinator_status
    assert self._sc is not None
           ^^^^^^^^^^^^^^^^^^^^

aiven-anton: I see, could you add a comment above this line, explaining shortly why hold it back at `<0.13.0`?

requirements/requirements-dev.txt
```diff
@@ -10,7 +10,7 @@ aiohappyeyeballs==2.6.1
     # via aiohttp
 aiohttp==3.13.3
     # via karapace (/karapace/pyproject.toml)
-aiokafka==0.10.0
+aiokafka==0.12.0
```
muralibasani, line 13: aiokafka 0.13 needs several changes. Will try in follow up pr.

starts like this, then changes to consumer manager and so on.
/karapace/core/coordinator/master_coordinator.py", line 156, in get_coordinator_status
assert self._sc is not None
^^^^^^^^^^^^^^^^^^^^

src/karapace/core/kafka/producer.py
muralibasani: so we should update ProducerSendParams to narrow the types but that breaks existing code. hence this type ignore.

src/karapace/core/kafka/admin.py
muralibasani: it is Future[TopicMetadata]. I am updating stub as well now, but with type ignore.

src/karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -504,38 +504,67 @@ async def fetch(self, internal_name: tuple[str, str], content_type: str, formats
...
             )
             read_bytes = 0
             start_time = time.monotonic()
-            poll_data = []
-            message_count = 0
-            read_buffered = True
-            while read_bytes < max_bytes and (start_time + timeout / 1000 > time.monotonic() or read_buffered):
-                read_buffered = False
-                time_left = start_time + timeout / 1000 - time.monotonic()
+            end_time = start_time + timeout / 1000
```
muralibasani, line 513: The old code was using poll() in a loop, fetching one message at a time. This change with efficient bulk fetching - gets multiple messages per call with the new consume method.

stubs/confluent_kafka/cimpl.pyi
muralibasani: Removed stubs and added types-confluent dependency. That worked.

muralibasani: Using the types from available dependency - https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/cimpl.pyi

HelenMel:
@muralibasani please, ignore my approval, I was just checking that I still have a read-only permissions

pyproject.toml
nosahama: **[non-blocking]**
I was under the impression that we will remove the deps from this file, that of which still baffles me, but yh we have the tooling around this.

src/karapace/kafka_rest_apis/consumer_manager.py
nosahama: [discussion, non-blocking]
There's a lot of calculation going on in this logic now, I would think that the `consume()` function should handle byte cap. 
I'd be keen to test the startup against a topic with a lot of schemas.

src/karapace/core/schema_reader.py
nosahama: Cool, might take away some logs i usually see.

src/karapace/core/kafka/admin.py
```diff
@@ -172,23 +197,25 @@ def get_offsets(self, topic: str, partition_id: int) -> dict[str, int]:
             latest_offset_futmap: dict[TopicPartition, Future] = self.list_offsets(
                 {
                     TopicPartition(topic, partition_id): OffsetSpec.latest(),
-                }
+                },
+                isolation_level=IsolationLevel.READ_UNCOMMITTED,
```
nosahama, line 201: **[discussion]**
why this choice?

src/karapace/core/kafka/admin.py
nosahama: [non-blocking]
A test for this and some other things will be nice, but it can be follow up 😄

src/karapace/backup/api.py
```diff
@@ -318,7 +318,10 @@ def _consume_records(
             raise translate_from_kafkaerror(error)
 
         yield record
-        last_offset = record.offset()
+        record_offset = record.offset()
+        if record_offset is None:
```
nosahama, line 322: 👍 , although we'd deprecate backup soon, idea is solid.

nosahama (approved):
Left some comments, follow up in another PR if needed, we can still discuss here if you like. 
There were also a lot more type-ignore added, but i understand how painful it can be.

pyproject.toml
aiven-anton: > I was under the impression that we will remove the deps from this file

We are only removing the specific version pins from this file, because the freezing process is already giving us pins (in the requirements.txt file). Sometimes it's still valid to put pins in this file, but they should typically be more inclusive. For example, `aiokafka<0.13.0` is preferable over `aiokafka==0.12.0`, because if there's a patch version `0.12.1` released, the former definition would use that.

Typically a valid reason is discovering that there is some breakage in a new version. All such things should have a comment though, so that a future reader understands why a dependency is held back. And we should over time fix all such breakages and work to keep the dependencies up to date.

It's also important to note that `pyproject.toml` should _only_ specify top-level dependencies, and not any transitive dependencies. Those also gets resolved in the freezing process, and are pinned in requirements.txt.

src/karapace/core/kafka/admin.py
muralibasani: Here in this get_offsets method, we are querying the offset positions but not reading data as such. Other option READ_COMMITTED is only the latest offset (for a consumer) but we wanted the offset range.

pyproject.toml
nosahama: I've enjoyed [poetry](https://python-poetry.org/) so far, but always happy to find out other ways, thanks for the insights.

PR #1196, opened 2026-01-19
Preempt topic creation action with metadata list

rwaweber:
In an ACL-enabled kafka environment trying to get karapace running as a schema registry, we ran into this hiccup since we aren't giving karapace full admin permissions.

Tracing down the topic authorization failure we were getting, it looks like karapace attempts topic creation before checking if the topic exists, then takes action based on that failure.

Instead, I think it makes sense to check the topic metadata before attempting to create the topic, as that's possible to do with a smaller, more scoped set of permissions.

- Perform a topic list that match the config topic name before we try to create the schema storage topic

References: N/A (I figured a small patch may set the scene better than a bug report, but happy to create one if you all prefer!)

# Why this way

In ACL-enabled kafka clusters, karapace would require a "CreateTopics" permission on the "Cluster" resource to complete this operation as-is. With these changes, the karapace identity only requires a "Describe" on a "Topic" resource with the "Metadata" API, which is easier to scope for an administrator perspective.

That being said, happy to consider alternative approaches, as I'm still pretty unfamiliar with this codebase at large!

### DIFF
--- a/src/karapace/core/schema_reader.py
+++ b/src/karapace/core/schema_reader.py
@@ -239,11 +239,13 @@ class KafkaSchemaReader(Thread, SchemaReaderStoppper):
 
             # Perform a metadata list, which is allowed in a describe permission, whereas
             # topic-create permissions are a coarser permission grant, not typically given out
+            #
+            # check if schema topic is already created, if not try to create it
             try:
                 schema_topic_exists = (
                     self.config.topic_name in self.admin_client.list_topics().topics.keys()
                 )
-            except TopicAuthorizationFailedError:
+            except Exception:
                 schema_topic_exists = False
                 LOG.warning(
                     "[Schema Topic] not authorized to list topics, assuming topic: %r does not exist",
--- a/src/karapace/core/schema_reader.py
+++ b/src/karapace/core/schema_reader.py
@@ -240,7 +240,7 @@ class KafkaSchemaReader(Thread, SchemaReaderStoppper):
             # Perform a metadata list, which is allowed in a describe permission, whereas
             # topic-create permissions are a coarser permission grant, not typically given out
             schema_topic_exists = (
-                self.config.topic_name in self.admin_client.list_topics(topic=self.config.topic_name).topics.keys()
+                self.config.topic_name in self.admin_client.list_topics().topics.keys()
             )
             while not self._stop_schema_reader.is_set() and not schema_topic_exists:
                 try:
@@ -350,7 +350,7 @@ class KafkaSchemaReader(Thread, SchemaReaderStoppper):
         except KafkaTimeoutError:
             LOG.warning("Reading begin offsets timed out.")
         except UnknownTopicOrPartitionError:
-            LOG.warning("Topic does not yet exist.")
+            LOG.warning("Topic: %r does not yet exist.", self.config.topic_name)
         except (LeaderNotAvailableError, NotLeaderForPartitionError):
             LOG.warning("Retrying to find leader for schema topic partition.")
         except Exception as e:
--- a/src/karapace/core/schema_reader.py
+++ b/src/karapace/core/schema_reader.py
@@ -237,7 +237,9 @@ class KafkaSchemaReader(Thread, SchemaReaderStoppper):
 
             assert self.consumer is not None
 
-            schema_topic_exists = False
+            # Perform a metadata list, which is allowed in a describe permission, whereas
+            # topic-create permissions are a coarser permission grant, not typically given out
+            schema_topic_exists = self.config.topic_name in self.admin_client.list_topics(topic=self.config.topic_name).topics.keys()
             while not self._stop_schema_reader.is_set() and not schema_topic_exists:
                 try:
                     LOG.debug("[Schema Topic] Creating %r", self.config.topic_name)
--- a/src/karapace/core/schema_reader.py
+++ b/src/karapace/core/schema_reader.py
@@ -160,6 +160,9 @@ class KafkaSchemaReader(Thread, SchemaReaderStoppper):
         self.kafka_error_handler: KafkaErrorHandler = KafkaErrorHandler(config=config)
         self._tracer = Tracer()
 
+        # indicate whether karapace created the schema_topic
+        self._karapace_created_schema_topic: bool = False
+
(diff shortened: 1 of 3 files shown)


muralibasani:
@rwaweber thanks for the pr.
Can you fix lint (pre-commit run --all-files) and you may run smoke tests locally too

rwaweber:
Hey @muralibasani  - so trying to run the mypy check locally, it seems like it does a bit more than just linting with the `make type-check-mypy-in-docker` command.

Looking at [this line](https://github.com/Aiven-Open/karapace/actions/runs/21582810523/job/62183801099?pr=1196#step:7:1695) in the GHA run, I can't seem to suss out what exactly is going wrong in the docker compose setup. Do you have any suggestions on where I should be digging around to see?

src/karapace/core/schema_reader.py
```diff
@@ -228,7 +228,9 @@ def run(self) -> None:
 
             assert self.consumer is not None
 
-            schema_topic_exists = False
+            # Perform a metadata list, which is allowed in a describe permission, whereas
+            # topic-create permissions are a coarser permission grant, not typically given out
+            schema_topic_exists = self.config.topic_name in self.admin_client.list_topics().topics.keys()
```
muralibasani: If list_topics() fails because of any permission issues (Describe for ex), then we might run into failures.
Hence a try/catch with `except TopicAuthorizationFailedError:` ?

rwaweber: Sounds good! In that instance, I'm thinking we handle the exception, set `schema_topic_exists = False` and then, just continue on or we `raise` to shortcircuit?

I'm just thinking practically, where one would grant a permission to be able to _create_ a topic but deny listing topics on the cluster seems unlikely, but I could be oversimplifying.

src/karapace/core/schema_reader.py
```diff
@@ -228,7 +228,18 @@ def run(self) -> None:
...
+            try:
+                schema_topic_exists = (
+                    self.config.topic_name in self.admin_client.list_topics().topics.keys()
+                )
+            except TopicAuthorizationFailedError:
+                schema_topic_exists = False
+                LOG.warning(
+                    "[Schema Topic] not authorized to list topics, assuming topic: %r does not exist",
+                    self.config.topic_name
+                )
```
muralibasani, line 242: Let's add broader exception catch block too

```
except Exception:                                                                                                                 
      schema_topic_exists = False                                                                                                   
      LOG.warning(                                                                                                                  
          "[Schema Topic] failed to list topics, assuming topic: %r does not exist",                                                
          self.config.topic_name                                                                                                    
      )    
```

src/karapace/core/schema_reader.py
```diff
@@ -228,7 +228,18 @@ def run(self) -> None:
 
             assert self.consumer is not None
 
-            schema_topic_exists = False
+            # Perform a metadata list, which is allowed in a describe permission, whereas
+            # topic-create permissions are a coarser permission grant, not typically given out
+            try:
+                schema_topic_exists = (
```
muralibasani, line 234: Can we add a test for this extra check?

rwaweber: Defiunitely! Doing a pass at the testing setup, I'm thinking it might make sense as an integration test, like:

- We test for when the schema_topic does exist, it should not attempt to create it
- We test for when the schema_topic doesnt exist, it _should_ try to create it

I'm thinking this might end up in `./tests/integration/test_schema_reader.py`? If so, you think this warrants its own separate test?

muralibasani: I think test_schema_reader.py is good yea with a separate test. Thanks for looking into this.

rwaweber:
Hey @muralibasani! Friendly ping, finally got around to adding some tests last week! They're definitely kinda long in comparison to the actual code that they're exercising, so would be happy to shrink them down a bit if you have some ideas. Was drawing pretty hard from the other tests in the file.

I added two other changes while I was in here that I'm happy to pull out to either a separate review or drop altogether. Those being these two commits: https://github.com/Aiven-Open/karapace/pull/1196/changes/3bcf8645849e242f46e204cf4127fe8d7a545b10, https://github.com/Aiven-Open/karapace/pull/1196/changes/4b0c69606aafcd50f4021e360edefa3f9c156028

rwaweber:
hey @muralibasani  - its a little tough to tell what failed in the smoke tests on 3.12 and 3.14, but it passed on 3.13. Going off the logs on those runs, they don't _seem_ related to the changes that I made, but I could be off base. Maybe a re-run will help?

src/karapace/core/schema_reader.py
```diff
@@ -237,7 +240,17 @@ def run(self) -> None:
 
             assert self.consumer is not None
 
-            schema_topic_exists = False
+            # Perform a metadata list, which is allowed in a describe permission, whereas
+            # topic-create permissions are a coarser permission grant, not typically given out
+            #
+            # check if schema topic is already created, if not try to create it
+            try:
+                schema_topic_exists = self.config.topic_name in self.admin_client.list_topics().topics.keys()
```
muralibasani, line 248: Could this use self.admin_client.list_topics(topic=self.config.topic_name) instead of the unscoped list_topics() on full cluster.

src/karapace/core/schema_reader.py
```diff
@@ -237,7 +240,17 @@ def run(self) -> None:
...
+            except Exception:
+                schema_topic_exists = False
+                LOG.warning(
```
muralibasani: can we log the actual exception? Now for all exceptions, it is assuming authorization issue.

src/karapace/core/schema_reader.py
rwaweber: See my [earlier comment](https://github.com/Aiven-Open/karapace/pull/1196#issuecomment-4001812867), but due to some unintuitive behavior in the confluent-kafka python client, we end up creating the topic if we specify it there when `auto.topics.create.enable=true` is set on the broker. As a result, the unscoped version, while less precise, has fewer unintended side effects.

(that behavior caused me a lot of early confusion with this PR, since that setting is enabled in the cluster that comes up with the tests)

PR #1226, opened 2026-03-07
Add sasl_oauth_token_provider_class for native OAUTHBEARER support

dylanbstorey:
## Summary

- Adds `sasl_oauth_token_provider_class` config option that accepts a Python import path (via pydantic `ImportString`) to a class implementing the `TokenWithExpiryProvider` protocol
- When configured, an instance is created and passed as `sasl_oauth_token_provider` to all Kafka client factories — admin, consumer, and producer — including the schema reader's internal clients
- Enables native OAUTHBEARER auth (e.g. AWS MSK IAM) without monkey-patching

### Problem

`kafka_utils.py` and `schema_reader.py` create Kafka clients but never pass `sasl_oauth_token_provider`, even though `_KafkaConfigMixin` already supports it via `KafkaClientParams` → `oauth_cb`. Anyone deploying Karapace against AWS MSK with IAM auth must monkey-patch the client factories or build a custom wrapper.

### Solution

New config option configurable via env var:
```
KARAPACE_SECURITY_PROTOCOL=SASL_SSL
KARAPACE_SASL_MECHANISM=OAUTHBEARER
KARAPACE_SASL_OAUTH_TOKEN_PROVIDER_CLASS=examples.msk_iam_token_provider:MSKIAMTokenProvider
```

The referenced class must implement:
```python
def token_with_expiry(self, config: str | None) -> tuple[str, int | None]
```

### Changes

- **`src/karapace/core/config.py`**: Add `sasl_oauth_token_provider_class: ImportString | None` field
- **`src/karapace/core/kafka_utils.py`**: Instantiate provider and pass to all 3 factory functions
- **`src/karapace/core/schema_reader.py`**: Same for the 2 internal factory functions
- **`examples/msk_iam_token_provider.py`**: Example AWS MSK IAM token provider using `aws-msk-iam-sasl-signer-python`
- **`tests/unit/test_oauth_token_provider.py`**: 14 unit tests covering config, all 5 factories (with/without provider)

## Test plan

- [x] 14 unit tests pass locally (`tests/unit/test_oauth_token_provider.py`)
- [x] Existing tests unaffected (no behavior change when option is unset)

### DIFF
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -1,0 +1,104 @@
+OAUTHBEARER Authentication
+==========================
+
+Karapace supports pluggable OAUTHBEARER token providers for authenticating to
+Kafka brokers. This is useful for deployments using AWS MSK with IAM
+authentication, or any Kafka cluster that uses the OAUTHBEARER SASL mechanism.
+
+How it works
+------------
+
+Set the ``sasl_oauth_token_provider_class`` config option to a Python import
+path pointing to your token provider class. Karapace will import and
+instantiate the class, then pass it to every Kafka client (admin, consumer,
+producer) it creates — including the schema reader's internal clients.
+
+Your class must implement a single method::
+
+    def token_with_expiry(self, config: str | None = None) -> tuple[str, int | None]:
+        """Return (token_string, expiry_epoch_timestamp_or_None)."""
+
+This matches the ``TokenWithExpiryProvider`` protocol defined in
+``karapace.core.kafka.common``.
+
+Configuration
+-------------
+
+Three config values are needed:
+
+.. code-block:: bash
+
+    # The Kafka security protocol
+    KARAPACE_SECURITY_PROTOCOL=SASL_SSL
+
+    # The SASL mechanism
+    KARAPACE_SASL_MECHANISM=OAUTHBEARER
+
+    # Python import path to your token provider class
+    KARAPACE_SASL_OAUTH_TOKEN_PROVIDER_CLASS=my_module:MyTokenProvider
+
+The import path format is ``dotted.module.path:ClassName``. Karapace uses
+pydantic's ``ImportString`` to resolve it, so standard Python import rules
+apply.
+
+Loading a provider that is not in the Karapace package
+------------------------------------------------------
+
+Your token provider class does **not** need to live inside the Karapace
+repository. It just needs to be importable by Python at runtime. Here are
+common approaches:
+
+**Option 1: Install as a Python package**
+
+Package your provider and install it into the same environment as Karapace:
+
+.. code-block:: bash
+
+    pip install my-msk-auth
+    # or for local development:
+    pip install -e /path/to/my-msk-auth
+
+Then reference it by its package name:
+
+.. code-block:: bash
+
+    KARAPACE_SASL_OAUTH_TOKEN_PROVIDER_CLASS=my_msk_auth.provider:MSKIAMTokenProvider
+
+**Option 2: Add to PYTHONPATH**
+
+Place your provider file anywhere on disk and add its parent directory to
+``PYTHONPATH``:
+
+.. code-block:: bash
+
+    export PYTHONPATH=/opt/custom:$PYTHONPATH
+    export KARAPACE_SASL_OAUTH_TOKEN_PROVIDER_CLASS=msk_iam_token_provider:MSKIAMTokenProvider
+
+Where ``/opt/custom/msk_iam_token_provider.py`` contains your class.
+
+**Option 3: Docker — mount and extend the path**
+
+.. code-block:: dockerfile
+
+    COPY msk_iam_token_provider.py /opt/custom/
+    ENV PYTHONPATH=/opt/custom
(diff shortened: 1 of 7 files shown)


src/karapace/core/kafka_utils.py
```diff
@@ -12,8 +12,15 @@
 import contextlib
 
 
+def _get_oauth_token_provider(config: Config) -> object | None:
```
muralibasani: This method seems to be defined twice. Can it be imported in schema_reader ?

src/karapace/core/kafka_utils.py
```diff
@@ -12,8 +12,15 @@
...
+    """Instantiate the configured OAuth token provider, if any."""
+    if config.sasl_oauth_token_provider_class is not None:
+        return config.sasl_oauth_token_provider_class()
```
muralibasani: Looks like this get instantiated for every client call ? can it be instantiated once and store in Config may be ?

examples/msk_iam_token_provider.py
```diff
@@ -0,0 +1,88 @@
+"""Example: AWS MSK IAM token provider for Karapace OAUTHBEARER auth.
+
+Copyright (c) 2024 Aiven Ltd
```
muralibasani:
```suggestion
Copyright (c) 2026 Aiven Ltd
```

src/karapace/core/schema_reader.py
```diff
@@ -89,12 +89,19 @@ class MessageType(Enum):
     no_operation = "NOOP"
 
 
+def _get_oauth_token_provider(config: Config) -> object | None:
+    """Instantiate the configured OAuth token provider, if any."""
+    if config.sasl_oauth_token_provider_class is not None:
+        return config.sasl_oauth_token_provider_class()
```
muralibasani: can we check for token_with_expiry ?

tests/unit/test_oauth_token_provider.py
```diff
@@ -0,0 +1,199 @@
+"""
+Tests for sasl_oauth_token_provider_class config and passthrough.
+
+Copyright (c) 2023 Aiven Ltd
```
muralibasani:
```suggestion
Copyright (c) 2026 Aiven Ltd
```

muralibasani (changes requested):
@dylanbstorey thanks for the pr. Have a few comments left.

dylanbstorey:
@muralibasani - just bumping this. Thanks !

src/karapace/core/kafka_utils.py
```diff
@@ -12,8 +12,32 @@
...
+
+    The provider class is instantiated once and cached for the lifetime of the
+    process.  The instance must expose a ``token_with_expiry`` method as
+    required by confluent-kafka's OAUTHBEARER flow.
+    """
+    if config.sasl_oauth_token_provider_class is None:
+        return None
+    cls = config.sasl_oauth_token_provider_class
+    if cls not in _oauth_token_provider_cache:
+        instance = cls()
```
muralibasani: if there are concurrent threads accessing this def, would we be having multiple instances of this class ? Is it better to add a lock or would like to instantiate this class during startup just once for all?

src/karapace/core/kafka_utils.py
```diff
@@ -12,8 +12,32 @@
 import contextlib
 
 
+_oauth_token_provider_cache: dict[type, object] = {}
```
muralibasani:
```suggestion
_oauth_token_provider_cache: dict[type, TokenWithExpiryProvider] = {}
```

is this better with mypy & typing ?

src/karapace/core/kafka_utils.py
dylanbstorey: went with the singleton.

dylanbstorey:
@muralibasani - thanks for the review !

examples/README.rst
```diff
@@ -0,0 +1,104 @@
...
+producer) it creates — including the schema reader's internal clients.
+
+Your class must implement a single method::
+
+    def token_with_expiry(self, config: str | None = None) -> tuple[str, int | None]:
+        """Return (token_string, expiry_epoch_timestamp_or_None)."""
+
+This matches the ``TokenWithExpiryProvider`` protocol defined in
+``karapace.core.kafka.common``.
+
```
muralibasani, line 23: Is it better to mention that this provider protocol is specific to librdkafka/[confluent-kafka](https://github.com/Aiven-Open/karapace/pull/1226/changes#diff-58a7cc9913b8a991a33d6b427eebc2d2904620e66d524f0791d919a0f39bcb62R20) and doesn't work with aiokafka and kafka-python, because AbstractTokenProvider returns string and token expiry/token generation wouldn't work. wdyt ?

dylanbstorey: good doc call out to avoid sharp edges.

muralibasani (approved):
Thank you for the PR and all the changes.

dylanbstorey:
thanks for the thoughtful review !

PR #1233, opened 2026-03-14
Support for Avro logical types

e11it:
Support for Avro logical types:

- timestamp-millis
- timestamp-micros
- date
- time-millis
- time-micros
- decimal

New config flag:
rest_avro_extended_json_parser: bool = False  # default: Confluent-compatible behaviour

When True:
- logical types(timestamp, date, time) also accept ISO 8601 strings in addition to integers
- allow to pass Decimal as int or string ("1234.34")


src/karapace/core/serialization.py
```diff
@@ -477,8 +551,18 @@ def write_value(config: Config, schema: TypedSchema, bio: io.BytesIO, value: dic
         # Backwards compatibility: Support JSON encoded data without the tags for unions.
         if avro.io.validate(schema.schema, value):
             data = value
+
+        # First, try to convert logical types on the original value. If the resulting
+        # value validates against the schema, use it as-is to preserve backwards
+        # compatibility with existing union encodings. Otherwise, fall back to
+        # flattening unions and then converting logical types.
+        converted = convert_logical_types(schema.schema, value)
+        if avro.io.validate(schema.schema, converted):
```
muralibasani: this condition `if avro.io.validate` at line 552 looks redundant then ?

src/karapace/core/serialization.py
```diff
@@ -477,8 +551,18 @@ def write_value(config: Config, schema: TypedSchema, bio: io.BytesIO, value: dic
...
+            data = converted
         else:
             data = flatten_unions(schema.schema, value)
+            flattened = flatten_unions(schema.schema, value)
+            data = convert_logical_types(schema.schema, flattened)
```
muralibasani: do we need this declaration ?

src/karapace/core/serialization.py
```diff
@@ -450,6 +452,78 @@ def get_name(obj) -> str:
     return value
 
 
+def convert_logical_types(schema: avro.schema.Schema, value: Any) -> Any:
```
muralibasani: I think this needs a docstring explaining its purpose, which types it handles, etc

src/karapace/core/serialization.py
```diff
@@ -450,6 +452,78 @@ def get_name(obj) -> str:
...
+
+    if isinstance(schema, avro.schema.MapSchema) and isinstance(value, dict):
+        return {k: convert_logical_types(schema.values, v) for (k, v) in value.items()}
+
+    if isinstance(schema, avro.schema.LogicalSchema):
+        logical_type = getattr(schema, "logical_type", None)
+
+        # Timestamps
+        if logical_type == "timestamp-millis" and isinstance(value, int):
+            epoch = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
```
muralibasani: nit - these constants are recreated on every call.

PR #1256, opened 2026-04-10
Add startup warning when REST proxy authorization is disabled 

muralibasani:

Add startup warning when REST proxy authorization is disabled                                                                                  

The REST proxy's rest_authorization config defaults to false, which causes all Kafka ACLs to be bypassed for REST proxy requests. Operators following the default configuration may not realize their Kafka ACLs are not being enforced.

References: #xxxxx

# Why this way


tvainika (dismissed):
LGTM

tvainika (approved):

PR #1261, opened 2026-04-22
Fix: Name strategy validation bypass for value_schema

### LINKED ISSUE #1260: REST Proxy: schema-to-subject validation can be bypassed with name_strategy_validation=true
When producing Avro messages through Karapace REST Proxy with `name_strategy_validation=true`, schema-to-subject validation can be bypassed in a real user flow.
If I first produce to topic A using a schema (so the proxy has already resolved that schema), then produce to topic B with the same schema, the second produce can succeed even though topic B has no corresponding subject in Schema Registry and should be rejected.
Expected behavior: with `name_strategy_validation=true`, producing to topic B should fail unless the schema is registered under topic B’s subject.  
Actual behavior: after a prior successful produce on another topic, producing to topic B may still succeed, which looks like a cache-related validation gap.

### Example
Schema Registry ACL: the REST user can register subjects only for `topic.v1`.
The request body is the same in all three cases:
```json
{
  "key_schema": "\"string\"",
  "value_schema": "..."
  "records": [...]
```

1. `POST http://localhost:8082/topics/topic.v2` -> Error
```json
{
  "error_code": 40801,
  "message": "Error when registering schema.format = AVRO, subject = topic.v2-value"
}
```
 
2. `POST http://localhost:8082/topics/topic.v1` -> Success
```json
{
  "key_schema_id": 1,
  "offsets": [
    {
      "offset": 5,
      "partition": 0
    }
  ],
  "value_schema_id": 2
}
```
3. `POST http://localhost:8082/topics/topic.v2` -> Unexpected success (should have failed)
```json
{
  "key_schema_id": 1,
  "offsets": [
    {
      "offset": 3,
      "partition": 0
    }
  ],
  "value_schema_id": 2
}
```
Also:
```
curl -u admin:pass localhost:8081/subjects
["topic.v1-key","topic.v1-value"]
```
So no new subjects were registered.

e11it:

Fix #1260 

Cache is still in place: both schemas_to_ids[str(schema)] and topic_schema_cache remain.

For value_schema requests:

1. REST derives the expected subject from the topic (<topic>-value / <topic>-key).
2. It may get a schema_id from cache by schema string.
3. New part: before reusing that cached ID, it checks that this ID is actually linked to the expected subject (via Schema Registry subjects for that ID).
4. If the subject matches, cache hit is accepted.
I5. f it does not match, treat it as a subject-level cache miss and continue normal flow: lookup -> register.

For explicit value_schema_id:
1. Behavior stays strict: if that ID is not linked to the topic subject, return 422 Invalid schema

### DIFF
--- a/src/karapace/core/serialization.py
+++ b/src/karapace/core/serialization.py
@@ -321,13 +321,24 @@ class SchemaRegistrySerializer:
             self.ids_to_schemas[schema_id] = schema
         return schema
 
-    async def upsert_id_for_schema(self, schema_typed: ValidatedTypedSchema, subject: str) -> SchemaId:
+    async def _schema_id_has_subject(self, schema_id: SchemaId, subject: Subject) -> bool:
+        def subject_not_included(_: TypedSchema, subjects: list[Subject]) -> bool:
+            return subject not in subjects
+
+        try:
+            _, subjects = await self.get_schema_for_id(schema_id, need_new_call=subject_not_included)
+        except SchemaRetrievalError:
+            return False
+
+        return subject in subjects
+
+    async def upsert_id_for_schema(self, schema_typed: ValidatedTypedSchema, subject: Subject) -> SchemaId:
         assert self.registry_client, "must not call this method after the object is closed."
 
         schema_ser = str(schema_typed)
-
-        if schema_ser in self.schemas_to_ids:
-            return self.schemas_to_ids[schema_ser]
+        cached_schema_id = self.schemas_to_ids.get(schema_ser)
+        if cached_schema_id is not None and await self._schema_id_has_subject(cached_schema_id, subject):
+            return cached_schema_id
 
         # note: the post is idempotent, so it is like a get or insert (aka upsert)
         schema_id = await self.registry_client.post_new_schema(subject, schema_typed)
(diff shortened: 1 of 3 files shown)


e11it:
With ghcr.io/aiven-open/karapace:latest
```http
# First
POST http://localhost:8082/topics/test.v1
Content-Type: application/vnd.kafka.avro.v2+json
{
  "key_schema": "{\"type\": \"string\"}",
  "value_schema": "{\"namespace\": \"example.avro\", \"type\": \"record\", \"name\": \"simple\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}",
  "records": [{"key": "abc", "value": {"name": "name0"}}]
}

# Second
POST http://localhost:8082/topics/test.v2
Content-Type: application/vnd.kafka.avro.v2+json
{
  "key_schema": "{\"type\": \"string\"}",
  "value_schema": "{\"namespace\": \"example.avro\", \"type\": \"record\", \"name\": \"simple\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}",
  "records": [{"key": "abc", "value": {"name": "name0"}}]
}

# Registered schemas:
GET http://localhost:8081/subjects

[
  "topic.v1-key",
  "topic.v1-value"
]
```
After patch:
```
...same two request

GET http://localhost:8081/subjects

[
  "topic.v1-key",
  "topic.v1-value",
  "topic.v2-key",
  "topic.v2-value"
]
```

muralibasani:
@e11it thanks for the PR. Will be reviewed soon.

PR #1262, opened 2026-04-23
Fix: Confusing doc for config security_protocol

jclarysse:

- Aligns supported values with [librdkafka config security.protocol reference](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md?plain=1#L64).
- Fixes inconsistent example for plaintext transport with SASL/PLAIN mechanism.
- Enriches config table with supported values.


muralibasani (approved):
LGTM

PR #1278, opened 2026-05-11
Split OIDC AuthN and AuthZ in Schema Registry

muralibasani:
Issue : 
sasl_oauthbearer_authorization_enabled gates both authentication and authorization in the Schema Registry. The naming is a bit misleading. There isn't a separate flag to enable authN without authZ; they're coupled behind this one switch even though implementation exists for both.

- Adds sasl_oauthbearer_authentication_enabled so token validation can be enabled without enforcing role-based authorization. sasl_oauthbearer_authorization_enabled continues to gate the role check and now requires authN.
- backwards compatibility: setting only authorization_enabled=true auto-enables authN with a deprecation warning.
- Distinct 401 reason for expired tokens ("Token expired" vs the previous generic "Invalid token/payload") — OIDCMiddleware.validate_jwt catches ExpiredSignatureError and raises a new TokenExpiredError.
- Brief inline docs in Config and kafka_rest_apis/authentication.py clarifying which OIDC fields apply to which service.
- Fix a couple of security issues

| authN | authZ | Behavior |
|---|---|---|
| F | F | No auth |
| T | F | JWT validation (401 on failure) |
| T | T | JWT + role check (401/403) |
| F | T | Invalid → coerced to T/T |

References: #xxxxx

# Why this way

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -86,8 +86,11 @@ services:
       KARAPACE_SERVER_TLS_CAFILE: /opt/karapace/certs/ca/rootCA.pem
       KARAPACE_SERVER_TLS_CERTFILE: /opt/karapace/certs/cert.pem
       KARAPACE_SERVER_TLS_KEYFILE: /opt/karapace/certs/key.pem
+      KARAPACE_SASL_OAUTHBEARER_AUTHENTICATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
+      # Dev-only: keycloak in compose serves http. Production must use https:// JWKS.
+      KARAPACE_SASL_OAUTHBEARER_ALLOW_INSECURE_JWKS: true
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_ISSUER: http://keycloak:8080/realms/karapace
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "karapace-audience"
       KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
@@ -144,14 +147,68 @@ services:
       KARAPACE_SERVER_TLS_CAFILE: /opt/karapace/certs/ca/rootCA.pem
       KARAPACE_SERVER_TLS_CERTFILE: /opt/karapace/certs/cert.pem
       KARAPACE_SERVER_TLS_KEYFILE: /opt/karapace/certs/key.pem
+      KARAPACE_SASL_OAUTHBEARER_AUTHENTICATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: true
       KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
+      # Dev-only: keycloak in compose serves http. Production must use https:// JWKS.
+      KARAPACE_SASL_OAUTHBEARER_ALLOW_INSECURE_JWKS: true
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_ISSUER: http://keycloak:8080/realms/karapace
       KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "karapace-audience"
       KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
       KARAPACE_SASL_OAUTHBEARER_CLIENT_ID: "karapace-client"
       KARAPACE_SASL_OAUTHBEARER_ROLES_CLAIM_PATH: "resource_access.karapace-client.roles"
       KARAPACE_SASL_OAUTHBEARER_METHOD_ROLES: '{"GET": ["schema:read", "subject:read"], "POST": ["schema:write", "subject:write"], "PUT": ["config_subject:update","config_global:update"], "DELETE": ["schema:delete", "subject:delete"]}'
+      KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
+
+  # Schema Registry running with OIDC authentication only (no role-based authorization).
+  # Used to verify the sasl_oauthbearer_authentication_enabled flag in isolation.
+  karapace-schema-registry-authn-only:
+    profiles:
+      - e2e
+    networks:
+      - karapace
+    image: ghcr.io/aiven-open/karapace:develop
+    hostname: karapace-schema-registry-authn-only
+    container_name: karapace-schema-registry-authn-only
+    entrypoint:
+      - python3
+      - -m
+      - karapace
+    depends_on:
+      - kafka
+      - opentelemetry-collector
+    ports:
+      - 8281:8281
+    volumes:
+      - ./certs:/opt/karapace/certs
+    environment:
+      KARAPACE_KARAPACE_REGISTRY: true
+      KARAPACE_ADVERTISED_HOSTNAME: karapace-schema-registry-authn-only
+      KARAPACE_ADVERTISED_PROTOCOL: https
+      KARAPACE_BOOTSTRAP_URI: kafka:29092
+      KARAPACE_PORT: 8281
+      KARAPACE_HOST: 0.0.0.0
+      KARAPACE_CLIENT_ID: karapace-schema-registry-authn-only
+      KARAPACE_GROUP_ID: karapace-schema-registry-authn-only
+      KARAPACE_MASTER_ELECTION_STRATEGY: highest
+      KARAPACE_MASTER_ELIGIBILITY: true
+      KARAPACE_TOPIC_NAME: _schemas_authn_only
+      KARAPACE_LOG_LEVEL: INFO
+      KARAPACE_COMPATIBILITY: FULL
+      KARAPACE_KAFKA_SCHEMA_READER_STRICT_MODE: false
+      KARAPACE_KAFKA_RETRIABLE_ERRORS_SILENCED: true
+      KARAPACE_TAGS__APP: karapace-schema-registry-authn-only
+      KARAPACE_SERVER_TLS_CAFILE: /opt/karapace/certs/ca/rootCA.pem
+      KARAPACE_SERVER_TLS_CERTFILE: /opt/karapace/certs/cert.pem
+      KARAPACE_SERVER_TLS_KEYFILE: /opt/karapace/certs/key.pem
+      KARAPACE_SASL_OAUTHBEARER_AUTHENTICATION_ENABLED: true
+      KARAPACE_SASL_OAUTHBEARER_AUTHORIZATION_ENABLED: false
+      KARAPACE_SASL_OAUTHBEARER_JWKS_ENDPOINT_URL: http://keycloak:8080/realms/karapace/protocol/openid-connect/certs
+      # Dev-only: keycloak in compose serves http. Production must use https:// JWKS.
+      KARAPACE_SASL_OAUTHBEARER_ALLOW_INSECURE_JWKS: true
+      KARAPACE_SASL_OAUTHBEARER_EXPECTED_ISSUER: http://keycloak:8080/realms/karapace
+      KARAPACE_SASL_OAUTHBEARER_EXPECTED_AUDIENCE: "karapace-audience"
+      KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
       KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
 
   karapace-rest-proxy:
--- a/src/karapace/api/middlewares/__init__.py
+++ b/src/karapace/api/middlewares/__init__.py
@@ -12,12 +12,58 @@ from prometheus_client import Counter
 from prometheus_fastapi_instrumentator import Instrumentator
 from prometheus_fastapi_instrumentator.metrics import Info, request_size, requests, response_size
 
-from karapace.api.oidc.middleware import OIDCMiddleware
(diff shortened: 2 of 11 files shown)


muralibasani:
@dylanbstorey would you be able to review this one ?

src/karapace/api/oidc/middleware.py
```diff
@@ -37,6 +41,13 @@ def __init__(self, app: FastAPI, config: Config) -> None:
 
         # Validate required fields if JWKS URL is set
         if self.jwks_url:
+            if not self.jwks_url.lower().startswith("https://"):
+                # Plain HTTP lets an in-path attacker swap the signing keys and
+                # forge tokens that pass validation. Use https:// in production.
+                log.warning(
+                    "OIDC: JWKS URL uses plain HTTP (%s) — INSECURE. Use https:// in production.",
+                    self.jwks_url,
+                )
```
dylanbstorey, line 55: This warning is the only signal a production deployment gets for an `http://` JWKS URL, and `container/compose.yml` ships `http://keycloak:8080/...` for the e2e/dev profile — so the warning fires constantly in dev and trains operators to ignore it. The PR description claims to "fix a couple of security issues," but as written an operator copy-pasting the docker-compose example into prod gets no friction at all.

Suggest escalating to `ValueError` by default with an explicit opt-out for dev:

```python
if not self.jwks_url.lower().startswith("https://"):
    if not config.sasl_oauthbearer_allow_insecure_jwks:
        raise ValueError(
            "OIDC config error: jwks_endpoint_url must use https://. "
            "Set sasl_oauthbearer_allow_insecure_jwks=true to override (dev only)."
        )
    log.warning("OIDC: JWKS URL uses plain HTTP — INSECURE. Override is active.")
```

The compose profiles can flip the override; production stays fail-closed.

src/karapace/api/oidc/middleware.py
```diff
@@ -73,13 +86,16 @@ def __init__(self, app: FastAPI, config: Config) -> None:
                     self.sasl_oauthbearer_roles_claim_path,
                 )
         else:
+            if self.authentication_enabled or self.authorization_enabled:
+                raise ValueError(
+                    "OIDC config error: sasl_oauthbearer_jwks_endpoint_url is required when "
+                    "authentication or authorization is enabled. Also set expected_issuer and expected_audience."
```
dylanbstorey, line 95: Style nit / consistency: the new `raise ValueError(...)` on these lines is the right shape for startup-validation failures. The pre-existing `raise HTTPException(status_code=403, ...)` calls above (lines 69–72 for the `client_id`/`roles_claim_path` check, and 78–81 for `missing_methods`) should be converted to `ValueError` too.

`HTTPException` from `__init__` is the wrong shape — there is no request to attach a status code to, and the traceback during process boot is confusing to operators (it surfaces as a 500-ish error in tooling that doesn't expect HTTPException at import/init time). Since the PR is already touching this constructor, it's a clean moment to make all startup-validation failures look the same.

src/karapace/api/middlewares/__init__.py
```diff
@@ -48,21 +48,30 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
...
-            try:
-                oidc_middleware.authorize_request(payload, request.method)
-            except HTTPException as e:
-                return JSONResponse(
-                    {"error": "Authorization error", "reason": e.detail},
-                    status_code=e.status_code,
-                )
+            if config.sasl_oauthbearer_authorization_enabled:
+                try:
+                    oidc_middleware.authorize_request(payload, request.method)
```
dylanbstorey: `payload` is bound inside the `try` above; this reference works today only because every `except` branch returns. If a future `except` branch is added that does **not** return, this becomes a silently-unbound reference. mypy/pylint in stricter configs will already flag it.

Cleanest fix is to move the authz block into an `else:` clause on the try, which makes the dependency on a successful validation explicit:

```python
try:
    payload = oidc_middleware.validate_jwt(token)
    request.state.user = (
        payload.get(oidc_middleware.claim_name) if oidc_middleware.claim_name else None
    )
    log.debug("Authenticated")
except TokenExpiredError:
    return JSONResponse(...)
except AuthenticationError:
    return JSONResponse(...)
else:
    if config.sasl_oauthbearer_authorization_enabled:
        try:
            oidc_middleware.authorize_request(payload, request.method)
        except HTTPException as e:
            return JSONResponse(...)
```

(Alternative: initialise `payload: dict = {}` above the `try`. Works, but doesn't communicate intent.)

dylanbstorey (commented):
Three blocking comments inline — the rest of my review notes are nits or docs polish. Summary: the authN/authZ split itself is clean and well-tested; what's worth gating merge on are (1) the HTTPS posture for JWKS URLs, (2) `HTTPException` raised from `__init__`, and (3) a possibly-unbound `payload` reference. Details on each below.

src/karapace/api/oidc/middleware.py
```diff
@@ -48,23 +64,23 @@ def __init__(self, app: FastAPI, config: Config) -> None:
                 self.audience,
             )
 
-            self._jwks_client = PyJWKClient(self.jwks_url)
+            # lifespan caps how long a key stays cached after IdP rotation/revocation;
+            # otherwise PyJWKClient caches indefinitely until process restart.
+            self._jwks_client = PyJWKClient(self.jwks_url, cache_keys=True, lifespan=300, max_cached_keys=16)
```
HelenMel, line 69: Why 16? Is it an estimation for  one token per role? I suggest make it configurable, if this a case. Is it LRU cache?

src/karapace/api/middlewares/__init__.py
```diff
@@ -48,21 +48,31 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
             token = auth_header.split(" ", 1)[1]
```
HelenMel: What if parsing failed. Token is empty, or not existing?

src/karapace/api/oidc/middleware.py
muralibasani: An identity provider usually has 1–3 active signing keys and during rotation, briefly 2–4.
Here 16 is a comfortable headroom. If they have many, then they would have bigger problems than cache. I shall add some docs here specific to caching

src/karapace/api/middlewares/__init__.py
muralibasani: Good point. I shall add a validation and throw 401 err

src/karapace/api/oidc/middleware.py
muralibasani: Good one. Updated.

src/karapace/api/oidc/middleware.py
muralibasani: Indeed, good you raised. Updated.

src/karapace/api/middlewares/__init__.py
muralibasani: Agree, moved it to else block.

src/karapace/api/middlewares/__init__.py
```diff
@@ -45,24 +45,39 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
...
+                return JSONResponse(
+                    status_code=401,
+                    content={"error": "Unauthorized", "reason": "Token expired"},
+                )
             except AuthenticationError:
                 return JSONResponse(
                     status_code=401,
                     content={"error": "Unauthorized", "reason": "Invalid token/payload"},
                 )
+            else:
```
HelenMel: Could you, please, extract authentication and authorization to a separate functions (return succeed or not succeed): at this point it is already hard to follow! I totaly forget that it is all part of  set_content_type function

container/compose.yml
```diff
@@ -154,6 +160,57 @@ services:
       KARAPACE_SASL_OAUTHBEARER_METHOD_ROLES: '{"GET": ["schema:read", "subject:read"], "POST": ["schema:write", "subject:write"], "PUT": ["config_subject:update","config_global:update"], "DELETE": ["schema:delete", "subject:delete"]}'
       KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
 
+  # Schema Registry running with OIDC authentication only (no role-based authorization).
+  # Used to verify the sasl_oauthbearer_authentication_enabled flag in isolation.
+  karapace-schema-registry-authn-only:
```
nosahama, line 165: **[non-blocking]**

Do we really need this? Maybe whoever needs it should just toggle the environment variable.
It's good to test it locally during development, but having this extra service in here adds a bit more delay to the docker-compose stack startup time, adds more resource usage on the backing kafka store, etc.

src/karapace/api/middlewares/__init__.py
```diff
@@ -45,24 +45,39 @@ async def set_content_types(request: Request, call_next: Callable[[Request], Awa
                 )
 
             # Header exists and starts with Bearer → validate JWT
-            token = auth_header.split(" ", 1)[1]
+            token = auth_header[len("Bearer ") :].strip()
```
nosahama: **[non-blocking]**
I actually think the split was more explicit, any reason why this changed?

src/karapace/api/oidc/middleware.py
nosahama: Would we then also now need to handle `ValueError` in the middleware or this is taken care of?
Also this will no longer have the `exc.detail and exc.status_code` attributes that are used somewhere in the middleware when handling `HTTPException`

src/karapace/core/config.py
```diff
@@ -202,6 +208,20 @@ def normalize_producer_acks(cls, v):
             return -1
         return v
 
+    @model_validator(mode="after")
+    def _coerce_oidc_authn_authz_flags(self) -> Config:
+        # Backwards-compat: prior to splitting authN/authZ, sasl_oauthbearer_authorization_enabled was the
+        # single OIDC switch and implied authentication. Auto-enable authN if only authZ was set, with a
+        # deprecation warning so operators migrate to setting both flags explicitly.
+        if self.sasl_oauthbearer_authorization_enabled and not self.sasl_oauthbearer_authentication_enabled:
+            logging.getLogger(__name__).warning(
```
nosahama: **[non-blocking]**
maybe move `logging.getLogger(__name__)` to the top of the file as we normally do. 

**[nit]**
there is no coercion happening here, we are actually enforcing `sasl_oauthbearer_authentication_enabled`, the function name could change, but naming is hard, just let's validate with a test too, not sure if I saw a test for this behaviour.

src/karapace/kafka_rest_apis/authentication.py
```diff
@@ -1,4 +1,8 @@
 """
+REST Proxy auth: Bearer JWT is forwarded to Kafka via SASL/OAUTHBEARER; Kafka validates
```
nosahama: **[nit]**
let's leave license text first and move this after or closer to where it's used.

container/compose.yml
nosahama: **[thinking]**
I see you used this in the e2e tests, now it's tricky, in our integration tests, we have a way to intercept the config before starting the binaries, maybe we move those tests there, but this is easier, so doesn't hurt.

src/karapace/api/oidc/middleware.py
muralibasani: No new handling needed. The two HTTPException → ValueError conversions are both inside OIDCMiddleware.__init__  and a bad config will crash the server which is expected behaviour

container/compose.yml
muralibasani: Yea with this profiling here, default docker compose and smoke tests are not affected and the e2e setup is providing us a way to test authentication only

src/karapace/api/middlewares/__init__.py
muralibasani: A header like "Bearer    " would pass startswith("Bearer ") and split(" ", 1)[1] returns "   " ... a non-empty string that flows into validate_jwt and fails , but with strip, token is "" and we have condition later which will return 401 earlier.

src/karapace/core/config.py
muralibasani: Good points, updated.

src/karapace/api/middlewares/__init__.py
muralibasani: Good one. Updated. It's better now.

container/compose.yml
nosahama: Ah nice, that would make sense, can you help me know where we spin up the e2e using this profile?
I've also never heard about profiles in docker-compose, thanks.

muralibasani: This is in GNUMakefile https://github.com/Aiven-Open/karapace/pull/1278/changes#diff-e3445fc75aa9c3e4a60fbe5394dcce12693022018216a8fbe0000fe9952850a6R143 which should do the trick

nosahama: That means this guy - https://github.com/Aiven-Open/karapace/blob/main/tests/e2e/conftest.py#L105 will now run against the service with the `e2e` profile, meaning this https://github.com/Aiven-Open/karapace/pull/1278/changes#diff-824bc3e5c28226a3c0f4581750fa93a7da7001aff9e0ff4aee4131e3edd7b144R233 now duplicates what we have, or rather we might need to add a profile to the other services too, or maybe I need to test locally later.

nosahama: Actually it's not duplicated, as I see the ports are different, but then the initial registry cluster should error as it's not started since it doesn't have the `e2e` profile.

muralibasani: One in GNUMakefile is for smoketests, and one in compose is for docker process I think

src/karapace/api/middlewares/__init__.py
```diff
@@ -12,14 +12,60 @@
...
 
 log = logging.getLogger(__name__)
 
 
+def _authenticate_and_authorize(request: Request, config: Config, oidc_middleware: OIDCMiddleware) -> JSONResponse | None:
+    """Run the OIDC auth gate. Return a JSONResponse on failure, or None to continue."""
+    auth_header = request.headers.get("Authorization")
+    if not auth_header or not auth_header.startswith("Bearer "):
+        return JSONResponse(
+            status_code=401,
```
nosahama, line 28: **[non-blocking]**
`status_code=401` shows up in a lot of places, we could use enums from both fastapi or python (`from http import HTTPStatus`).

nosahama (approved):

PR #1281, opened 2026-05-13
fix: 504 on map-of-union Avro fetch

smirnoffs:
DatumReader and DatumWriter were constructed fresh on every message. For schemas with multi-way union values in maps/arrays, the avro library traverses the full schema graph on each __init__, causing O(messages × schema_complexity) CPU work that blocks the event loop and triggers 504s on large batches.

Cache DatumReader/DatumWriter instances keyed by schema fingerprint so construction cost is paid once per unique schema across all messages and requests.

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/core/serialization.py
+++ b/src/karapace/core/serialization.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 from aiohttp import BasicAuth
 from async_lru import alru_cache
 from avro.io import BinaryDecoder, BinaryEncoder, DatumReader, DatumWriter
-from cachetools import TTLCache
+from cachetools import LRUCache, TTLCache
 from collections.abc import Callable, MutableMapping
 from google.protobuf.message import DecodeError
 from jsonschema import ValidationError
@@ -42,6 +42,9 @@ import struct
 START_BYTE = 0x0
 HEADER_FORMAT = ">bI"
 HEADER_SIZE = 5
+
+_datum_reader_cache: MutableMapping[str, DatumReader] = LRUCache(maxsize=100)
+_datum_writer_cache: MutableMapping[str, DatumWriter] = LRUCache(maxsize=100)
 
 
 class DeserializationError(Exception):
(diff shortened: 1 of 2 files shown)


src/karapace/core/serialization.py
```diff
@@ -43,6 +43,9 @@
 HEADER_FORMAT = ">bI"
 HEADER_SIZE = 5
 
+_datum_reader_cache: MutableMapping[str, DatumReader] = LRUCache(maxsize=100)
```
nosahama, line 46: **[non-blocking]**
I would avoid a global cache here.

nosahama: **[non-blocking]**
Since we have more reads, the writers cache might not be necessary now.

nosahama: **[non-blocking]**
Also `LRUCache(maxsize=100)`, might be too small of a cap and this would mean that we keep those objects in memory even without any recent usage.
We could verify the existing and new resources usage.

tests/unit/test_serialization.py
```diff
@@ -442,3 +443,76 @@ def test_name_strategy_for_protobuf(expected_subject: Subject, strategy: NameStr
...
+
+
+def _encode_map_union_message(record: dict) -> bytes:
+    """Encode a record using MAP_UNION_SCHEMA into raw Avro binary (no framing header)."""
+    bio = io.BytesIO()
+    write_value(None, MAP_UNION_SCHEMA, bio, record)  # type: ignore[arg-type]
+    return bio.getvalue()
+
+
+def test_read_value_reuses_datum_reader_for_map_union_schema(karapace_container: KarapaceContainer) -> None:
```
nosahama, line 471: **[non-blocking]**
An integration or e2e test would be nice to test this, in addition to this.

muralibasani:
@smirnoffs closing this PR. Changes are already main from another PR.

PR #1282, opened 2026-05-14
serializer,avro,rest-proxy: event loop improvements

nosahama:
This PR improves Avro deserialization behavior in the schema registry serializer and REST proxy consumer path by caching Avro datum readers, offloading Avro reads from the event loop, and ensuring Avro bytes fields can be returned safely as JSON-compatible base64 strings.

### Changes

- Added per-thread cached Avro DatumReader usage and offloaded Avro reads via asyncio.to_thread.
- Converted Avro bytes/bytearray values to base64 strings during deserialization.
- Added unit, integration, and e2e coverage for complex Avro unions, map unions, and bytes-heavy Debezium-style schemas.

### Performance 

There is a significant improvement to the memory usage while running the performance tests for rest-proxy, all with 0 failures. 

IT/e2e tests were added, It would be nice to test this further with load.

### DIFF
--- a/src/karapace/core/serialization.py
+++ b/src/karapace/core/serialization.py
@@ -36,8 +36,11 @@ from urllib.parse import quote
 import asyncio
 import avro
 import avro.schema
+import base64
 import io
 import struct
+import threading
+import weakref
 
 START_BYTE = 0x0
 HEADER_FORMAT = ">bI"
@@ -306,6 +309,10 @@ class SchemaRegistrySerializer:
         self.ids_to_schemas: dict[int, TypedSchema] = {}
         self.ids_to_subjects: MutableMapping[int, list[Subject]] = TTLCache(maxsize=10000, ttl=600)
         self.schemas_to_ids: dict[str, SchemaId] = {}
+        self._avro_readers_lock = threading.Lock()
+        self._avro_readers_by_thread: weakref.WeakKeyDictionary[threading.Thread, TTLCache[SchemaId, DatumReader]] = (
+            weakref.WeakKeyDictionary()
+        )
 
     async def close(self) -> None:
         if self.registry_client:
@@ -380,12 +387,48 @@ class SchemaRegistrySerializer:
                 schema, _ = await self.get_schema_for_id(schema_id)
                 if schema is None:
                     raise InvalidPayload("No schema with ID from payload")
-                ret_val = read_value(self.config, schema, bio)
+                if schema.schema_type is SchemaType.AVRO:
+                    ret_val = await asyncio.to_thread(self._read_avro_value, SchemaId(schema_id), schema, bio)
+                else:
+                    ret_val = read_value(self.config, schema, bio)
                 return ret_val
             except (UnicodeDecodeError, TypeError, avro.errors.InvalidAvroBinaryEncoding) as e:
                 raise InvalidPayload("Data does not contain a valid message") from e
             except avro.errors.SchemaResolutionException as e:
                 raise InvalidPayload("Data cannot be decoded with provided schema") from e
+
+    def _get_avro_reader(self, schema_id: SchemaId, schema: TypedSchema) -> DatumReader:
+        current_thread = threading.current_thread()
+        with self._avro_readers_lock:
+            reader_cache = self._avro_readers_by_thread.get(current_thread)
+            if reader_cache is None:
+                reader_cache = TTLCache(maxsize=10000, ttl=600)
+                self._avro_readers_by_thread[current_thread] = reader_cache
+
+        reader = reader_cache.get(schema_id)
+        if reader is None:
+            reader = DatumReader(writers_schema=schema.schema)
+            reader_cache[schema_id] = reader
+        return reader
+
+    def _read_avro_value(self, schema_id: SchemaId, schema: TypedSchema, bio: io.BytesIO) -> Any:
+        return read_value(self.config, schema, bio, avro_reader=self._get_avro_reader(schema_id, schema))
+
+
+def _jsonify_avro_payload(value: Any) -> Any:
+    if isinstance(value, bytes):
+        return base64.b64encode(value).decode("ascii")
+
+    if isinstance(value, bytearray):
+        return base64.b64encode(bytes(value)).decode("ascii")
+
+    if isinstance(value, list):
+        return [_jsonify_avro_payload(item) for item in value]
+
+    if isinstance(value, dict):
+        return {key: _jsonify_avro_payload(item) for key, item in value.items()}
+
+    return value
 
 
 def flatten_unions(schema: avro.schema.Schema, value: Any) -> Any:
@@ -450,10 +493,10 @@ def flatten_unions(schema: avro.schema.Schema, value: Any) -> Any:
     return value
 
 
-def read_value(config: Config, schema: TypedSchema, bio: io.BytesIO):
+def read_value(config: Config, schema: TypedSchema, bio: io.BytesIO, avro_reader: DatumReader | None = None):
     if schema.schema_type is SchemaType.AVRO:
-        reader = DatumReader(writers_schema=schema.schema)
-        return reader.read(BinaryDecoder(bio))
+        reader = avro_reader if avro_reader is not None else DatumReader(writers_schema=schema.schema)
+        return _jsonify_avro_payload(reader.read(BinaryDecoder(bio)))
     if schema.schema_type is SchemaType.JSONSCHEMA:
         value = json_decode(bio)
         try:
--- a/src/karapace/kafka_rest_apis/consumer_manager.py
+++ b/src/karapace/kafka_rest_apis/consumer_manager.py
@@ -23,7 +23,13 @@ from karapace.core.config import Config
 from karapace.core.kafka.common import translate_from_kafkaerror
 from karapace.core.kafka.consumer import AsyncKafkaConsumer
 from karapace.core.kafka.types import DEFAULT_REQUEST_TIMEOUT_MS, Timestamp
-from karapace.core.serialization import DeserializationError, InvalidMessageHeader, InvalidPayload, SchemaRegistrySerializer
+from karapace.core.serialization import (
+    DeserializationError,
+    InvalidMessageHeader,
+    InvalidPayload,
+    SchemaRegistrySerializer,
+    SchemaRetrievalError,
+)
 from karapace.core.utils import json_decode, JSONDecodeError
 from karapace.kafka_rest_apis.convert_to_int import convert_to_int
 from karapace.kafka_rest_apis.authentication import get_kafka_client_auth_parameters_from_config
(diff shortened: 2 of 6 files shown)


src/karapace/core/serialization.py
```diff
@@ -214,7 +216,19 @@ async def get_schema(
     async def get_schema_for_id(self, schema_id: SchemaId) -> tuple[TypedSchema, list[Subject]]:
         result = await self.client.get(f"schemas/ids/{schema_id}", params={"includeSubjects": "True"})
         if not result.ok:
-            raise SchemaRetrievalError(result.json()["message"])
```
nosahama, line 217: **[note to reviewer]**
If there was any key error thrown here, raising the response exception would result to crashing the server.

src/karapace/core/serialization.py
```diff
@@ -380,13 +396,51 @@ async def deserialize(self, bytes_: bytes) -> dict:
...
                 raise InvalidPayload("Data does not contain a valid message") from e
             except avro.errors.SchemaResolutionException as e:
                 raise InvalidPayload("Data cannot be decoded with provided schema") from e
 
+    def _get_avro_reader(self, schema_id: SchemaId, schema: TypedSchema) -> DatumReader:
+        thread_id = threading.get_ident()
+        with self._avro_readers_lock:
+            reader_cache = self._avro_readers_by_thread.get(thread_id)
+            if reader_cache is None:
+                reader_cache = TTLCache(maxsize=10000, ttl=600)
```
nosahama, line 405: **[note to reviewer]**
A follow up could be to make these ttl values come from config/env.

src/karapace/core/serialization.py
```diff
@@ -380,13 +396,51 @@ async def deserialize(self, bytes_: bytes) -> dict:
...
+        if reader is None:
+            reader = DatumReader(writers_schema=schema.schema)
+            reader_cache[schema_id] = reader
+        return reader
+
+    def _read_avro_value(self, schema_id: SchemaId, schema: TypedSchema, bio: io.BytesIO) -> Any:
+        return read_value(self.config, schema, bio, avro_reader=self._get_avro_reader(schema_id, schema))
+
+
+def _jsonify_avro_payload(value: Any) -> Any:
```
nosahama, line 418: **[note to reviewer]**
Extendible, mostly now for bytes, no major focus on other types which mostly already have concrete json support.

tests/integration/kafka_rest_apis/test_rest_consumer.py
```diff
@@ -27,6 +33,91 @@
...
+).read_text()
+DEBEZIUM_BYTES_TYPED_SCHEMA = ValidatedTypedSchema.parse(SchemaType.AVRO, DEBEZIUM_BYTES_SCHEMA_JSON)
+DEBEZIUM_BYTES_RECORD = {
+    "before": None,
+    "after": {
+        "id": 1,
+        "uuid": "11111111-1111-1111-1111-111111111111",
+        "rate_table_id": 100,
+        "prediction_log_id": 200,
+        "probability": {"scale": 2, "value": b"\x01\x02"},
```
nosahama, line 87: **[note to reviewer]**
This bytes output would end up causing the JSON encode error.

src/karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -682,7 +688,14 @@ async def deserialize_strict(bytes_: bytes | None) -> dict | None:
                 return None
             try:
                 return await self.deserializer.deserialize(bytes_)
```
nosahama, line 690: **[note to reviewer]**
Another follow up could be to spin up a task group and do the parsing explicitly in the event loop, for now the loop is unblocked.

src/karapace/core/serialization.py
```diff
@@ -306,6 +320,8 @@ def __init__(
         self.ids_to_schemas: dict[int, TypedSchema] = {}
         self.ids_to_subjects: MutableMapping[int, list[Subject]] = TTLCache(maxsize=10000, ttl=600)
         self.schemas_to_ids: dict[str, SchemaId] = {}
+        self._avro_readers_lock = threading.Lock()
+        self._avro_readers_by_thread: dict[int, TTLCache[SchemaId, DatumReader]] = {}
```
nosahama: **[note to self]**
maybe If this reader does not cache it's processing based of the schemas, somehow this cache would not really work, or we just cache that memory block, or better long lived storage for these if they are really reused.

src/karapace/core/serialization.py
```diff
@@ -450,10 +504,10 @@ def get_name(obj) -> str:
     return value
 
 
-def read_value(config: Config, schema: TypedSchema, bio: io.BytesIO):
+def read_value(config: Config, schema: TypedSchema, bio: io.BytesIO, avro_reader: DatumReader | None = None):
     if schema.schema_type is SchemaType.AVRO:
-        reader = DatumReader(writers_schema=schema.schema)
-        return reader.read(BinaryDecoder(bio))
+        reader = avro_reader if avro_reader is not None else DatumReader(writers_schema=schema.schema)
+        return _jsonify_avro_payload(reader.read(BinaryDecoder(bio)))
```
muralibasani, line 499: read_value seems to be a shared path for sr and rest proxy. I think we want this fix only in rest proxy ?

src/karapace/core/serialization.py
```diff
@@ -380,13 +396,51 @@ async def deserialize(self, bytes_: bytes) -> dict:
...
                 if schema is None:
                     raise InvalidPayload("No schema with ID from payload")
-                ret_val = read_value(self.config, schema, bio)
+                if schema.schema_type is SchemaType.AVRO:
+                    ret_val = await asyncio.to_thread(self._read_avro_value, SchemaId(schema_id), schema, bio)
+                else:
+                    ret_val = read_value(self.config, schema, bio)
                 return ret_val
+            except SchemaRetrievalError as e:
+                raise InvalidPayload(f"Failed to retrieve schema for payload: {e}") from e
```
muralibasani: I am not sure about this propagation. I think we loose some observability, because SchemaRetrievalError can be for registry is down, or schema deleted or invalid payload, and now with this, we are loosing the precision. Can you pls double check.

src/karapace/core/serialization.py
```diff
@@ -380,13 +396,51 @@ async def deserialize(self, bytes_: bytes) -> dict:
                 schema, _ = await self.get_schema_for_id(schema_id)
                 if schema is None:
                     raise InvalidPayload("No schema with ID from payload")
-                ret_val = read_value(self.config, schema, bio)
+                if schema.schema_type is SchemaType.AVRO:
+                    ret_val = await asyncio.to_thread(self._read_avro_value, SchemaId(schema_id), schema, bio)
```
muralibasani, line 391: this to_thread would be applied on all decodes irrespective of size of payload. I think we need some benchmarking here ? what's the main reason behind this ?

src/karapace/core/serialization.py
muralibasani: here if running thread exits, it could still stay in the dict?  May be use another type of dict ?

src/karapace/core/serialization.py
```diff
@@ -214,7 +216,19 @@ async def get_schema(
...
+            error_payload = result.json()
+            if isinstance(error_payload, dict):
+                error_message = (
+                    error_payload.get("message")
+                    or error_payload.get("detail")
+                    or error_payload.get("reason")
+                    or error_payload.get("error")
+                )
+                if error_message is None:
+                    error_message = str(error_payload)
```
muralibasani: There could be sensitive info in this error_payload. May be cast it to another shape and preserve relevant info

src/karapace/core/serialization.py
muralibasani, line 405: TTLCache isn't thread safe, any alternatives ?

src/karapace/core/serialization.py
muralibasani, line 418: this is called recursively even when there is empty payload/bytes

muralibasani (changes requested):
I have a few comments, and also noticed there are some gaps in tests, documentation for new base64 returns. I did not check if these are breaking changes but worth documenting.

src/karapace/core/serialization.py
nosahama: I will revert this, it'd be nice if we maybe face the error itself and can debug the actual value.

src/karapace/core/serialization.py
nosahama: `TTLCache` is not thread-safe when the same cache instance is accessed from multiple threads, in our usage, each instance is thread-confined: one cache per worker thread.

src/karapace/core/serialization.py
nosahama: yes that makes sense, thanks.

src/karapace/core/serialization.py
nosahama: prior to any of the changes here, this function was blocking the event loop with the synchronous reads and also started a reader for every call.

src/karapace/core/serialization.py
nosahama: I think i will propagate it to the caller, that feels better.

nosahama: There's also no observability tied at that level.
`_create_format_strict_func` catches all those errors and raises `DeserializationError`.

src/karapace/core/serialization.py
nosahama: For empty bytes, the correct JSON-safe/base64-normalized value is `""`, not raw `b""` and not `None`, if we also have somehow a `[None, None, ...]`, it will not match any of the conditionals.

nosahama:
> I have a few comments, and also noticed there are some gaps in tests, documentation for new base64 returns. I did not check if these are breaking changes but worth documenting.

@muralibasani Thanks for the review. 

The Debezium schema did not break Avro decoding itself, it exposed a bug in Karapace’s Avro-to-JSON response path, which we didn't test for or surprisingly no customer has used until now. Avro bytes fields deserialize in Python as raw bytes, REST proxy responses must be JSON, and raw Python bytes are not JSON-serializable.

- It'd be nice to help me understand the gaps in tests you speak off.
- As for the documentation need, if there is any related section to rest-proxy responses, i can add this, but this is not breaking compatibility as this is fixing a bug, so Avro bytes had never worked in rest-proxy.

src/karapace/core/serialization.py
nosahama: Off a [quick search](https://github.com/search?q=repo%3AAiven-Open%2Fkarapace+read_value%28&type=code), this is used only in the `deserialize()` method of the `SchemaRegistrySerializer`, that method is then only used in the consumer manager, also from a [quick search](https://github.com/search?q=repo%3AAiven-Open%2Fkarapace+deserialize%28&type=code).
The concern is valid, I am not sure if we use this in schema-registry, or if callers use directly `deserialize()` outside the standard karapace code.

muralibasani (approved):

PR #1285, opened 2026-05-19
Move main entrypoint to a main() function

irdkwmnsb:
pyproject.toml's `[project.scripts]` section states that karapace script should run the main function, so I create it.

# Why this way

Because of https://github.com/Aiven-Open/karapace/blob/main/pyproject.toml#L68


irdkwmnsb:
@muralibasani can you take a look at this please?

PR #1286, opened 2026-05-19
Karapace: added testacases for schema registry and core

roshini-aiven:


muralibasani:
I expect the new test files to be under core, kafka_rest_apis dirs. Can you pls move them from root dir?

roshini-aiven:
Good point @muralibasani,  switched the location.

muralibasani (approved):

Issue #1274, opened 2026-05-07
When Karapace schema registry has OIDC configured, schema backed requests to Karapace REST fail


KeatonSpiller:
# What happened?

When both OIDC and basicAuth are enabled at the same time within the schema-registry back-end, then all client requests providing basicAuth and OIDC will break.

Ideally we wanted the Schema-Registry to authenticate with the token provided from SchemaRest, but if I'm not mistaken that functionality is currently not supported?

We are successfully sending requests from an external client (OIDC) -> Karapace Rest -> KafkaSASL with JSON 
```
curl -k \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/vnd.kafka.json.v2+json" \
  -X POST "https://$REST_HOST/topics/$TOPIC" \
  -d '{
    "records": [
      {
        "value": {
          "age": "'"$RANDOM"'"
        }
      }
    ]
  }'
```
and we are failing when using AVRO when OIDC is enabled,
External Client (OIDC) -> Karapace-Rest (OIDC) -> Karapace-Schema
```
curl -k \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/vnd.kafka.avro.v2+json" \
  -X POST "https://${REST_HOST}/topics/$TOPIC" \
  -d '{
    "value_schema": "{\"type\": \"record\", \"name\": \"simple\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}",
    "records": [
      {
        "value": {
          "name": "'"$RANDOM"'"
        }
      }
    ]
  }'
echo -e "\n" 
```
  
External Client Terminal Log Error: 
```
{"error_code":40801,"message":"Error when registering schema.format = AVRO, subject = <TOPICNAME>-value"}
```
  
KarapaceRest Server log Error:
```
aiohttp.access MainThread INFO 0.011162s - "POST /topics/<TOPICNAME> HTTP/1.1" 408 "curl/7.76.1" response=401b request_body=225b
```

# What did you expect to happen?

When both OIDC and basicAuth are enabled at the same time within schema-registry,
  
External Client (OIDC) -> Schema-Registry
```
TOKEN=$(curl -k -X POST "$TOKEN_URL" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "grant_type=password" \
  -d "client_id=schema-registry" \
  -d "username=<USER>" \
  -d "password=<PASSWORD>" \
  | jq -r .access_token)
[...]

PR #1287, opened 2026-05-19
rest-proxy: forward inbound OIDC bearer to Schema Registry

muralibasani:
<!-- All contributors please complete these sections, including maintainers -->

When the Schema Registry has `sasl_oauthbearer_authentication_enabled=true`, the REST Proxy was calling SR with its static `BasicAuth` (or nothing) and getting back 401s, surfaced to clients as the generic `40801 "Error when registering schema"`. This PR threads the inbound caller's `Authorization` header through to SR for the duration of the request, gated by the same flag that already turns on SR's `OIDCMiddleware`.                                                                                                                                  
                                                                                                                                                                   
The forwarding is implemented with a module-level `contextvars.ContextVar` (`sr_authorization_ctx`) rather than a kwarg threaded through ~10 method signatures across the produce/consume path. asyncio copies the context across `await` boundaries and per task in `asyncio.gather`, so per-request isolation is automatic.                                                                                                                                                       
                                                                                                                                                               
## Behavior matrix                    
                                                        
Forwarding is gated by `config.sasl_oauthbearer_authentication_enabled`. When `False` (the default and every existing deployment), behavior is byte-identical to before.                            
                                                        
| flag | inbound `Authorization` | `registry_user`/`password` | outgoing to SR |                                                                                 
|------|--------------------------|----------------------------|----------------|
| `False` | any | any | unchanged — Basic if creds set, else nothing |                                                                                           
| `True`  | `Bearer …` | any | `Authorization: Bearer …` (overrides Basic for the call) |                                                                        
| `True`  | `Basic …`  | any | forwarded as-is |                                                                                                                 
| `True`  | none       | yes | Basic (existing fallback) |                                                                                                       
| `True`  | none       | no  | unauthenticated → SR 401 | 

<!-- Provide a small sentence that summarizes the change. -->

<!-- Provide the issue number below if it exists. -->
References: #1274 

# Why this way

<!-- Provide a small explanation on why this is the approach you took for solving this problem. -->

### DIFF
--- a/container/compose.yml
+++ b/container/compose.yml
@@ -160,8 +160,8 @@ services:
       KARAPACE_SASL_OAUTHBEARER_METHOD_ROLES: '{"GET": ["schema:read", "subject:read"], "POST": ["schema:write", "subject:write"], "PUT": ["config_subject:update","config_global:update"], "DELETE": ["schema:delete", "subject:delete"]}'
       KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
 
-  # Schema Registry running with OIDC authentication only (no role-based authorization).
-  # Used to verify the sasl_oauthbearer_authentication_enabled flag in isolation.
+  # OIDC authentication only (no role-based authorization).
+  # Used to isolate the sasl_oauthbearer_authentication_enabled flag in tests.
   karapace-schema-registry-authn-only:
     profiles:
       - e2e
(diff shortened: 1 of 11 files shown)


muralibasani:
@KeatonSpiller would you be able to review and test this PR branch ? 
This is related to https://github.com/Aiven-Open/karapace/issues/1274

HelenMel (dismissed):

tests/unit/kafka_rest_apis/test_sr_authorization_forwarding.py
```diff
@@ -0,0 +1,175 @@
+"""
+Copyright (c) 2025 Aiven Ltd
```
HelenMel: 2026

container/compose.yml
```diff
@@ -211,6 +211,151 @@ services:
       KARAPACE_SASL_OAUTHBEARER_SUB_CLAIM_NAME: sub
       KARAPACE_SASL_OAUTHBEARER_SKIP_AUTH_PATHS: '["/_health", "/metrics"]'
 
+  # REST Proxy with the forwarding gate ON, against the authn-only SR.
+  karapace-rest-proxy-oidc:
```
nosahama, line 215: **[non-blocking]**
These new services are not feeling nice to exist when we can toggle a config for what we need to change. 
We can revisit, maybe move those e2e tests to integration tests where it's easy to spin up the process with custom config.

src/karapace/core/serialization.py
```diff
@@ -122,9 +144,13 @@ def __init__(
         schema_registry_url: str = "http://localhost:8081",
         server_ca: str | None = None,
         session_auth: BasicAuth | None = None,
+        *,
+        cache_maxsize: int,
```
nosahama: **[non-blocking]**
No default value. so this has to be added to call sites, since it used to be `100`, maybe default to 100 here too.

tests/unit/test_serialization.py
```diff
@@ -693,3 +695,177 @@ def test_name_strategy_for_protobuf(expected_subject: Subject, strategy: NameStr
...
     )
+
+
+# Authorization forwarding via sr_authorization_ctx. Tested through observed headers
+# on the mocked Client — covers post_new_schema, _get_schema_recursive, get_schema_for_id,
+# and the @alru_cache partitioning on get_schema.
+
+
+@pytest.fixture
+def reset_sr_authorization_ctx():
```
nosahama: **[nit]**
this is repeated.

container/compose.yml
muralibasani: Agree, but also integration tests run slow per test. 
We can have a followup to move to integration tests.

nosahama: Yes that is true, they take a lot more time, I was checking recently.
These new services should be exemplar and not concrete.

src/karapace/core/serialization.py
```diff
@@ -212,10 +240,24 @@ async def get_schema(
...
+        # least once per fingerprint per cache lifetime, not per request. Cached entries
+        # outlive server-side token expiry — acceptable for immutable schema bytes, not a
+        # substitute for revocation. Empty fingerprint = the unauthenticated slot.
+        return await self._get_schema_cached(subject, version, _token_fingerprint())
+
+    async def _get_schema_cached(
+        self,
+        subject: Subject,
+        version: Version | None,
+        token_fingerprint: str,
```
nosahama, line 253: **[nit]**
this does not seem to be used.

src/karapace/kafka_rest_apis/__init__.py
```diff
@@ -641,6 +642,8 @@ async def seek_beginning_consumer_offsets(
         )
 
     async def fetch(self, group_name: str, instance: str, content_type: str, *, request: HTTPRequest) -> None:
+        if self.config.sasl_oauthbearer_authentication_enabled:
```
nosahama, line 645: **[non-blocking]**
it would be nice if this config naming could also hint that it is related to schema-registry auth forwarding from rest-proxy.

src/karapace/core/serialization.py
nosahama: probably for the cache key.

KeatonSpiller:
> @KeatonSpiller would you be able to review and test this PR branch ? This is related to #1274

I might not be able to get to it today, but I'll try to test this as soon as I can.

nosahama (approved):

PR #1291, opened 2026-05-21
test: restore schema workflow for residual check

qsq8t65pxs:
Minimal security validation PR. Restores the historical schema workflow only to verify whether the residual active workflow can still be reached from a fresh external fork PR. No exploit payload; will close after observing run status.


qsq8t65pxs:
Security validation complete; closing the minimal draft PR.

PR #1293, opened 2026-05-27
OIDC JWT validation leeway, sub require, typ azp

muralibasani:

3 new configs
```
sasl_oauthbearer_leeway_seconds
sasl_oauthbearer_require_at_jwt_typ
sasl_oauthbearer_enforce_azp
```

- Add `leeway` for clock skew (default 30s, configurable via sasl_oauthbearer_leeway_seconds) so minor IdP/Karapace drift no longer produces spurious 401s.
- Require the configured subject claim (default `sub`) in PyJWT's `require` list. Tokens missing the subject are now rejected with 401 rather than silently producing `request.state.user = None`.
- Drop the `audiences=None` fallback in validate_jwt; raise AuthenticationError("audience missing") as defense-in-depth alongside the existing __init__ guard.
- Surface the underlying InvalidTokenError detail in logs (`log.warning("JWT validation failed: %s", exc)`) so operators can debug misconfigurations. The 401 response body stays opaque.
- Add opt-in `sasl_oauthbearer_require_at_jwt_typ` (RFC 9068): rejects ID tokens being used as access tokens by checking `typ: at+jwt` on the unverified header before signature validation.
- Add opt-in `sasl_oauthbearer_enforce_azp`: enforces `azp == client_id` for OIDC authorized-party binding. Fail-closed at startup if enabled without client_id.

References: #xxxxx

# Why this way

### DIFF
--- a/.claude/skills/oidc-auth/SKILL.md
+++ b/.claude/skills/oidc-auth/SKILL.md
@@ -194,6 +194,9 @@ return http_authorizer if config.registry_authfile else no_auth_authorizer
 | `sasl_oauthbearer_roles_claim_path` | `None` | Dot-path to roles list in JWT |
 | `sasl_oauthbearer_method_roles` | `{GET/POST/PUT/DELETE: []}` | Per-method allowed roles |
 | `sasl_oauthbearer_skip_auth_paths` | `["/_health","/metrics"]` | Bypass paths |
+| `sasl_oauthbearer_leeway_seconds` | `30` | Clock-skew tolerance for `exp`/`nbf`/`iat` |
+| `sasl_oauthbearer_require_at_jwt_typ` | `False` | RFC 9068: require header `typ: at+jwt` (or `application/at+jwt`) |
+| `sasl_oauthbearer_enforce_azp` | `False` | Require `azp == client_id`; needs `client_id` |
 | `schema_registry_client_cache_maxsize` | `100` | LRU cap on `(subject, version, fp)` |
 | `registry_authfile` | `None` | Selects basic-auth `HTTPAuthorizer` |
 
--- a/.claude/skills/oidc-auth/SKILL.md
+++ b/.claude/skills/oidc-auth/SKILL.md
@@ -194,8 +194,8 @@ return http_authorizer if config.registry_authfile else no_auth_authorizer
 | `sasl_oauthbearer_roles_claim_path` | `None` | Dot-path to roles list in JWT |
 | `sasl_oauthbearer_method_roles` | `{GET/POST/PUT/DELETE: []}` | Per-method allowed roles |
 | `sasl_oauthbearer_skip_auth_paths` | `["/_health","/metrics"]` | Bypass paths |
-| `sasl_oauthbearer_leeway_seconds` | `30` | Clock-skew tolerance for `exp`/`nbf`/`iat` |
-| `sasl_oauthbearer_require_at_jwt_typ` | `False` | RFC 9068: require header `typ: at+jwt` (or `application/at+jwt`) |
+| `sasl_oauthbearer_leeway_seconds` | `0` | Clock-skew tolerance for `exp`/`nbf`/`iat` (must be >= 0) |
+| `sasl_oauthbearer_require_access_token_typ` | `False` | Require header `typ: at+jwt` (or `application/at+jwt`) |
 | `sasl_oauthbearer_enforce_azp` | `False` | Require `azp == client_id`; needs `client_id` |
 | `schema_registry_client_cache_maxsize` | `100` | LRU cap on `(subject, version, fp)` |
 | `registry_authfile` | `None` | Selects basic-auth `HTTPAuthorizer` |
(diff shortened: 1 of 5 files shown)


src/karapace/api/oidc/middleware.py
```diff
@@ -99,32 +104,41 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
+
             signing_key = self._jwks_client.get_signing_key_from_jwt(token)
-            # Split the comma-separated audience string into a list and strip whitespace
-            if self.audience:
-                audiences = [aud.strip() for aud in self.audience.split(",") if aud.strip()]
-            else:
-                audiences = None  # or leave it out
+            audiences = [aud.strip() for aud in self.audience.split(",") if aud.strip()]
+            require = ["exp", "iss", "aud"]
+            if self.claim_name and self.claim_name not in require:
```
HelenMel: you can just create a set and add to set, a value will be unique

src/karapace/api/oidc/middleware.py
```diff
@@ -99,32 +104,41 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
                 audience=audiences,
                 issuer=self.issuer,
-                # Require exp/iss/aud so a token missing any of these is rejected,
-                # not silently accepted (PyJWT does not require them by default).
-                options={"require": ["exp", "iss", "aud"]},
+                leeway=self.leeway_seconds,
+                # PyJWT does not require these by default; enforce presence explicitly.
+                options={"require": require},
             )
+            if self.enforce_azp and payload.get("azp") != self.client_id:
```
HelenMel: this is all separate checks, separate functions. It is much better to have them as a list or just queue of calls:
check_audience
check_header
check_client_id

Easier to read, and support

tests/unit/test_oidc.py
```diff
@@ -526,3 +529,208 @@ def test_middleware_invalid_token_still_returns_invalid_reason(monkeypatch):
...
     assert r.status_code == 401
     assert r.json()["reason"] == "Invalid token/payload"
+
+
+# ---------------------------------------------------------------------------
+# Hardening: leeway, require sub, log reason, audience guard, typ:at+jwt, azp.
+# ---------------------------------------------------------------------------
+
+
+@patch("karapace.api.oidc.middleware.PyJWKClient")
```
HelenMel, line 539: Instead of Patching, try to use include it as a dependency, and load in a test as a mock.

.claude/skills/oidc-auth/SKILL.md
```diff
@@ -194,6 +194,9 @@ return http_authorizer if config.registry_authfile else no_auth_authorizer
 | `sasl_oauthbearer_roles_claim_path` | `None` | Dot-path to roles list in JWT |
 | `sasl_oauthbearer_method_roles` | `{GET/POST/PUT/DELETE: []}` | Per-method allowed roles |
 | `sasl_oauthbearer_skip_auth_paths` | `["/_health","/metrics"]` | Bypass paths |
+| `sasl_oauthbearer_leeway_seconds` | `30` | Clock-skew tolerance for `exp`/`nbf`/`iat` |
```
Claudenw: 30 seconds seems like a large value for clock skew.
Q: since existing code has no clock skew correction, should this be 0 and users can set it to something else if they want to adjust for skew?  Then the default operation is the same as before the change.

src/karapace/api/oidc/middleware.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, app: FastAPI, config: Config) -> None:
         self.client_id = config.sasl_oauthbearer_client_id
         self.sasl_oauthbearer_method_roles: dict[str, list[str]] = config.sasl_oauthbearer_method_roles
         self.sasl_oauthbearer_roles_claim_path = config.sasl_oauthbearer_roles_claim_path
+        self.leeway_seconds = config.sasl_oauthbearer_leeway_seconds
```
Claudenw, line 38: What happens if I set the `leeway_seconds` to a negative number?  I didn't see a test for that below.

Claudenw (commented):
I don't know enough to approve or request changes; just ask questions.

src/karapace/api/oidc/middleware.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
+        self.require_at_jwt_typ = config.sasl_oauthbearer_require_at_jwt_typ
```
nosahama: **[non-blocking]**
`require_at_jwt_typ`, "at" is a bit confusing.

src/karapace/api/oidc/middleware.py
muralibasani: Good point. done.

src/karapace/api/oidc/middleware.py
muralibasani: Indeed it was not readable. Done.

tests/unit/test_oidc.py
muralibasani: This is a big change, implementing with DI, and worth doing it. Ok if I have a follow up PR ?

.claude/skills/oidc-auth/SKILL.md
muralibasani: Agree. Updated default to 0.

src/karapace/api/oidc/middleware.py
muralibasani: It shouldn't be negative. Added a condition and test.

src/karapace/api/oidc/middleware.py
muralibasani: Good one. Renamed it to access_token.

muralibasani:
@HelenMel @Claudenw @nosahama thanks for the reviews.
Updated PR based on the comments.

HelenMel (approved):

PR #1303, opened 2026-06-02
Karapace: added unit + integration tests for OIDC method check

roshini-aiven:
- Fixed the dead-code set(keys) - keys check in OIDCMiddleware.__init__ to validate method_roles against a canonical REQUIRED_HTTP_METHODS = {GET, POST, PUT, DELETE} (case-insensitive), short-circuited OPTIONS and mapped HEAD→GET in authorize_request to handle FastAPI-auto-generated methods
- added unit tests + parametrized integration tests (tests/integration/test_oidc_startup_validation.py) that subprocess python -m karapace to verify startup fails with the expected error, and padded existing test_authorize_request_roles rows with _FULL_METHOD_ROLES so they pass the new completeness check.

### DIFF
--- a/src/karapace/api/oidc/middleware.py
+++ b/src/karapace/api/oidc/middleware.py
@@ -19,6 +19,10 @@ class TokenExpiredError(AuthenticationError):
 
 log = logging.getLogger(__name__)
 
+# Methods served by SR routers — each must appear in ``method_roles`` when authz
+# is on, otherwise ``authorize_request`` silently 403s that method.
+REQUIRED_HTTP_METHODS: frozenset[str] = frozenset({"GET", "POST", "PUT", "DELETE"})
+
 
 class OIDCMiddleware:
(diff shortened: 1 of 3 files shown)


src/karapace/api/oidc/middleware.py
```diff
@@ -158,6 +162,12 @@ def authorize_request(self, payload: dict, request_method: str) -> bool:
             return True
 
         request_method = request_method.upper()
+        # OPTIONS is CORS preflight, never carries auth context.
+        if request_method == "OPTIONS":
+            return True
+        # FastAPI auto-generates HEAD for every GET route; share GET's policy.
```
muralibasani: this comment looks misleading. I think FastAPI does not auto-generate HEAD for every GET route.

muralibasani (approved):
LGTM, thanks.

PR #1307, opened 2026-06-08
oidc: add pingidentity support

nosahama:

Add local OIDC support for both Keycloak and PingFederate in the Karapace dev and test workflow.

## What changed

- add shared local OIDC configuration in compose
- split auth-specific services into a dedicated compose overlay
- add PingFederate provisioning and token helper scripts
- support provider-specific OIDC token flows in e2e tests
- update middleware and config for insecure JWKS handling and role parsing
- keep `/master_available` open for readiness checks
- update smoke tests and docs for the new local OIDC workflow

## Decoded tokens

The token can be verified using https://www.jwt.io/, after running `make print-pingfederate-oidc-token` or  `make print-keycloak-oidc-token`

### Keycloak

```json
{
  "exp": 1780918699,
  "iat": 1780918399,
  "jti": "01278555-e9cf-40eb-b64d-c05243abb7e9",
  "iss": "http://keycloak:8080/realms/karapace",
  "aud": "karapace-audience",
  "sub": "3c6e6035-e771-45cd-8797-3a12674dde0f",
  "typ": "Bearer",
  "azp": "karapace-client",
  "resource_access": {
    "karapace-client": {
      "roles": [
        "schema:delete",
        "schema:read",
        "subject:write",
        "subject:delete",
        "config_global:update",
        "subject:read",
        "schema:write",
        "config_subject:update"
      ]
    }
  },
  "scope": "openid email profile",
  "email_verified": false,
  "clientHost": "192.168.224.18",
  "preferred_username": "service-account-karapace-client",
  "clientAddress": "192.168.224.18",
  "client_id": "karapace-client"
}
```

### Ping Identity

```json
{
  "scope": "openid",
  "authorization_details": [],
  "client_id": "karapace-client",
  "iss": "https://pingfederate:9031",
  "aud": "karapace-audience",
  "roles": [
    "schema:read schema:write schema:delete subject:read subject:write subject:delete config_subject:update config_global:update"
  ],
  "exp": 1780925583
}
```

## Caveats

- Keycloak remains the default local OIDC provider
- PingFederate uses `client_id` as the identity claim for this local flow
- This was tested against local ping federate docker container
  - We need to test against a deployed version or a SaaS version, ideally it should work the same. This can be done as a follow up to this. 
  - The only ideal change in that case would be to set: `OIDC_JWKS_ENDPOINT_URL`, `OIDC_EXPECTED_ISSUER`, `OIDC_ALLOW_INSECURE_JWKS` and `OIDC_TOKEN_URL`


container/oidc/keycloak/keycloak-oidc-flow
```diff
@@ -0,0 +1,179 @@
+2026-06-05 13:11:56,227 INFO OIDC request POST http://keycloak:8080/realms/master/protocol/openid-connect/token
```
nosahama: **[note to reviewer]**
I have added this and also [this](https://github.com/Aiven-Open/karapace/pull/1307/changes#diff-2786be94f71c6ffa4e9362eeae22e3e011cdf1c3c1715649081988ab6859766b) so we have a log of the request/response data for provisioning the OIDC resources for both Keycloak and PingIdentity, that should aid understanding, extending and debugging OIDC related things.

src/karapace/api/oidc/middleware.py
```diff
@@ -75,7 +77,15 @@ def __init__(self, app: FastAPI, config: Config) -> None:
...
             # max_cached_keys=16 is generous; IdPs typically expose 1-3 active signing keys.
-            self._jwks_client = PyJWKClient(self.jwks_url, cache_keys=True, lifespan=300, max_cached_keys=16)
+            self._jwks_client = PyJWKClient(
+                self.jwks_url,
+                cache_keys=True,
+                lifespan=300,
+                max_cached_keys=16,
+                headers={"Accept": "application/json"},
+                timeout=30,
+                ssl_context=ssl._create_unverified_context() if self.allow_insecure_jwks else None,
```
muralibasani, line 87: how about using public api ssl.create_default_context ?

tests/unit/test_oidc.py
```diff
@@ -230,7 +231,12 @@ def test_validate_token_invalid_token(mock_jwt_decode, mock_pyjwks_client, dummy
         ({"realm_access": {}}, "realm_access.roles", []),
         ({}, "realm_access.roles", []),
         ({"custom": {"nested": {"roles": ["reader"]}}}, "custom.nested.roles", ["reader"]),
-        ({"custom": {"nested": {"roles": "notalist"}}}, "custom.nested.roles", []),
+        ({"custom": {"nested": {"roles": "notalist"}}}, "custom.nested.roles", ["notalist"]),
```
muralibasani: This changed from [] to ["notalist"] . I could not find relevant code in oidc/middleware to handle the roles for keycloak/pi

src/karapace/api/oidc/middleware.py
```diff
@@ -190,17 +200,29 @@ def authorize_request(self, payload: dict, request_method: str) -> bool:
     @staticmethod
     def get_roles_from_claim_path(payload: dict[str, Any], path: str) -> list[str]:
         try:
+            if path in payload:
```
muralibasani: A token containing a flat "karapace.roles" claim now bypasses the dotted descent. How compatible is this with PI and KC ?

src/karapace/api/oidc/middleware.py
```diff
@@ -190,17 +200,29 @@ def authorize_request(self, payload: dict, request_method: str) -> bool:
...
             pass
         return []
+
+    @staticmethod
+    def _normalize_role_values(value: Any) -> list[str]:
+        if isinstance(value, list) and all(isinstance(role, str) for role in value):
+            if len(value) == 1:
+                flattened = [item for item in value[0].replace(",", " ").split() if item]
+                if flattened:
+                    return flattened
```
muralibasani: [""] hits the len(value) == 1 branch, flattened is empty → falls through to return value → returns [""]
Authorization will then test "" in allowed_roles, which is ok, but there is a log.warning which also logs the value. kind of leak.

tests/unit/core/kafka/test_admin.py
```diff
@@ -0,0 +1,67 @@
...
+    topic = admin_client.new_topic("topic")
+
+    assert topic.topic == "topic"
+    mock_single_futmap_result.assert_called_once_with({"topic": "future"})
+    assert admin_client.list_topics.call_args_list == [call(topic="topic", timeout=1), call(topic="topic", timeout=1)]
+    mock_sleep.assert_called_once_with(0.1)
+
+
+@patch("karapace.core.kafka.admin.time.sleep")
+def test_wait_for_topic_metadata_retries_after_kafkaexception(mock_sleep: MagicMock) -> None:
```
muralibasani, line 41: This test change `_wait_for_topic_metadata` is unrelated to oidc I think. Was it flaky ?

bin/oidc/get_oidc_token.py
```diff
@@ -0,0 +1,168 @@
...
+
+class TokenClient:
+    def __init__(self, settings: Settings, http_client: httpx.Client) -> None:
+        self.settings = settings
+        self.http_client = http_client
+
+    def _request_with_retry(self, method: str, url: str, **kwargs: Any) -> httpx.Response:
+        end_time = time.monotonic() + DEFAULT_READY_TIMEOUT_SECONDS
+
+        while True:
```
muralibasani, line 75: this while loop, retries for any http error. Ex 401 as well. It should retry in case of connectivity issues for ex.

bin/oidc/get_oidc_token.py
```diff
@@ -0,0 +1,168 @@
...
+
+    @field_validator("provider", mode="before")
+    @classmethod
+    def normalize_provider(cls, value: object) -> Provider:
+        if not isinstance(value, str):
+            return "keycloak"
+        normalized = value.strip().lower()
+        if normalized in {"pingfederate", "pingidentity"}:
+            return "pingfederate"
+        return "keycloak"
```
muralibasani, line 45: instead of sending keycloak, how about
raise ValueError(f"OIDC_PROVIDER={value!r} is not supported (expected: keycloak, pingfederate)") ?

container/oidc/keycloak/keycloak-oidc-flow
muralibasani: This log is not required. Not a good pattern.
We can add how tokens/patterns look like in readme.

muralibasani: Even though they are from dev container, having secret keys/blobs look like there is some real data.

src/karapace/api/oidc/middleware.py
```diff
@@ -190,17 +200,29 @@ def authorize_request(self, payload: dict, request_method: str) -> bool:
...
-            if isinstance(value, list):
-                if all(isinstance(role, str) for role in value):
-                    return value
+            return OIDCMiddleware._normalize_role_values(value)
         except Exception:
             pass
         return []
+
+    @staticmethod
+    def _normalize_role_values(value: Any) -> list[str]:
```
muralibasani, line 216: Can you pls add any unit tests for this?

muralibasani:
[e2e/integration tests]
CI only exercises Keycloak; PingFederate path is effectively manual unless someone runs it locally with their Ping DevOps credentials.

container/oidc/keycloak/keycloak-oidc-flow
nosahama: Okay, i reckon it might help understand the contract with the providers we tested with, it's also not related to the runtime logic. I can save local copies and drop this.

src/karapace/api/oidc/middleware.py
nosahama: I think the default context needs verification.

muralibasani (changes requested):
Thanks for the PR.

Reg e2e tests, I see parameterization, but looks like only keycloak is in the job.
Atleast one e2e test against the existing karapace-schema-registry (full authn+authz) compose service (for kc and pi/pf) would be good.
 
Have a few minor comments too.

bin/oidc/get_oidc_token.py
nosahama: This is used only in tests and dev tooling, i think we can update what we handle if the issue surfaces.

bin/oidc/get_oidc_token.py
nosahama: I was thinking to keep keycloak as the default, the way it was.

container/oidc/keycloak/keycloak-oidc-flow
nosahama: I can redact the secrets @muralibasani

tests/unit/core/kafka/test_admin.py
nosahama: Yes, it improve the e2e tests, I was checking and see that it was failing intermittently with this.
I also wanted to not make too much changes to existing code, but this didn't hurt.

container/oidc/keycloak/keycloak-oidc-flow
muralibasani: You can add anything important about token semantics/structure to readme. It helps there.

bin/oidc/get_oidc_token.py
muralibasani: retry on 401 is not a correct behaviour in general.

nosahama: Yes definitely, but this is a test script  and the retry is time boxed.

src/karapace/api/oidc/middleware.py
nosahama: It already exist, i will update with the scenarios you've hinted.

src/karapace/api/oidc/middleware.py
nosahama: I will add this case too to the tests, also reverting the logic to what it was, just adding extension to handle roles like: `"roles": ["a:a b:b c:c"] | "roles": ["a:a,b:b,c:c"]` as compared to what is the default expectation of `"roles": ["a:a", "b:b", "c:c"]`.

muralibasani (changes requested):
For PingFederate provisioning, pls add to compose and gnumakefile ?       
```
PINGFEDERATE_ROLES_CLAIM: roles
PINGFEDERATE_TOKEN_AUDIENCE: karapace-audience
PINGFEDERATE_TOKEN_ISSUER: https://pingfederate:9031
```

nosahama:
> For PingFederate provisioning, pls add to compose and gnumakefile ?
> 
> ```
> PINGFEDERATE_ROLES_CLAIM: roles
> PINGFEDERATE_TOKEN_AUDIENCE: karapace-audience
> PINGFEDERATE_TOKEN_ISSUER: https://pingfederate:9031
> ```

@muralibasani those values are already [set as defaults in the provisioner settings](https://github.com/Aiven-Open/karapace/pull/1307/changes#diff-2cddacabe9b1d63aab7419140e04e0cede00f5a7ef5cf0b21c5f1992811e8d21R29), we can override them using env vars with the same naming, I would not make any changes for those.
It'd be nice if you can give some insights into where this failed and why you needed to have them explicitly in your env.

muralibasani (approved):

PR #1313, opened 2026-06-19
refactoring oidc: revert PingFederate dev scaffolding, keep runtime src changes

muralibasani:

Refactor the OIDC test/dev scaffolding from #1307 to make room for adding additional IdPs (Entra, others) in follow-up PRs.
The current setup hard-wires PingFederate provisioning, compose overlays, and Makefile targets in a way that doesn't generalize. 

A real PingIdentity IdP still authenticates via the standard `KARAPACE_SASL_OAUTHBEARER_*` env vars — no PI-specific runtime code is removed.

These will be re-introduced per IdP under a shared layout (`container/oidc/{keycloak,pingfederate,entra}/`) in follow-up PRs.

References: #xxxxx

# Why this way


bin/get_oidc_token.py
```diff
@@ -0,0 +1,64 @@
...
+    resp.raise_for_status()
+    return resp.json()["access_token"]
+
+
+if __name__ == "__main__":
+    admin_token = get_admin_token()
+    client_uuid = get_client_uuid(admin_token)
+    client_secret = get_client_secret(client_uuid, admin_token)
+    token = get_oidc_token(client_secret)
+    print(token)
```
aindriu-aiven, line 68: Is the purpose of this class to allow a developer to get a token from the cli for use with the API?
If so maybe we could add a comment at the top of the class to describe this as it wasn't immediately obvious to me what the classes purpose was.

aindriu-aiven: this is actually being use din the smoke test in 
bin/smoke-test-schema-registry.sh 

Might still be worth adding a comment.

tests/e2e/schema_registry/test_oidc.py
```diff
@@ -156,12 +152,11 @@ async def test_schema_registry_oidc_authn_only_skip_paths(
...
     assert res.status_code == 200
 
-    res = await registry_async_client_oidc_authn_only_no_auth_header.get("master_available")
-    assert res.status_code == 200
-
 
 # Pins the OIDC-side 404 body shape; full forbidden-vs-missing parity is
 # exercised in tests/integration/test_schema_registry_auth.py.
+
+
```
aindriu-aiven: nit: extra lines can be removed

GNUmakefile
```diff
@@ -134,6 +103,7 @@ cleanest: cleaner
 	rm -fr '$(VENV_DIR)'
 
 .PHONY: requirements
+requirements:
```
aindriu-aiven: two `requirements:`  in a row

GNUmakefile
```diff
@@ -144,108 +114,56 @@ pin-requirements:
...
 stop-karapace-docker-resources:
-	$(DOCKER_COMPOSE_STANDARD) down -v --remove-orphans || true
-	$(DOCKER_COMPOSE_AUTH_PINGFEDERATE) down -v --remove-orphans || true
-	docker rm -f $(DOCKER_CONTAINERS) >/dev/null 2>&1 || true
+	$(DOCKER_COMPOSE) -f container/compose.yml down -v --remove-orphans
 
-.PHONY: prepare-docker-resources
-prepare-docker-resources:
+.PHONY: start-karapace-docker-resources
+start-karapace-docker-resources:
```
aindriu-aiven: duplicate `start-karapace-docker-resources:` here too

bin/get_oidc_token.py
muralibasani: Yes, this script helps for development purpose. Added a docstring above.

GNUmakefile
muralibasani: Good one. Removed.

aindriu-aiven (approved):
LGTM

RyanSkraba (approved):

PR #1320, opened 2026-06-25
fix(protobuf): preserve map_entry through binary deserialize/serialize and restore map<K,V> shorthand on normalisation

### LINKED ISSUE #1319: protobuf: binary-registered schemas silently drop map_entry, and ?normalize=true does not restore map<K,V> shorthand
## What happened?

Two related bugs affect PROTOBUF schemas registered as a base64-encoded `FileDescriptorProto` binary — the format produced by Confluent Java, C#, and .NET SDK serialisers.

**Bug 1 — `map_entry` is silently stripped.**

When a binary schema is registered, `_deserialize_msg()` in `src/karapace/core/protobuf/serialization.py` constructs `MessageElement` without reading `MessageOptions`. The `option map_entry = true` that marks a synthetic entry message as belonging to a `map<K,V>` field is silently dropped regardless of what the binary contained. Without it, consumers cannot distinguish a map field from a plain `repeated` message field.

**Bug 2 — `?normalize=true` does not restore `map<K,V>` shorthand.**

Even if Bug 1 is fixed, normalisation leaves the expanded entry-message form in the output rather than converting it back to `map<K,V>` syntax. Confluent Schema Registry's `?normalize=true` produces clean `map<string, string>` syntax; Karapace produces `repeated .Foo.LabelsEntry labels = 1` with a dangling `LabelsEntry` nested message.

Both behaviours are deviations from Confluent Schema Registry. Schemas registered as plain `.proto` text strings are **not** affected — only the binary `FileDescriptorProto` registration path triggers these bugs.

Reproducer:

```bash
# Register a binary FileDescriptorProto containing a map<string,string> field
# (this is the format KafkaProtobufSerializer sends at runtime)
curl -s -X POST http://localhost:8081/subjects/map-test-value/versions \
  -H "Content-Type: application/vnd.schemaregistry.v1+json" \
  -d '{
    "schemaType": "PROTOBUF",
    "schema": "ImQKCk1hcE1lc3NhZ2USJwoGbGFiZWxzGAEgAygLMhcuTWFwTWVzc2FnZS5MYWJlbHNFbnRyeRotCgtMYWJlbHNFbnRyeRILCgNrZXkYASABKAkSDQoFdmFsdWUYAiABKAk6AjgBYgZwcm90bzM="
  }'

# Bug 1 — fetch without normalization: map_entry is absent from LabelsEntry
curl -s http://localhost:8081/subjects/map-test-value/versions/latest | jq -r .schema

# Bug 2 — fetch with normalization: entry-message form not converted to map<>
curl -s "http://localhost:8081/subjects/map-test-value/versions/latest?normalize=true" | jq -r .schema
```

Actual output (both with and without `?normalize=true`):

```proto
message MapMessage {
  repeated .MapMessage.LabelsEntry labels = 1;

  message LabelsEntry {
    string key = 1;
    string value = 2;
  }
}
```

---

## What did you expect to happen?

Karapace should match Confluent Schema Registry behaviour:

**Without `?normalize=true`** — `map_entry = true` preserved in the entry message:

```proto
message MapMessage {
  repeated .MapMessage.LabelsEntry labels = 1;

  message LabelsEntry {
    option map_entry = true;
    string key = 1;
    string value = 2;
  }
}
```

**With `?normalize=true`** — expanded form converted back to `map<K,V>` shorthand:

```proto
message MapMessage {
  map<string, string> labels = 1;
}
```

---

## What else do we need to know?

**Root causes:**

- **Bug 1** — `_deserialize_msg()` never reads `DescriptorProto.options`, so `MessageOptions` (including `map_entry`) is always discarded. The same gap in `_serialize_msgtype()` means binary round-trips also lose the option.

- **Bug 2** — `message_element_with_sorted_options()` in `proto_normalizations.py` only sorts options; it has no logic to detect the entry-message pattern and convert it back to `map<K,V>`. Confluent SR's normaliser explicitly detects messages with `map_entry = true` and emits `map<K,V>` syntax instead.

**Impact:**

- Code generators (`protoc`, `grpc_tools`) receiving the schema from the registry will generate `List<LabelsEntry>` instead of `Map<String, String>`, causing API mismatches at runtime.
- Reflection-based tools that rely on `map_entry: true` to identify map fields will misclassify them.
- Schema compatibility checks may incorrectly flag a map field against its expanded form as a type change.
- `?normalize=true` produces output that differs from Confluent SR for all binary-registered schemas containing map fields, breaking schema equivalence checks between the two implementations.

nathan-c:
Fixes #1319

## Summary

Two related bugs affect PROTOBUF schemas registered as a base64-encoded `FileDescriptorProto` binary (the format produced by Confluent Java, C#, and .NET SDK serialisers).

### Bug 1 — `map_entry` silently stripped on binary registration

`_deserialize_msg()` constructed `MessageElement` without reading `MessageOptions`, so `option map_entry = true` was silently dropped from any `map<K,V>` entry message. The same gap in `_serialize_msgtype()` meant binary round-trips also lost the option.

### Bug 2 — `?normalize=true` did not restore `map<K,V>` shorthand

Normalisation only sorted options; it had no logic to detect the expanded entry-message pattern and convert it back to `map<K,V>` syntax, diverging from Confluent Schema Registry's `?normalize=true` behaviour.

## Changes

**`src/karapace/core/protobuf/serialization.py`**
- Add `_deserialize_message_options()` which reads `map_entry` and `deprecated` from `MessageOptions` via `HasField()`
- Pass the result as `options=` to `MessageElement` in `_deserialize_msg()`
- Write `map_entry` and `deprecated` back onto `DescriptorProto.options` in `_serialize_msgtype()`

**`src/karapace/core/protobuf/proto_normalizations.py`**
- Add `_is_map_entry()` to detect synthetic entry messages by `option map_entry = true`
- Add `_restore_map_shorthand()` which replaces the `repeated .Foo.LabelsEntry` field with `map<K,V>` and suppresses the entry message from the nested types list
- Call `_restore_map_shorthand()` in `message_element_with_sorted_options()` before processing fields and nested types

**`README.rst`**
- Document map shorthand restoration as a normalisation behaviour alongside field ordering

## Tests

**`tests/unit/test_protobuf_binary_serialization.py`** — 3 new tests:
- `test_map_entry_option_preserved_on_deserialize`
- `test_map_entry_option_preserved_on_serialize_roundtrip`
- `test_map_entry_option_only_on_entry_message`

**`tests/unit/protobuf/test_protobuf_normalization.py`** — 4 new tests:
- `test_normalize_converts_explicit_map_entry_to_map_shorthand`
- `test_normalize_converts_binary_map_field_to_map_shorthand`
- `test_normalize_map_is_idempotent`
- `test_normalize_map_field_order_when_map_is_in_the_middle`

## Verification

After both fixes, `karapace-patched` output matches `confluent-sr` exactly across all 8 registration modes (text/binary × shorthand/explicit × with/without `?normalize=true`), confirmed locally.

### DIFF
--- a/src/karapace/core/protobuf/proto_normalizations.py
+++ b/src/karapace/core/protobuf/proto_normalizations.py
@@ -9,6 +9,7 @@ from collections.abc import Sequence
 from karapace.core.protobuf.enum_constant_element import EnumConstantElement
 from karapace.core.protobuf.enum_element import EnumElement
 from karapace.core.protobuf.extend_element import ExtendElement
+from karapace.core.protobuf.field import Field
 from karapace.core.protobuf.field_element import FieldElement
 from karapace.core.protobuf.group_element import GroupElement
 from karapace.core.protobuf.known_dependency import KnownDependency
@@ -27,6 +28,60 @@ import abc
 
 def sort_by_name(element: OptionElement) -> str:
     return element.name
+
+
+def _is_map_entry(message: MessageElement) -> bool:
+    """Return True if this message has option map_entry = true."""
+    return any(opt.name == "map_entry" and (opt.value is True or opt.value == "true") for opt in message.options)
+
+
+def _restore_map_shorthand(
+    message_element: MessageElement,
+) -> tuple[list[FieldElement], list[TypeElement]]:
+    """Convert binary-deserialized map fields back to map<K,V> shorthand syntax during normalisation.
+
+    A binary FileDescriptorProto expands map<K,V> into a synthetic nested entry
+    message (with option map_entry = true) plus a repeated field pointing to it.
+    This function replaces the repeated field with a map<K,V> field and removes the
+    synthetic entry message, matching Confluent SR's ?normalize=true behaviour.
+    """
+    entry_messages: dict[str, MessageElement] = {
+        nt.name: nt
+        for nt in message_element.nested_types
+        if isinstance(nt, MessageElement) and _is_map_entry(nt)
+    }
+    if not entry_messages:
+        return list(message_element.fields), list(message_element.nested_types)
+
+    new_fields: list[FieldElement] = []
+    for field in message_element.fields:
+        entry_name = field.element_type.split(".")[-1]
+        entry = entry_messages.get(entry_name)
+        if entry is not None and field.label == Field.Label.REPEATED:
+            entry_fields = {f.name: f for f in entry.fields}
+            key_field = entry_fields.get("key")
+            value_field = entry_fields.get("value")
+            if key_field is not None and value_field is not None:
+                new_fields.append(
+                    NormalizedFieldElement(
+                        location=field.location,
+                        label=None,  # map fields carry no label in .proto syntax
+                        element_type=f"map<{key_field.element_type}, {value_field.element_type}>",
+                        name=field.name,
+                        default_value=field.default_value,
+                        json_name=field.json_name,
+                        tag=field.tag,
+                        documentation=field.documentation,
+                        options=field.options,
+                    )
+                )
+                continue
+        new_fields.append(field)
+
+    new_nested_types: list[TypeElement] = [
+        nt for nt in message_element.nested_types if not (isinstance(nt, MessageElement) and nt.name in entry_messages)
+    ]
+    return new_fields, new_nested_types
 
 
 class NormalizedRpcElement(RpcElement):
@@ -178,10 +233,13 @@ def message_element_with_sorted_options(
     message_element: MessageElement, package: str, type_tree: TypeTree
 ) -> NormalizedMessageElement:
     sorted_options = None if message_element.options is None else list(sorted(message_element.options, key=sort_by_name))
+
+    shorthand_fields, shorthand_nested_types = _restore_map_shorthand(message_element)
+
     sorted_nested_types = [
-        type_element_with_sorted_options(nested_type, package, type_tree) for nested_type in message_element.nested_types
+        type_element_with_sorted_options(nested_type, package, type_tree) for nested_type in shorthand_nested_types
     ]
-    sorted_fields = [normalize_type_field_element(field, package, type_tree) for field in message_element.fields]
+    sorted_fields = [normalize_type_field_element(field, package, type_tree) for field in shorthand_fields]
     sorted_one_ofs = [one_ofs_with_sorted_options(one_of, package, type_tree) for one_of in message_element.one_ofs]
 
     return NormalizedMessageElement(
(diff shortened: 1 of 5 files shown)


nathan-c:
@muralibasani not sure if you are the right person to ping to get some attention on this PR?

nathan-c:
The smoke test errors appear to be unrelated to my changes. Are these expected?

src/karapace/core/protobuf/proto_normalizations.py
```diff
@@ -29,6 +30,58 @@ def sort_by_name(element: OptionElement) -> str:
...
+                        json_name=field.json_name,
+                        tag=field.tag,
+                        documentation=field.documentation,
+                        options=field.options,
+                    )
+                )
+                continue
+        new_fields.append(field)
+
+    new_nested_types: list[TypeElement] = [
```
muralibasani, line 95: The new_nested_types filter removes every map_entry nested message by name, but a field is only rewritten to map<> if key/value are both present and the field is REPEATED. If a map_entry message exists but the matching field is missing key/value (malformed) or isn't repeated, you remove the nested type but keep the repeated .Foo.LabelsEntry field pointing at a now-deleted message -> dangling type reference. Suggest only removing entry messages that were actually collapsed.

src/karapace/core/protobuf/proto_normalizations.py
```diff
@@ -29,6 +30,58 @@ def sort_by_name(element: OptionElement) -> str:
...
+    """
+    entry_messages: dict[str, MessageElement] = {
+        nt.name: nt for nt in message_element.nested_types if isinstance(nt, MessageElement) and _is_map_entry(nt)
+    }
+    if not entry_messages:
+        return list(message_element.fields), list(message_element.nested_types)
+
+    new_fields: list[FieldElement] = []
+    for field in message_element.fields:
+        entry_name = field.element_type.split(".")[-1]
```
muralibasani: If two different nested messages resolve to the same simple name (e.g. A.LabelsEntry and B.LabelsEntry both nested), the wrong entry could match. Minor/rare given entry messages are synthetic, but worth a guard on the field's full type path rather than just the last segment.

src/karapace/core/protobuf/proto_normalizations.py
nathan-c: addressed and added unit tests

src/karapace/core/protobuf/proto_normalizations.py
nathan-c: addressed and added unit tests

muralibasani:
@nathan-c thanks. Can you check the lint error pls?

nathan-c:
@muralibasani should be good now

muralibasani (approved):
LGTM, thanks for addressing the comments.

PR #1333, opened 2026-07-09
Oidc - Entra, pi parameterisation

muralibasani:

- Makes Karapace's OIDC wiring provider-agnostic via OIDC_PROVIDER and OIDC_* environment variables.
- The local Keycloak setup remains the default and is unchanged in behavior

References: #xxxxx

# Why this way


aindriu-aiven (approved):
LGTM Thanks

PR #1335, opened 2026-07-13
Improve OIDC testing

muralibasani:

The current OIDC testing seems insufficient. Potentially missing scenarios:

- Not all authenticated routes are tested.
- RBAC (roles, per-method role grants, missing/empty/wrong role claims, etc.)
- Various fields of the JWT token 
- Authorization-header parsing
- Unauthenticated and unauthorized probing against subject names, subject versions, etc. 

References: #xxxxx

# Why this way

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -39,7 +39,7 @@ RUN --mount=type=cache,target=/root/.cache/pip \
 # Karapace image, i.e. production.
 FROM builder AS karapace
 # Install protobuf compiler.
-ARG PROTOBUF_COMPILER_VERSION="3.21.12-3"
+ARG PROTOBUF_COMPILER_VERSION="3.21.12-3+deb12u1"
 RUN apt-get update \
  && apt-get install --assume-yes --no-install-recommends \
     protobuf-compiler=$PROTOBUF_COMPILER_VERSION \
--- a/container/keycloak/realm-export.json
+++ b/container/keycloak/realm-export.json
@@ -44,6 +44,34 @@
           }
         }
       ]
+    },
+    {
+      "clientId": "karapace-client-limited",
+      "enabled": true,
+      "protocol": "openid-connect",
+      "publicClient": false,
+      "redirectUris": ["*"],
+      "standardFlowEnabled": false,
+      "directAccessGrantsEnabled": false,
+      "serviceAccountsEnabled": true,
+      "defaultClientScopes": ["email", "profile", "roles"],
+      "optionalClientScopes": ["address", "phone"],
+      "protocolMappers": [
+        {
+          "name": "aud-claim",
+          "protocol": "openid-connect",
+          "protocolMapper": "oidc-hardcoded-claim-mapper",
+          "consentRequired": false,
+          "config": {
+            "claim.name": "aud",
+            "claim.value": "karapace-audience",
+            "claim.type": "String",
+            "id.token.claim": "true",
+            "access.token.claim": "true",
+            "userinfo.token.claim": "true"
+          }
+        }
+      ]
     }
   ],
   "users": [
@@ -75,6 +103,14 @@
           "karapace.config_subject:update",
           "karapace.config_global:update"
         ]
+      }
+    },
+    {
+      "username": "service-account-karapace-client-limited",
+      "enabled": true,
+      "serviceAccountClientId": "karapace-client-limited",
+      "clientRoles": {
+        "karapace-client": ["karapace.schema:read", "karapace.schema:write"]
       }
     }
   ],
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -171,7 +171,8 @@ def oidc_token():
     token_url = os.environ.get("OIDC_TOKEN_URL")
     client_secret = os.environ.get("OIDC_CLIENT_SECRET")
     if not token_url or not client_secret:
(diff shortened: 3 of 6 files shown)


container/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN --mount=type=cache,target=/root/.cache/pip \
 # Karapace image, i.e. production.
 FROM builder AS karapace
 # Install protobuf compiler.
-ARG PROTOBUF_COMPILER_VERSION="3.21.12-3"
+ARG PROTOBUF_COMPILER_VERSION="3.21.12-3+deb12u1"
```
muralibasani, line 42: (unrelated) Had to update this proto version, as previous one was removed.

container/keycloak/realm-export.json
```diff
@@ -44,6 +44,34 @@
           }
         }
       ]
+    },
+    {
+      "clientId": "karapace-client-limited",
```
muralibasani, line 49: Adding this to test the deny on authZ

tests/unit/test_oidc.py
```diff
@@ -554,10 +555,15 @@ def test_middleware_skip_paths_bypass_auth(monkeypatch):
 # ---------------------------------------------------------------------------
 
 
+# Encode at runtime so no literal ``Basic <base64>`` credential lands in source
+# (a hardcoded one trips secret scanners).
+_BASIC_HEADER = "Basic " + base64.b64encode(b"user:pass").decode()
```
muralibasani, line 560: This change is not related to this PR, but to remove the basic-auth secret hardcoding.

container/Dockerfile
aindriu-aiven: It looks like we are well behind on protobuf compiler versions, should we open a ticket to address this and get up closer to the latest?

Image

tests/e2e/conftest.py
```diff
@@ -137,37 +137,62 @@ def fixture_new_topic(admin_client: KafkaAdminClient) -> NewTopic:
...
-        client_secret = os.environ.get("OIDC_CLIENT_SECRET")
-        if not token_url or not client_secret:
-            pytest.skip(f"OIDC_PROVIDER={provider} requires OIDC_TOKEN_URL and OIDC_CLIENT_SECRET")
-        verify_tls = os.environ.get("OIDC_VERIFY_TLS", "true").strip().lower() not in ("0", "false", "no", "off")
+        return _keycloak_client_credentials_token(client_id, client_secret=os.environ.get("OIDC_CLIENT_SECRET"))
 
+    token_url = os.environ.get("OIDC_TOKEN_URL")
+    client_secret = os.environ.get("OIDC_CLIENT_SECRET")
+    if not token_url or not client_secret:
+        pytest.skip(f"OIDC_PROVIDER={provider} requires OIDC_TOKEN_URL and OIDC_CLIENT_SECRET")
```
aindriu-aiven: Should this be a failure, someone might write a test and pipeline passes but they haven't got the setup correct?

tests/e2e/conftest.py
```diff
@@ -137,37 +137,62 @@ def fixture_new_topic(admin_client: KafkaAdminClient) -> NewTopic:
...
+
+    scope = os.environ.get("OIDC_SCOPE", "openid")
     data = {"grant_type": "client_credentials", "client_id": client_id, "client_secret": client_secret, "scope": scope}
     response = requests.post(token_url, data=data, verify=verify_tls)
     response.raise_for_status()
     return response.json()["access_token"]
 
 
+@pytest.fixture(scope="function", name="oidc_token_limited")
+def oidc_token_limited():
```
aindriu-aiven: I'd consider renaming this function with the schema_write and schema_read so people will know what the limitations are when reading the function. But its not a deal breaker.

tests/unit/test_oidc.py
```diff
@@ -793,3 +898,214 @@ def test_authorize_request_silently_denies_when_method_unmapped():
...
+
+
+@pytest.fixture
+def build_real_middleware(rsa_keypair):
+    """Build an OIDCMiddleware whose JWKS client returns the in-test public key,
+    so validate_jwt runs the real PyJWT decode against real signatures."""
+    _, public_key = rsa_keypair
+
+    def _build(**config_overrides) -> OIDCMiddleware:
+        with patch("karapace.api.oidc.middleware.PyJWKClient") as mock_client:
```
HelenMel: It is quite dangerous to patch fully a client we develop internally. Is it possible to create a stub or a fixture instead?

HelenMel (commented):
Overall looks really good. One small comment

aindriu-aiven (commented):
A few nits but nothing major looks really great

container/Dockerfile
muralibasani: Indeed, I could address this in the https://github.com/Aiven-Open/karapace/pull/1336 PR.

tests/e2e/conftest.py
muralibasani: Good one. Fixed. Changed to pytest.fail.

tests/e2e/conftest.py
muralibasani: Done. Renamed to oidc_token_schema_read_write

tests/unit/test_oidc.py
muralibasani: Agree, mocking the whole client is not good. Now it returns a small sign key instead of mock obj.

HelenMel (approved):

PR #1337, opened 2026-07-16
Refactoring - Rename OIDCMiddleware to OIDCTokenValidator and split auth gate

muralibasani:

- Rename src/karapace/api/oidc/middleware.py -> validator.py
- Split setup_middlewares into two single-purpose middlewares:
    - oidc_auth_gate (registered last -> outermost, rejects unauthenticated requests before routing) and 
    - set_content_types (Content-Type only).

Pure refactoring.

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/api/middlewares/__init__.py
+++ b/src/karapace/api/middlewares/__init__.py
@@ -12,7 +12,7 @@ from prometheus_client import Counter
 from prometheus_fastapi_instrumentator import Instrumentator
 from prometheus_fastapi_instrumentator.metrics import Info, request_size, requests, response_size
 
-from karapace.api.oidc.middleware import OIDCMiddleware, TokenExpiredError
+from karapace.api.oidc.validator import OIDCTokenValidator, TokenExpiredError
 from karapace.core.auth import AuthenticationError
 from karapace.core.config import Config
 import logging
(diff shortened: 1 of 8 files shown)


src/karapace/api/middlewares/__init__.py
```diff
@@ -66,31 +66,43 @@ def _authenticate_and_authorize(request: Request, config: Config, oidc_middlewar
     return None
 
 
+# Paths that bypass the OIDC auth gate: API docs endpoints are always public.
+_DOCS_PATHS = frozenset({"/docs", "/docs/oauth2-redirect", "/redoc", "/openapi.json"})
```
aindriu-aiven, line 70: does openapi.json give a swagger page or is it just the api spec?

aindriu-aiven (approved):
LGTM

src/karapace/api/middlewares/__init__.py
muralibasani: Correct, it is just raw spec, and there is no swagger page. If we run swagger-ui and provide this spec, it can give us the swagger page.

muralibasani:
@aindriu-aiven thanks for the review.

PR #1338, opened 2026-07-21
Add json draft 09 and 12 validation & storage support

muralibasani:

Karapace hardcoded Draft7Validator for all JSON Schema parsing, so schemas declaring a newer $schema were silently validated against the Draft-7 metaschema.

This PR detects the draft from $schema and routes to the matching validator (Draft7 / Draft201909 / Draft202012), while keeping Draft-7 as the default for schemas with no $schema (or an unrecognized one) — preserving all existing behavior.

- 2019-09 & 2020-12 schemas register
- draft-specific keywords validate
- no $schema → Draft-7 (no regression)
- existing Draft-7 schemas unchanged

Scope: validation + storage only

References: #xxxxx

# Why this way

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -840,10 +840,9 @@ migrating between drafts.
 Supported drafts and ``$schema`` URIs
 -------------------------------------
 
-Karapace validates every JSON Schema as **Draft-7**, regardless of the ``$schema`` value declared in the document. This
-matches the behaviour of Confluent Schema Registry, which also standardizes on Draft-7. The ``$schema`` keyword is
-currently **not used to select a validator** — it is accepted and stored as part of the schema, but does not change how
-the schema is validated or compared.
+Karapace selects the JSON Schema draft from the ``$schema`` keyword and validates the schema against that draft's
+metaschema. When ``$schema`` is absent or its value is not recognized, the schema is validated as **Draft-7**, which
+preserves the previous behaviour and matches the default used by Confluent Schema Registry.
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies = [
   "confluent-kafka",
   "cryptography",
   "fastapi[standard]",
-  "jsonschema",
+  "jsonschema>=4.18",
```
aindriu-aiven: Should we give a range or a specific version or are you happy to have it greater then or equal to here?

tests/integration/test_jsonschema.py
```diff
@@ -1225,3 +1225,38 @@ async def test_same_jsonschema_must_have_same_id(
         )
         assert second_res.status_code == 200
         assert first_id == second_res.json()["id"]
+
+
+async def test_schemaregistry_register_retrieve_list_draft_2020_12(registry_async_client: Client) -> None:
+    """A Draft 2020-12 schema (declared via $schema, using prefixItems) can be registered,
```
aindriu-aiven: Should we have a draft 2019 test as well?

README.rst
aindriu-aiven: validated as draft 7, unknown keywords ignored? Like draft 4 + 6 ?

aindriu-aiven (approved):
LGTM some minor nits I will let @muralibasani answer or alter before merging

pyproject.toml
muralibasani: The lowest minimum is good, but also added an upper limit now, to be on safe side for sometime.

tests/integration/test_jsonschema.py
muralibasani: Good one, added 2019 now.

README.rst
muralibasani: Ah, that readme was stale after the new changes. Updated it now.

muralibasani:
> LGTM some minor nits I will let @muralibasani answer or alter before merging

Thanks for the review. Addressed the comments.

aindriu-aiven (approved):
LGTM

PR #1340, opened 2026-07-24
Implement JSON Schema Draft 2019-09 / 2020-12 compatibility checking

muralibasani:

**Summary :** 
Karapace's JSON Schema compatibility engine was written entirely around Draft-7 keyword shapes, so newer-draft keywords (prefixItems, $defs, dependentRequired, unevaluatedProperties, $dynamicRef) were silently ignored during compatibility checks — a false-compatible hazard. This PR makes compatibility checking correct for Draft 2019-09 / 2020-12.

1. Canonicalize renamed/relocated keywords → Draft-7 shapes (compatibility/jsonschema/utils.py)
2. Fail-closed guard for keywords the engine can't evaluate (schema_compatibility.py)
3. Unit and intgrtn tests supporting cross-draft compatible/breaking evolution, and fail-closed reject

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/core/compatibility/jsonschema/utils.py
+++ b/src/karapace/core/compatibility/jsonschema/utils.py
@@ -9,14 +9,70 @@ from karapace.core.typing import JsonSchemaValidator
 from typing import Any, TypeVar, Union
 
 import re
+import warnings
 
 T = TypeVar("T")
 JSONSCHEMA_TYPES = Union[Instance, Subschema, Keyword, type[BooleanSchema]]
+
+# Newer-draft (2019-09 / 2020-12) keyword names that are not part of the Draft-7 `Keyword` enum.
+PREFIX_ITEMS = "prefixItems"
+DEFS = "$defs"
+DEFINITIONS = "definitions"
+DEPENDENT_REQUIRED = "dependentRequired"
+
+# Keywords the compatibility engine cannot yet evaluate. Their presence makes a Draft-7-shaped
+# comparison unreliable, so compatibility checking fails closed rather than returning a possibly
+# wrong verdict.
+UNSUPPORTED_COMPAT_KEYWORDS = frozenset(
+    {
+        "$dynamicRef",
+        "$dynamicAnchor",
+        "$recursiveRef",
+        "$recursiveAnchor",
+        "unevaluatedProperties",
+        "unevaluatedItems",
+        "$vocabulary",
+    }
+)
+
+# Keywords whose *values are maps keyed by names chosen by the schema author* (property names,
+# definition names, patterns). Those keys are not schema keywords, so the scanner must recurse into
+# the values only — never treat the keys as keywords.
+_NAME_KEYED_MAP_KEYWORDS = frozenset(
+    {
+        "properties",
+        "patternProperties",
+        "$defs",
+        "definitions",
+        "dependentSchemas",
+        "dependentRequired",
+    }
+)
+
+# Keywords whose values are arbitrary instance data, not subschemas. The scanner must not descend
+# into them, otherwise a user value that happens to be an object with a keyword-like key would be
+# misread as a schema keyword.
+_DATA_VALUE_KEYWORDS = frozenset({"enum", "const", "default", "examples"})
 
 
 def normalize_schema(validator: JsonSchemaValidator) -> Any:
     original_schema = validator.schema
     return normalize_schema_rec(validator, original_schema)
+
+
+def _resolver_of(validator: JsonSchemaValidator) -> Any:
+    """Return the validator's ref resolver.
+
+    ``Validator.resolver`` is deprecated as of jsonschema 4.18 in favour of the ``referencing``
+    library, but ``normalize_schema`` still relies on its ``push_scope``/``pop_scope``/``resolve``
+    API. The pin ``jsonschema>=4.18,<5`` guarantees the attribute stays available (deprecations are
+    not removed within a major series), so we access it in one place and silence the warning here.
+
+    TODO: migrate reference resolution to the ``referencing`` library and drop this shim.
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
+        return validator.resolver
 
 
 def normalize_schema_rec(validator: JsonSchemaValidator, original_schema: Any) -> Any:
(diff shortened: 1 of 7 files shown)


src/karapace/core/compatibility/jsonschema/utils.py
```diff
@@ -9,24 +9,80 @@
...
+
+    ``Validator.resolver`` is deprecated as of jsonschema 4.18 in favour of the ``referencing``
+    library, but ``normalize_schema`` still relies on its ``push_scope``/``pop_scope``/``resolve``
+    API. The pin ``jsonschema>=4.18,<5`` guarantees the attribute stays available (deprecations are
+    not removed within a major series), so we access it in one place and silence the warning here.
+
+    TODO: migrate reference resolution to the ``referencing`` library and drop this shim.
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
```
aindriu-aiven, line 74: Do we have a ticket to migrate away from it? So we don't lose track of it?

src/karapace/core/compatibility/jsonschema/utils.py
```diff
@@ -54,6 +112,72 @@ def normalize_schema_rec(validator: JsonSchemaValidator, original_schema: Any) -
...
+    between schemas written in different drafts become well-defined (a single canonical form).
+    Keywords that cannot be canonicalized safely are handled by a fail-closed guard before
+    compatibility checking, not here.
+    """
+    # 2020-12 array model: `prefixItems` holds the positional (tuple) schemas that Draft-7 spells as
+    # the list form of `items`; when `prefixItems` is present, `items` holds the "additional items"
+    # schema that Draft-7 spells as `additionalItems`.
+    if PREFIX_ITEMS in schema:
+        prefix_items = schema.pop(PREFIX_ITEMS)
+        additional_items = schema.pop(Keyword.ITEMS.value, None)
```
aindriu-aiven, line 129: Given schema.pop is deprecated is there an alternative we can start using here so the future migration away could be smoother and a pattern already setup for others to follow?

src/karapace/core/compatibility/jsonschema/utils.py
```diff
@@ -54,6 +112,72 @@ def normalize_schema_rec(validator: JsonSchemaValidator, original_schema: Any) -
...
+            if key in UNSUPPORTED_COMPAT_KEYWORDS:
+                found.add(key)
+
+            if key in _DATA_VALUE_KEYWORDS:
+                # Value is arbitrary instance data, never a subschema — do not descend.
+                continue
+            if key in _NAME_KEYED_MAP_KEYWORDS and isinstance(value, dict):
+                # Keys here are author-chosen names; only the values are subschemas.
+                for subschema in value.values():
+                    found |= find_unsupported_compat_keywords(subschema)
```
aindriu-aiven, line 172: is there potentially a recursion issue here that we need to guard against?

tests/integration/test_jsonschema.py
```diff
@@ -1271,3 +1271,99 @@ async def test_schemaregistry_register_retrieve_list_newer_drafts(
...
+
+    await _register(registry_async_client, subject, base)
+    config_res = await registry_async_client.put(f"config/{subject}", json={"compatibility": "BACKWARD"})
+    assert config_res.status_code == 200
+
+    res = await registry_async_client.post(
+        f"subjects/{subject}/versions",
+        json={"schema": json.dumps({**base, **unsupported_fragment}), "schemaType": SchemaType.JSONSCHEMA.value},
+    )
+    assert res.status_code == 409, res.json()
```
aindriu-aiven, line 1357: Can you also check the message returned, which i think lists the keywords that are not supported?

tests/integration/test_jsonschema.py
```diff
@@ -1271,3 +1271,99 @@ async def test_schemaregistry_register_retrieve_list_newer_drafts(
...
+
+
+async def test_schemaregistry_unsupported_keyword_allowed_under_none(registry_async_client: Client) -> None:
```
aindriu-aiven: I think you can also add the same parameterization here from the previous test.

aindriu-aiven (commented):
looks great a few questions and suggestions.

src/karapace/core/compatibility/jsonschema/utils.py
muralibasani: Migrating off the deprecated resolver is a full rewrite of the $ref resolution, which every json schema compatibility checks on. That would be a very high risk change, and moreover it neds jsonschema lib to be >5.
So we will not do this in the near future too. 
We have a very handful of libs with pinned versions in pyproj.toml, which we constantly look for. So I am sure we don't miss this one.

muralibasani: But good we pointed it.

src/karapace/core/compatibility/jsonschema/utils.py
muralibasani: `schema` here is a plain dict, and dict.pop(key, default) is a standard one and non-deprecated Python built-in. I Think there is no migration concern for it. Comparing it to the previous comment, probably I mis-read it?

muralibasani: But if your point is about mutating the dict in place with pop feels risky, I can look into it

src/karapace/core/compatibility/jsonschema/utils.py
muralibasani: Good one. Actually the input is the raw parsed json schema and the scanner treats $ref as an opaque string rather than following it, so recursion always ends. But the existing normalize_schema_rec has a recursive compatibility check but that will hit Python's max recursion depth. So this removes the RecursionError possibility entirely for this function.

muralibasani:
> looks great a few questions and suggestions.

Thanks for the thorough review. Added a few comments and pushed a commit for tests.

src/karapace/core/compatibility/jsonschema/utils.py
aindriu-aiven: No I think it is ok, apologies, I read about the deprecated versions we were dependent on ``push_scope``/``pop_scope``/``resolve`` and I saw 'pop' I should have double checked and I would have seen pop_scope.

aindriu-aiven (approved):
LGTM

PR #1341, opened 2026-07-24
faster Avro encode/decode (compiled decoder + single-pass write validation)

smirnoffs:
Two performance changes to the Avro path in `src/karapace/core/serialization.py`, plus tests for them. The on-the-wire format is unchanged and encoder output is byte-identical.

The REST proxy was observed to time out decoding records with multi-way unions. The cost is in `avro.io.DatumReader.read`, which always runs the writer/reader schema-resolution machinery even though on this path the writer schema *is* the reader schema: a `readers_schema.match(writers_schema)` at every node, an O(N) linear scan over union branches (`for s in schemas: if s.match(...)`, which is why multi-way unions blow up) and millions of `isinstance`/abc `__instancecheck__` dispatches, followed by a second full recursive walk (`_jsonify_avro_payload`) to base64-encode `bytes`.

**Decode: compile each schema once into a tree of decoder closures** (new module `avro_fast_decoder.py`). Type dispatch is resolved at build time, unions decode by direct branch index in O(1), and `bytes`/`fixed` are base64-encoded inline, folding in the second pass. Byte-level primitives (varint, UTF-8, float/double, decimal, date/time) are still delegated to avro's own `BinaryDecoder`, so no decoder internals are re-implemented and the decoded values are unchanged. The compiled decoder is cached on the `TypedSchema`, which the serializer already caches per schema id. Because the decoders are immutable and stateless, the per-thread `DatumReader` cache (`threading.Lock` + `WeakKeyDictionary`) is no longer needed and is removed; recursive schemas break the compile-time cycle with a lazy placeholder registered in a build memo.

One behaviour difference is intentional: a negative enum symbol index or union branch index is rejected with `SchemaResolutionException`, which `deserialize` turns into `InvalidPayload`. avro's reader only checks the upper bound, so such an index silently selected a symbol or branch counting from the end of the list, or raised `IndexError`.

**Encode: drop a redundant validation pass in `write_value`.** The datum tree was validated twice per message, once via `avro.io.validate()` to choose the flatten-unions branch and again inside `DatumWriter.write()`. The value is now written as is and tagged unions are flattened only on `AvroTypeException`, so validation runs once; the same predicate decides, so untagged/tagged backwards compatibility and the output bytes are unchanged. `DatumWriter.write()` validates before emitting anything, but the encoder can raise `AvroTypeException` by itself after part of the value is already written (a decimal whose exponent does not fit the schema scale raises `AvroOutOfScaleException`), so the retry rewinds and truncates the buffer to where the first attempt started and can never append a value to a partial one.

Tests: Hypothesis generates schema trees (records, unions, arrays, maps, enums, fixed and every supported logical type) together with data valid under them, and asserts that the compiled decoders agree with `DatumReader` on the decoded value, on the number of bytes consumed, and on the failure mode for random and truncated payloads. A round-trip table states the expected value per type independently of avro, and further properties cover decoder reuse across concatenated values, arbitrary array/map block splits and recursive schemas. Statement and branch coverage of `karapace.core.avro_fast_decoder` is 100%.

Decode benchmark of `read_value`, median total over repeated runs, comparing against the previous path with its cached `DatumReader` (i.e. the already-optimised baseline):

| scenario | decodes | before (µs) | after (µs) | speedup |
|---|---:|---:|---:|---:|
| flat record (3 scalar fields) | 4000 | 15,710 | 3,831 | 4.1× |
| optional fields + array + map + nested record | 3000 | 72,972 | 16,122 | 4.5× |
| multi-way-union record, small array | 1500 | 171,306 | 29,329 | 5.8× |
| multi-way-union record, large array | 200 | 209,066 | 33,887 | 6.2× |
| bytes + array&lt;bytes&gt; (base64 path) | 2500 | 23,259 | 5,692 | 4.1× |
| **total** | **11,200** | **489,571** | **88,367** | **5.5×** |

That is ~23k → ~127k decodes/s. Encoding is ≈1.2× faster (≈17% less time) on a 40-key map-of-union record, and byte-identical.

### DIFF
--- a/src/karapace/core/avro_fast_decoder.py
+++ b/src/karapace/core/avro_fast_decoder.py
@@ -1,0 +1,247 @@
+"""
+karapace - compile an Avro schema into a tree of decoder closures for single-schema decoding
+
+Copyright (c) 2026 Aiven Ltd
+See LICENSE for details
+"""
+
+from __future__ import annotations
+
+from avro.io import BinaryDecoder
+from collections.abc import Callable
+from typing import Any, cast
+
+import avro.constants
+import avro.errors
+import avro.schema
+import base64
+
+_b64encode = base64.b64encode
+
+_DATE = avro.constants.DATE
+_TIME_MILLIS = avro.constants.TIME_MILLIS
+_TIME_MICROS = avro.constants.TIME_MICROS
+_TIMESTAMP_MILLIS = avro.constants.TIMESTAMP_MILLIS
+_TIMESTAMP_MICROS = avro.constants.TIMESTAMP_MICROS
+
+Decoder = Callable[[BinaryDecoder], Any]
+
+
+class _LazyDecoder:
+    __slots__ = ("fn",)
+
+    def __init__(self) -> None:
+        self.fn: Decoder | None = None
+
+    def __call__(self, decoder: BinaryDecoder) -> Any:
+        return self.fn(decoder)  # type: ignore[misc]
+
+
+def _decimal_props(schema: avro.schema.Schema) -> tuple[int, int]:
+    """Return the (precision, scale) of a decimal schema.
+
+    Only call this when ``schema.logical_type == "decimal"``: avro's schema parsing sets the
+    logical type only after validating precision and scale (see avro.schema.DecimalLogicalSchema),
+    and warns and leaves the logical type unset for an invalid annotation. So the properties need
+    no revalidation here, they are known to be integers.
+    """
+    return cast(int, schema.get_prop("precision")), cast(int, schema.get_prop("scale"))
+
+
+def _build_bytes_decoder(schema: avro.schema.Schema) -> Decoder:
+    if getattr(schema, "logical_type", None) == "decimal":
+        precision, scale = _decimal_props(schema)
+
+        def decode_decimal(decoder: BinaryDecoder, _p: int = precision, _s: int = scale) -> Any:
+            return decoder.read_decimal_from_bytes(_p, _s)
+
+        return decode_decimal
+
+    def decode_bytes(decoder: BinaryDecoder) -> str:
+        return _b64encode(decoder.read_bytes()).decode("ascii")
+
+    return decode_bytes
+
+
+def _build_fixed_decoder(schema: avro.schema.FixedSchema) -> Decoder:
+    size = schema.size
+    if getattr(schema, "logical_type", None) == "decimal":
+        precision, scale = _decimal_props(schema)
+
+        def decode_decimal(decoder: BinaryDecoder, _p: int = precision, _s: int = scale, _sz: int = size) -> Any:
+            return decoder.read_decimal_from_fixed(_p, _s, _sz)
+
+        return decode_decimal
+
+    def decode_fixed(decoder: BinaryDecoder, _size: int = size) -> str:
+        return _b64encode(decoder.read(_size)).decode("ascii")
+
+    return decode_fixed
+
+
+def _build_enum_decoder(schema: avro.schema.EnumSchema) -> Decoder:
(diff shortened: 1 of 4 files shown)


src/karapace/core/avro_fast_decoder.py
```diff
@@ -0,0 +1,255 @@
...
+    return decode_record
+
+
+def _build_union_decoder(schema: avro.schema.UnionSchema, memo: dict[int, Decoder]) -> Decoder:
+    branch_decoders = tuple(build_decoder(branch, memo) for branch in schema.schemas)
+    n = len(branch_decoders)
+
+    def decode_union(decoder: BinaryDecoder, _branches: tuple = branch_decoders, _n: int = n) -> Any:
+        index = decoder.read_long()
+        if index >= _n:
```
muralibasani: Only upper bounds seem to be checked.  Can we add a index < 0 too ?
Also to the enum decoder can be applied.

tests/unit/test_serialization.py
```diff
@@ -854,3 +854,54 @@ async def test_get_schema_cache_unauthenticated_path_unchanged(reset_sr_authoriz
     await sr_client.get_schema(subject)
     await sr_client.get_schema(subject)
     assert sr_client.client.get.call_count == 1
+
```
muralibasani, line 864: Can we add a couple of tests too. 

There is no systematic decode test covering: all primitives, fixed, enum (incl. default), nested/recursive records, deeply nested unions, arrays/maps with multi-block counts, and every logical type (decimal on bytes and fixed, date, time-millis/micros, timestamp-millis/micros, and any uuid/duration if used). 

Also a test to generate random records, encode, decode with both the new decoder and avro.io.DatumReader, assert equal. Without that, silent wrong decodes on some schema shape are the likeliest field failure.

src/karapace/core/serialization.py
```diff
@@ -461,22 +461,6 @@ def _read_avro_value(self, schema_id: SchemaId, schema: TypedSchema, bio: io.Byt
         return read_value(self.config, schema, bio, avro_reader=self._get_avro_reader(schema_id, schema))
```
muralibasani: I think this needs some clean up/refactor.  May be delete _get_avro_reader, and remove _avro_readers_lock, _avro_readers_by_thread from __init__ and simplify _read_avro_value to: return read_value(self.config, schema, bio)

After this the DatumReader import can go too.

tests/unit/test_serialization.py
```diff
@@ -491,42 +493,78 @@ async def fake_to_thread(func, *args, **kwargs):
...
+                        },
+                    },
+                    {"name": "props", "type": {"type": "map", "values": ["null", "string"]}},
+                ],
+            }
+        ),
+    )
     mock_registry_client = Mock()
     get_latest_schema_future = asyncio.Future()
-    get_latest_schema_future.set_result((1, MAP_UNION_AVRO_SCHEMA, Versioner.V(1)))
```
RyanSkraba, line 500: Nitpick, we can remove `MAP_UNION_AVRO_SCHEMA` now (and I'm happier to see the schema close to where it's being used).

RyanSkraba (dismissed):
This LGTM. It's pretty hefty for us to be rewriting the binary decoder for Avro though and we should probably make it a priority to contribute this upstream.

tests/unit/test_serialization.py
smirnoffs: Thanks. Now it's removed.

muralibasani:
@smirnoffs can you pls rebase?

smirnoffs:
@muralibasani rebased and pushed

muralibasani (approved):

PR #1345, opened 2026-07-28
fix(protobuf): allow empty line comments

### LINKED ISSUE #1343: Protobuf registration returns 500 when a block comment contains blank lines
# What happened?

Registering a Protobuf schema whose block comment contains a blank line returns HTTP 500 after approximately 60 seconds.

The schema is entirely ASCII and does not use references. Removing only the blank line from the comment makes the same schema register successfully, so ordinary line breaks and multiline comments themselves are not the trigger.

Minimal reproducer:

```proto
syntax = "proto3";

/*
  Example Event

  Overview
  This is an example event.
*/

message Event {
  string id = 1;
}
```

Example request:

```bash
curl -i -X POST \
  -H 'Content-Type: application/vnd.schemaregistry.v1+json' \
  --data-binary '{"schemaType":"PROTOBUF","schema":"syntax = \"proto3\";\n\n/*\n  Example Event\n\n  Overview\n  This is an example event.\n*/\n\nmessage Event {\n  string id = 1;\n}\n"}' \
  http://localhost:8081/subjects/protobuf-empty-comment-line/versions
```

Observed result:

```text
HTTP 500 after approximately 60 seconds
Internal Server Error
```

As a control, this version registers successfully:

```proto
/*
  Example Event
  Overview
  This is an example event.
*/
```

The differential test results were:

| Schema | Result |
|---|---|
| ASCII multiline block comment without blank documentation lines | HTTP 200 |
| Same ASCII block comment with a blank documentation line | HTTP 500 after 60.09 seconds |

Blank lines are valid inside Protobuf block comments, so the schema should register successfully.

At minimum, an invalid schema should be rejected immediately with a client-facing validation error rather than being written and then timing out with HTTP 500.

Test environment:

- Karapace `6.2.0`
- Kafka `3.9.1`
- Schema Registry HTTP API

The relevant logic remains unchanged in Karapace `6.2.1` and the current `main` branch.

## Root cause

The original block comment is parsed successfully. The failure occurs after Karapace renders the parsed schema and the schema reader replays it from the schema topic.

[`append_documentation`](https://github.com/Aiven-Open/karapace/blob/6.2.1/src/karapace/core/protobuf/utils.py#L14-L26) renders every documentation line with a `// ` prefix. A blank line is therefore rendered as an empty line comment:

```proto
// Example Event
// 
// Overview
// This is an example event.
```

During schema-topic replay, [`SyntaxReader.read_comment`](https://github.com/Aiven-Open/karapace/blob/6.2.1/src/karapace/core/protobuf/syntax_reader.py#L213-L262) skips the single space after `//` and obtains an empty string. The final `if not result` check then rejects that valid empty comment with:

```text
unexpected '/'
```

This creates a parser/formatter round-trip inconsistency:

```text
parse(original schema)                succeeds
render(parsed schema)                 emits an empty `// ` comment
parse(rendered schema) during replay  fails
```

The schema record has already been written to the schema topic, but the schema reader does not add it to the in-memory database. The request then waits for the reader and eventually fails with the schema-reader timeout, producing HTTP 500.

This is separate from the non-ASCII parsing issue in #1171 and #1317. It is related to #913 only at the higher level that a Protobuf schema is accepted initially but fails during schema-topic replay; #913 has a different parser trigger.

## Workaround

Remove blank lines from Protobuf documentation comments, or replace them with a non-empty ASCII separator such as `---`.

## Suggested fix

Allow `SyntaxReader.read_comment` to accept valid empty `//` and `// ` comments, and add a parser/formatter round-trip regression test covering:

- A block comment containing blank lines.
- An empty `//` comment.
- `parse(render(parse(schema)))` for documentation containing blank lines.

Gezi-lzq:

Fix Protobuf schema registrations that return HTTP 500 when documentation comments contain blank lines.

Karapace renders block comments as multiple `//` comments. A blank documentation line is rendered as `// `. When the schema reader replays the schema from the schema topic, `SyntaxReader.read_comment()` parses that line as an empty string, but the subsequent `if not result` check incorrectly treats the valid empty comment as an invalid `/`.

This PR:

- Accepts valid empty `//` and `// ` comments.
- Adds a parser/formatter round-trip test for a block comment containing blank lines.
- Adds a Schema Registry integration test covering mixed block and single-line comments, blank documentation lines, registration, and exact retrieval.
- Adds a regression test confirming that a standalone `/` is still rejected.

Fixes #1343

# Why this way

Comment validity should be determined by the delimiter following `/`, not by whether the comment body is non-empty.

The previous `if not result` check conflated two cases:

- `/` is followed by neither `*` nor `/`, which is an invalid delimiter.
- `//` is matched but its body is empty, which is a valid comment.

Moving the error handling to the delimiter branch's `else` restores delimiter-based parsing and matches the behavior of the Square Wire parser from which this code was ported.

The production change is limited to one control-flow branch and does not alter parsing of non-empty comments. Completely empty block comments remain outside the scope of this fix.

The successful POST assertion is the key regression check for the schema-reader replay path: registration returns HTTP 200 only after the reader has replayed the schema record successfully. Before this fix, replay failed and the POST timed out with HTTP 500. The subsequent GET asserts the complete canonical schema content, including the blank documentation lines.

# Testing

- `tests/unit/protobuf/test_proto_parser.py`: 78 passed
- `tests/integration/test_schema_protobuf.py::test_protobuf_schema_with_blank_documentation_lines`: 2 passed (default and `use_protobuf_formatter=True`)
- `ruff check`: passed
- `ruff format --check`: passed
- `mypy src/karapace/core/protobuf/syntax_reader.py`: passed

### DIFF
--- a/src/karapace/core/protobuf/syntax_reader.py
+++ b/src/karapace/core/protobuf/syntax_reader.py
@@ -257,7 +257,7 @@ class SyntaxReader:
                     self.newline()
                     break
             result = self.data[start : self.pos - 1]
-        if not result:
+        else:
             self.unexpected("unexpected '/'")
         return result
 
--- a/tests/integration/test_schema_protobuf.py
+++ b/tests/integration/test_schema_protobuf.py
@@ -186,6 +186,34 @@ async def test_protobuf_schema_normalization(registry_async_client: Client) -> N
 
 
 @pytest.mark.parametrize("registry_cluster", [{"config": {}}, {"config": {"use_protobuf_formatter": True}}], indirect=True)
+async def test_protobuf_schema_with_blank_documentation_lines(registry_async_client: Client) -> None:
+    subject = create_subject_name_factory("test_protobuf_blank_documentation_lines")()
+    schema = trim_margin(
+        """
+        |syntax = "proto3";
+        |
+        |/*
+        | Some Important Event
+        |
+        | Overview
+        | What this event is for.
+        |*/
+        |message Test {}
+        """
+    )
+
+    res = await registry_async_client.post_subjects_versions(
+        subject=subject,
+        json={"schemaType": "PROTOBUF", "schema": schema},
+    )
+    assert res.status_code == 200
+
+    res = await registry_async_client.get_subjects_subject_version(subject=subject, version="latest")
+    assert res.status_code == 200
+    assert res.json()["schemaType"] == "PROTOBUF"
+
+
+@pytest.mark.parametrize("registry_cluster", [{"config": {}}, {"config": {"use_protobuf_formatter": True}}], indirect=True)
 async def test_protobuf_schema_references(registry_async_client: Client) -> None:
     customer_schema = """
                 |syntax = "proto3";
--- a/tests/integration/test_schema_protobuf.py
+++ b/tests/integration/test_schema_protobuf.py
@@ -196,9 +196,25 @@ async def test_protobuf_schema_with_blank_documentation_lines(registry_async_cli
         | Some Important Event
         |
         | Overview
-        | What this event is for.
         |*/
+        |// Additional context
+        |//
+        |// What this event is for.
         |message Test {}
+        """
+    )
+    expected_schema = trim_margin(
+        f"""
+        |syntax = "proto3";
+        |
+        |// Some Important Event
+        |//{" "}
+        |// Overview
+        |// Additional context
+        |//{" "}
+        |// What this event is for.
+        |message Test {{}}
+        |
         """
     )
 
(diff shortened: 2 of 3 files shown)


roshini-aiven:
@Gezi-lzq  Nice fix — really clean root-cause diagnosis and good test coverage overall. A couple of small requests before merging:

Strengthen the integration test's assertion. test_protobuf_schema_with_blank_documentation_lines currently only checks status_code and schemaType on the GET. Since the bug is specifically about documentation content being mangled/rejected on replay, could we also assert the returned schema actually contains the expected text (e.g. assert "Overview" in res.json()["schema"])? That way a future regression that returns 200 but silently corrupts the comment wouldn't slip through.

Call out why the POST returning 200 matters in the PR description/testing section — it's the part that actually proves the schema-reader replay path is fixed, not just the follow-up GET. Worth a sentence so future readers don't mistake it for a plain smoke test.

Gezi-lzq:
@roshini-aiven  Thanks for the suggestions! Updated in `5ee0c2f9`:

- The integration test now asserts the exact canonical schema content, using mixed block and single-line comments with blank documentation lines.
- Added a note explaining that the POST returning 200 proves the schema-reader replay completed successfully.
- Updated the PR description with the same context.

The targeted integration test passes with both the default configuration and `use_protobuf_formatter=True` (`2 passed`).

Gezi-lzq:
Hi @roshini-aiven, I’ve updated the PR based on your feedback. Could you please take another look when you have a chance? Thanks!

Gezi-lzq:
Hi @roshini-aiven, just following up on this PR. The requested changes and targeted tests have been completed. Could you please take another look when you have time?

roshini-aiven:
Will take a look at it today , Apologies!

roshini-aiven:
looks good to me, I don't see the CI green, but yes @muralibasani, would this be up for merging,

muralibasani (approved):
LGTM, thanks for the pr @Gezi-lzq

PR #1360, opened 2026-08-21
feat(rest-proxy): support Kafka message headers on produce and consume

muralibasani:

Add optional per-record `headers` to the REST Proxy. 

- On produce, records may include a headers list of {"key": str, "value": base64-str|null} objects, which are decoded and forwarded to the Kafka producer. Values are decoded with validate=True, so non-base64 values fail fast with a 422.

- On consume, each record now returns a symmetric `headers` field. The field is omitted when the record has no headers, so the response shape is unchanged for existing consumers on header-less topics. Header values are base64-encoded since they are raw bytes.

- Never breaks consumers: header encoding can never fail a fetch — any malformed/unexpected header degrades to omitting the field and logging a warning, so existing and new consumers are unaffected regardless of who produced the record.

- Headers are opt-in and orthogonal to schema validation: value/key bytes and schema IDs are unchanged, so existing producers, consumers, and stored data are unaffected.

Integration tests cover produce/consume with headers — each side verified against a raw Kafka client, plus a full REST round-trip and malformed-header 422 cases. A unit test covers the consume-side header-encoding robustness (non-UTF-8 keys, encode failures).

Docs updated with examples: https://mbasani-rest-proxy-headers.karapace.pages.dev/docs/api-examples#rest-proxy

# Why this way

### DIFF
--- a/src/karapace/kafka_rest_apis/__init__.py
+++ b/src/karapace/kafka_rest_apis/__init__.py
@@ -1057,7 +1057,8 @@ class UserRestProxy:
             if key is not None:
                 key = await self.serialize(content_type, key, ser_format, key_schema_id)
             value = await self.serialize(content_type, value, ser_format, value_schema_id)
-            prepared_records.append((key, value, record.get("partition", default_partition)))
+            headers = self._decode_record_headers(record)
+            prepared_records.append((key, value, record.get("partition", default_partition), headers))
         return prepared_records
 
     async def get_partition_info(self, topic: str, partition: str, content_type: str) -> dict:
@@ -1161,12 +1162,13 @@ class UserRestProxy:
                     status=HTTPStatus.BAD_REQUEST,
                 )
             convert_to_int(r, "partition", content_type)
-            if set(r.keys()).difference({subject_type.value for subject_type in SubjectType}):
+            if set(r.keys()).difference({*(subject_type.value for subject_type in SubjectType), "headers"}):
                 KafkaRest.unprocessable_entity(
                     message="Invalid request format",
                     content_type=content_type,
                     sub_code=RESTErrorCodes.HTTP_UNPROCESSABLE_ENTITY.value,
                 )
+            self.validate_record_headers(r, content_type)
         # disallow missing id and schema for any key/value list that has at least one populated element
         if formats["embedded_format"] in {"avro", "jsonschema", "protobuf"}:
             for subject_type, code in zip(SUBJECT_VALID_POSTFIX, RECORD_CODES):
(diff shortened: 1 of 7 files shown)


src/karapace/kafka_rest_apis/__init__.py
```diff
@@ -1161,12 +1162,13 @@ async def validate_publish_request_format(self, data: dict, formats: dict, conte
                     status=HTTPStatus.BAD_REQUEST,
                 )
             convert_to_int(r, "partition", content_type)
-            if set(r.keys()).difference({subject_type.value for subject_type in SubjectType}):
+            if set(r.keys()).difference({*(subject_type.value for subject_type in SubjectType), "headers"}):
```
aindriu-aiven, line 1165: Should headers just be added to SubjectType?

src/karapace/kafka_rest_apis/__init__.py
```diff
@@ -1188,6 +1190,51 @@ async def validate_publish_request_format(self, data: dict, formats: dict, conte
...
+                    base64.b64decode(value, validate=True)
+                except (B64DecodeError, ValueError):
+                    KafkaRest.unprocessable_entity(
+                        message="header 'value' must be a base64-encoded string",
+                        content_type=content_type,
+                        sub_code=RESTErrorCodes.HTTP_UNPROCESSABLE_ENTITY.value,
+                    )
+
+    @staticmethod
+    def _decode_record_headers(record: dict) -> list[tuple[str, bytes | None]] | None:
```
aindriu-aiven, line 1230: Should this tuple be [str,str|None] as you are decoding the bytes?

src/karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -700,6 +700,36 @@ async def deserialize_strict(bytes_: bytes | None) -> dict | None:
 
         return deserialize_strict
 
+    @staticmethod
+    def _header_bytes(msg) -> int:
+        # Raw header size for the max_bytes budget; never raises so it can't break a fetch.
+        try:
+            return sum((len(k) if k else 0) + (len(v) if v else 0) for k, v in msg.headers() or [])
+        except Exception:
+            return 0
```
aindriu-aiven, line 709: We don't do anything with this exception what behaviour can we expect, does it stop returning the headers?

src/karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -700,6 +700,36 @@ async def deserialize_strict(bytes_: bytes | None) -> dict | None:
...
+
+    @staticmethod
+    def _encode_headers(msg) -> list[dict]:
```
aindriu-aiven, line 712: is it worth having a strict flag? basically the headers could be important and not worth sending the message without for some users, so have it the producer retry and fix the issue if they require headers?

src/karapace/kafka_rest_apis/consumer_manager.py
```diff
@@ -756,14 +789,19 @@ async def deserialize_message(msg):
...
                 "partition": msg.partition(),
                 "offset": msg.offset(),
                 "timestamp": msg.timestamp()[1] if msg.timestamp()[0] != Timestamp.NOT_AVAILABLE else None,
                 "key": key,
                 "value": value,
             }
+            # Only present when the record actually has headers, to keep header-less responses unchanged.
+            headers = self._encode_headers(msg)
+            if headers:
+                record["headers"] = headers
```
aindriu-aiven: this is probably a stylistic thing more then anything we can keep this here or we could add it to the return 
e.g.

``` 
encoded_headers = self._encode_headers(msg)
return {
                "topic": msg.topic(),
                "partition": msg.partition(),
                "offset": msg.offset(),
                "timestamp": msg.timestamp()[1] if msg.timestamp()[0] != Timestamp.NOT_AVAILABLE else None,
                "key": key,
                "value": value,
                **( {headers:  encoded_headers} if encoded_headers else {} ) 
            }
```

aindriu-aiven (commented):
Some really good stuff just a few questions.

src/karapace/kafka_rest_apis/__init__.py
muralibasani: No. Headers have no schema/subject, so adding headers to that enum is not correct imo. So they are left out.

src/karapace/kafka_rest_apis/__init__.py
muralibasani: No, because _decode_record_headers takes base64 str and decodes it to bytes as producer requires bytes.

src/karapace/kafka_rest_apis/consumer_manager.py
muralibasani: It does not stop returning headers. _header_bytes is used only for the max_bytes fetch loop through rest proxy.
headers are still encoded and returned separately by _encode_headers.

src/karapace/kafka_rest_apis/consumer_manager.py
muralibasani: I think the genuine strict place is on the produce side and we have enforced it with validate=true. On consume side, we just re-emit bytes that are always encodable. So prefer not to have a flag here.

src/karapace/kafka_rest_apis/consumer_manager.py
muralibasani: Agree, updated.

aindriu-aiven (approved):
LGTM

PR #1361, opened 2026-08-21
feature: add import mode

vtainio:

This changes adds an IMPORT mode to Karapace which allows users to migrate schemas from another system into Karapace while maintaining the existing schema IDs.

# Why this way

This change was implemented in a manner that is compatible with the APIs and migration patterns in the wider ecosystem. The import mode is just an API level changes and the schemas and contents that are stored within the internal Kafka topic remain the same as before.


src/karapace/core/schema_registry.py
```diff
@@ -392,6 +411,57 @@ async def write_new_schema_local(
...
+        self,
+        subject: Subject,
+        new_schema: ValidatedTypedSchema,
+        new_schema_references: Sequence[Reference] | None,
+        explicit_schema_id: SchemaId | None,
+        explicit_version: Version | None,
+    ) -> int:
+        """Register a schema in IMPORT mode, bypassing compatibility checks and allowing explicit id/version."""
+        all_schema_versions = self.database.find_subject_schemas(subject=subject, include_deleted=True)
+
```
muralibasani, line 424: Probably a rare case.

1. POST schema A with id=1, version=5 → creates v5 → id 1
2. POST schema B with id=2, version=5 → id check passes (id 2 is unused), v5 is silently repointed to id 2

id 1 → schema A still lives in the global self.schemas map, but subject s's version 5 no longer points to it

src/karapace/api/routers/requests.py
```diff
@@ -22,6 +22,8 @@ class SchemaRequest(BaseModel):
     references: list[SchemaReference] | None = None
     metadata: Any | None = None
     ruleSet: Any | None = None
+    schema_id: int | None = Field(alias="id", default=None)
```
muralibasani, line 25: Here id/version aren't range-validated.
Could we constrain at the request edge so it 422s before reaching the registry?

May be a couple of tests are better

src/karapace/api/controller.py
```diff
@@ -856,6 +872,14 @@ async def subject_post(
                         "message": str(exc),
                     },
                 ) from exc
+            except OperationNotPermittedInMode as exc:
```
muralibasani, line 875: I think we don't have a test for this subject post branch.

muralibasani (changes requested):
Overall looks good. Thanks.

- Can you add some documentation on the IMPORT mode contraints and behaviour in different situations (also in website/docs)
- how about consider rejecting same-content-different-id
- document that a failed import leaves partial state in IMPORT mode and how to resume, as there is no rollback right, and not sure how to recover.
- documentation : as schema ids are global and unique per registry and handling multiple sources (with same schema id 1 for ex in two sources) will fail.

PR #1369, opened 2026-09-03
Support OIDC (JWT) and basic auth side by side on the Schema Registry

muralibasani:

Support OIDC (JWT) and basic auth side by side on the Schema Registry

Currently when sasl_oauthbearer_authentication_enabled is on, basic auth doesn't work. The JWT check is Bearer-only and rejects anything else before basic auth runs. This is difficult for migrating customers.

Solution : 

With sasl_oauthbearer_authentication_enabled=true

| Header value      | Authentication                                              | Authorization                                                    |
|--------------------|--------------------------------------------------------------|--------------------------------------------------------------------|
| `Bearer <jwt>`     | OIDC JWT validation; invalid/expired → `401`, no fallback     | authz **on** → method-based roles; authz **off** → allowed (authn only) |
| `Basic <b64>`      | authfile credentials (requires `registry_authfile`)           | authfile ACL — always enforced, regardless of the OIDC authz flag  |
| missing / other    | —                                                              | `401`                                                               |

If sasl_oauthbearer_authentication_enabled is true and sasl_oauthbearer_authorization_enabled=false, and if its bearer token, we go with oidc validation and if invalid token, it fails. 

and if sasl_oauthbearer_authentication_enabled is true and sasl_oauthbearer_authorization_enabled=true, and bearer is sent, and if valid token, then authz is done based on method/roles.

if sasl_oauthbearer_authentication_enabled=true and sasl_oauthbearer_authorization_enabled=false or true, and if Basic is sent inheader instead of Bearer, then we go with basic authentication and basic authorization. 

if sasl_oauthbearer_authentication_enabled=false, everything basic authn and authz

New tests to simulate this fallback situation.

Update website docs

References: #xxxxx

# Why this way

### DIFF
--- a/src/karapace/api/middlewares/__init__.py
+++ b/src/karapace/api/middlewares/__init__.py
@@ -21,49 +21,63 @@ log = logging.getLogger(__name__)
 
 
 def _authenticate_and_authorize(request: Request, config: Config, oidc_validator: OIDCTokenValidator) -> JSONResponse | None:
-    """Run the OIDC auth gate. Return a JSONResponse on failure, or None to continue."""
+    """Dispatch on the Authorization scheme: Bearer -> OIDC check, Basic -> defer to basic auth.
+
+    Return a JSONResponse on failure, or None to continue (route runs; Basic is handled by the
+    basic-auth dependency). Bearer takes priority; there is no fallback between schemes.
+    """
     auth_header = request.headers.get("Authorization")
-    if not auth_header or not auth_header.startswith("Bearer "):
-        return JSONResponse(
-            status_code=401,
-            content={"error": "Unauthorized", "reason": "Missing or invalid Authorization header"},
-        )
+    # Auth schemes are case-insensitive; split "<scheme> <credentials>" and match on the lowered scheme.
+    scheme, _, credentials = (auth_header or "").partition(" ")
+    scheme = scheme.lower()
 
-    token = auth_header[len("Bearer ") :].strip()
-    if not token:
-        return JSONResponse(
-            status_code=401,
-            content={"error": "Unauthorized", "reason": "Missing or invalid Authorization header"},
-        )
-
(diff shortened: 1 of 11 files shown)


src/karapace/api/middlewares/__init__.py
```diff
@@ -21,49 +21,63 @@
...
+
+        # Expose only the configured subject claim to handlers, never the full payload.
+        # Prevents downstream code from picking up attacker-controlled claims (e.g. "roles")
+        # and using them for authz decisions.
+        request.state.user = payload.get(oidc_validator.claim_name) if oidc_validator.claim_name else None
+        # Marks the request as OIDC-authenticated so the basic-auth dependency is skipped.
+        request.state.oidc_authenticated = True
+        log.debug("Authenticated")
+
+        if config.sasl_oauthbearer_authorization_enabled:
```
aindriu-aiven, line 63: if config.sasl_oauthbearer_authorization_enabled is false, shouldn't we throw an exception, because someone is trying to authorize with an unknown/unauthorized token?

muralibasani: If config.sasl_oauthbearer_authorization_enabled is false which means config.sasl_oauthbearer_authentication_enabled could be true or false.

Assuming config.sasl_oauthbearer_authentication_enabled is true and config.sasl_oauthbearer_authorization_enabled is false

So we have 2 below cases

1. Header starts with Bearer : 
	- If invalid token, we deny with 401 unauthenticated already. Authorization flow is not touched even.
	- If valid token, then it authorizes fully as authz is set to false.


2. Header starts with Basic : 
	- If invalid credentials, we deny with 401 unauthenticated.
	- If valid credentials, authfile is validated for authorization. Results in 200 or deny error.

aindriu-aiven (approved):
LGTM

PR #1372, opened 2026-09-09
fix(rest-proxy): let registry credentials coexist with forwarded auth header

muralibasani:
1. **Let `registry_user`/`registry_password` coexist with a forwarded `Authorization`
     header.** Previously, whenever OIDC token forwarding was on, these had to be unset —
     otherwise the forwarded bearer/basic header collided with the proxy's session-level
     basic auth and aiohttp raised `Cannot combine AUTHORIZATION header with AUTH argument`,
     surfacing as a 500.

  2. **Surface SR auth failures as real 401/403** instead of the generic `40801`
     ("Error when registering schema"), which masked whether a failure was auth, schema
     incompatibility, or the registry being down.

 ## What changed

  **Per-request auth (fix 1)**
  - `core/client.py`: no longer bakes `session_auth` into the aiohttp `ClientSession`. A new
    `_resolve_auth` picks credentials per request — a forwarded `Authorization` header wins;
    otherwise fall back to `session_auth` (`registry_user`/`registry_password`).

  **Auth error propagation (fix 2)**
  - `kafka_rest_apis/error_codes.py`: add `HTTP_UNAUTHORIZED` (401), `HTTP_FORBIDDEN` (403).
  - `core/serialization.py`: `SchemaRetrievalError` carries the registry's HTTP `status_code`.
  - `kafka_rest_apis/__init__.py`: map SR 401/403 to a matching proxy response; the
    consume-by-id path no longer masks an auth failure as an invalid schema.

 ## Behavior matrix (SR OIDC + authfile)

  | Client → proxy | Sent to SR | Result |
  |---|---|---|
  | `Bearer <token>` | forwarded token | OIDC-validated |
  | `Basic <creds>` | forwarded basic | authfile-validated |
  | nothing | `registry_user`/`registry_password` | service-credential fallback |
  | invalid/expired token or creds | (as above) | `401`/`403` |

## Tests

  - Unit: `tests/unit/core/test_client.py` (`_resolve_auth` + on-the-wire selection);
    `tests/unit/kafka_rest_apis/test_registry_auth_error.py` (status propagation + 401/403 mapping).
  - E2E (`tests/e2e/kafka_rest_apis/test_oidc_forwarding.py`): coexistence (bearer + configured
    creds), forwarded-basic to an OIDC+authfile SR, and updated auth-rejection cases now
    asserting 401. Two new compose proxy services (`-oidc-with-creds`, `-oidc-basic`).

  ## Docs

  - `website/docs/authentication.md` and `README.rst`: coexistence and the 401/403 vs 40801/42205
    error behavior.


src/karapace/core/serialization.py
```diff
@@ -165,7 +169,7 @@ async def post_new_schema(
         headers = {"Content-Type": "application/vnd.schemaregistry.v1+json", **(_authorization_headers() or {})}
         result = await self.client.post(f"subjects/{quote(subject)}/versions", json=payload, headers=headers)
         if not result.ok:
-            raise SchemaRetrievalError(result.json())
+            raise SchemaRetrievalError(result.json(), status_code=result.status_code)
```
aindriu-aiven, line 172: This proxies back every result code, I'm wondering is their a scenario that we only want to proxy it back if it is a 401/403 or are we happy to allow everything to come back.

I'm mainly thinking of giving information to users they may not have previously had. Like something that exists or a username that is confirmed as existing but the password being wrong etc?

muralibasani: Good point. 

Actually we do not return all http status codes, but only 3.
unauthorized 401
forbidden 403
remaining all be returned as "schema retrieval error"

And earlier, only one which was "schema retrieval error". This was unclear on what the problem was with the request, if failed. 

This change is not related to the actual coexist behaviour, but while testing this feature, it was confusing as why a request failed.

It is not a breaking change primarily, however we are documenting this error handling behaviour which is an improved process.

tests/e2e/kafka_rest_apis/test_oidc_forwarding.py
```diff
@@ -7,9 +7,12 @@
 Topology (compose profile: e2e):
 - karapace-schema-registry-authn-only:8281  — OIDC SR (forwarding gate target).
 - karapace-rest-proxy-oidc:8382              — REST Proxy with the gate ON.
+- karapace-rest-proxy-oidc-with-creds:8582   — gate ON *and* registry_user/password set (coexistence).
 - karapace-rest-proxy-no-forward:8482        — REST Proxy with the gate OFF (regression guard).
 - karapace-schema-registry-basic:8581        — Basic-auth SR.
 - karapace-rest-proxy-basic:8682             — Basic-auth proxy paired with it (issue #1274 baseline).
+- karapace-schema-registry-oidc-basic:8681   — OIDC+authfile SR (scheme dispatch).
+- karapace-rest-proxy-oidc-basic:8782         — proxy that forwards a client's Basic header to it.
```
aindriu-aiven: Nit but I think this is just out of alignment.
```suggestion
- karapace-rest-proxy-oidc-basic:8782        — proxy that forwards a client's Basic header to it.
```

aindriu-aiven: as is line 8

src/karapace/core/serialization.py
muralibasani: Here are the error responses.

`{"error_code":401,"message":"Schema registry authentication failed; check the forwarded credentials or the configured registry_user/registry_password."}% `

`{"error_code":40801,"message":"Error when registering schema.format = AVRO, subject = avrotopic5-value"}%
`

aindriu-aiven (approved):
LGTM

PR #1374, opened 2026-09-17
feat import mode

muralibasani:
Adds an IMPORT mode to the Schema Registry so schemas can be migrated from another registry keeping their original schema IDs and version numbers.

Key features

1. IMPORT mode, global or per subject. 
2. Schema identity is preserved. POST /subjects/{subject}/versions accepts id and version in the body
3. New endpoints: 
- PUT /mode, 
- PUT /mode/{subject}, 
- DELETE /mode/{subject}, 

```
curl -X PUT  -d '{"mode": "IMPORT"}' localhost:8081/mode/my-subject

curl -X POST -d '{"schema": "...", "id": 1001, "version": 5}' \
  localhost:8081/subjects/my-subject/versions      # -> {"id":1001}

curl -X DELETE localhost:8081/mode/my-subject         # -> {"mode":"READWRITE"}

GET /mode/{subject} also works with ?defaultToGlobal=true/false
```

Subject has its own override. If there is no override, then global mode serves.
But for a non-existent subject, with ?defaultToGlobal=true returns global mode

Examples
```
curl -s "$SR/mode/no-such-subject"                        # 404 {"error_code":40401,"message":"Subject 'no-such-subject' not found."}
curl -s "$SR/mode/no-such-subject?defaultToGlobal=false"  # 404  (identical to omitting it)
curl -s "$SR/mode/no-such-subject?defaultToGlobal=true"   # 200 {"mode":"READWRITE"}

curl -s -X POST -H "$CT" -d '{"schema":"{\"type\":\"string\"}"}' $SR/subjects/exists/versions
curl -s "$SR/mode/exists"                                 # 200 {"mode":"READWRITE"}   <- global fallback, no flag needed
curl -s "$SR/mode/exists?defaultToGlobal=true"            # 200 {"mode":"READWRITE"}   <- same

curl -s -X PUT -H "$CT" -d '{"mode":"IMPORT"}' $SR/mode/imported
curl -s "$SR/mode/imported?defaultToGlobal=true"          # 200 {"mode":"IMPORT"}
```

4. Guardrails, so a bad import fails loudly rather than corrupting quietly. An ID is bound to its content permanently; a version slot can't be repointed; and the
presence of an id is what selects the mode — supplying one outside IMPORT, or omitting one inside it, is rejected rather than reinterpreted. Ten documented
rejections across 42202, 42204, 42205, 42207.

Ex. 
```
POST {"schema": A, "id": 1001, "version": 1}   -> 200 {"id":1001}
POST {"schema": B, "id": 1001, "version": 2}   -> 422 .. Overwrite new schema with id 128200200 is not permitted.

POST {"schema": A, "id": 100, "version": 5}    -> 200
POST {"schema": B, "id": 200, "version": 5}    -> 422 ... version 5 is already registered with schema id ..

POST {"schema": A, "id": 1001}     # subject is in READWRITE
  -> 422 {"error_code":42205,"message":"Subject 'demo-noimport' is not in IMPORT mode."}

POST {"schema": A}                 # subject is in IMPORT
-> 422 {"error_code":42205,"message":"Subject 'demo-noid' is in IMPORT mode,
          a schema id is required to register."}
 IMPORT deliberately skips compatibility checks.
```

For all unreachable versions, 422 ... The specified version '0' is not valid.

5. Two operator flags.

mode_mutability (default true) — a cluster-wide operator switch. Set false and every mode change is refused, so IMPORT can never be entered at runtime. Config-only, requires a restart.

allow_duplicate_schema_ids (default true) false refuses an import whose content already has a different ID, keeping IDs one-to-one with content

6. Modes are MODE records in _schemas topic, so they survive a restart.
subject: None means the global mode.

7. Compatibility

id/version on POST /subjects/{subject}/versions were previously parsed and ignored; they're now honoured or rejected.

8. Testing
  
1218 unit tests & 34 integration tests

Updated docs
https://mbasani-feat-import-mode.karapace.pages.dev/docs/import-mode

--------

Local tests

```
IMPORT mode demo run 20260918-140209
  kafka    : localhost:9092
  registry : http://localhost:8081
  topic    : _schemas_import_demo_20260918-140209  (fresh, so global IMPORT starts empty)
  logs     : ./.runs/20260918-140209

── global mode
   11 checks passed
── scenarios
   100 checks passed
── restart replay (seeding)
   seeded
── restart replay
   7 checks passed
── mode_mutability=false
   8 checks passed
── allow_duplicate_schema_ids=false
   9 checks passed

================================================================
### IMPORT mode verified against a live registry

| Phase | Checks | Result |
| --- | --- | --- |
| global mode | 11 | pass |
| scenarios | 100 | pass |
| restart replay | 7 | pass |
| mode_mutability=false | 8 | pass |
| allow_duplicate_schema_ids=false | 9 | pass |

Total: 135 checks. Kafka localhost:9092, registry on a fresh topic per run.

Covered: id and version preserved on import; idempotent replay; an id refused when
already bound to other content; the same content allowed under a second id; a version
slot refused when taken; importing over a populated subject with force; several
pre-existing versions; resuming a part done batch; hard delete dropping the mode;
explicit id refused outside IMPORT and required inside it; id and version range checks;
compatibility skipped inside and enforced again after; soft deleted versions; ids shared
across subjects; JSON Schema and Protobuf including normalize; references; global IMPORT
and subject level override; MODE records replayed from Kafka after a restart;
mode_mutability=false refusing every mode change; and
allow_duplicate_schema_ids=false refusing one schema under a second id.
```

References: #xxxxx

# Why this way


src/karapace/core/schema_registry.py
```diff
@@ -392,6 +417,63 @@ async def write_new_schema_local(
...
+        """Register a schema in IMPORT mode, bypassing compatibility checks, with an explicit id."""
+        # include_deleted: a soft deleted version still owns its version slot and its id binding.
+        all_schema_versions = self.database.find_subject_schemas(subject=subject, include_deleted=True)
+
+        schema_id = explicit_schema_id
+        # An id is bound to its content for good, so it cannot be rebound. Re-importing the
+        # same content under it stays allowed, for replays and for ids shared across subjects.
+        bound_schema = self.database.find_schema(schema_id=schema_id)
+        if bound_schema is not None and bound_schema != new_schema:
+            raise OperationNotPermittedInMode(f"Overwrite new schema with id {schema_id} is not permitted.")
```
aindriu-aiven: I think this particular error would be better as a 409 as it is a conflict on the schemaId ?

WDYT?

src/karapace/core/schema_registry.py
```diff
@@ -392,6 +417,63 @@ async def write_new_schema_local(
...
+            version = explicit_version
+        elif not all_schema_versions:
+            version = Version(1)
+        else:
+            version = self.database.get_next_version(subject=subject)
+
+        occupied = all_schema_versions.get(version)
+        if occupied is not None and (occupied.schema_id != schema_id or occupied.schema != new_schema):
+            # Overwriting the slot would repoint the version and orphan its schema.
+            raise OperationNotPermittedInMode(
```
aindriu-aiven: this could also be a 409

src/karapace/core/schema_registry.py
```diff
@@ -406,10 +488,54 @@ def get_subject_versions_for_schema(
...
+        async with self.schema_lock:
+            if validated_mode == Mode.import_mode and not force:
+                # An empty target keeps imported ids and versions from colliding. Soft deleted
+                # schemas do not count. `force` skips this check and deletes nothing.
+                if subject is not None:
+                    existing = bool(self.database.find_subject_schemas(subject=subject, include_deleted=False))
+                else:
+                    existing = bool(self.database.find_subjects(include_deleted=False))
+                if existing:
+                    raise OperationNotPermittedInMode("Cannot import since found existing subjects")
```
aindriu-aiven: possibly this one as a 409 as well

src/karapace/api/routers/mode.py
```diff
@@ -38,16 +40,94 @@ async def mode_get(
...
+    user: Annotated[User, Depends(get_current_user)],
+    schema_registry: KarapaceSchemaRegistry = Depends(Provide[SchemaRegistryContainer.schema_registry]),
+    forward_client: ForwardClient = Depends(Provide[SchemaRegistryContainer.karapace_container.forward_client]),
+    authorizer: AuthenticatorAndAuthorizer = Depends(Provide[AuthContainer.authorizer]),
+    controller: KarapaceSchemaRegistryController = Depends(Provide[SchemaRegistryContainer.schema_registry_controller]),
+    force: bool = False,
+) -> ModeResponse:
+    subject = Subject(unquote_plus(subject))
+    if authorizer and not authorizer.check_authorization(user, Operation.Write, f"Subject:{subject}"):
+        raise subject_not_found(subject)
```
aindriu-aiven: this should probably be put into its own function it gets called a lot in this class

src/karapace/api/routers/requests.py
```diff
@@ -34,6 +44,27 @@ def validate_schema(cls, schema_str: str) -> str:
             )
         return schema_str
 
+    # mode="before" so non-integer input also gets a Karapace shaped error body.
+    @field_validator("schema_id", mode="before")
```
aindriu-aiven, line 48: Sorry can you explain the mode=before?
Is this validating the inputs before executing the method or something like that?

src/karapace/core/schema_registry.py
muralibasani: Done. There was no 409 http code at all in codebase. Added now.

src/karapace/api/routers/mode.py
muralibasani: Done. Good one.

src/karapace/core/client.py
```diff
@@ -248,8 +248,26 @@ async def delete_config_subject(self, *, subject: str) -> Result:
     async def get_mode(self) -> Result:
         return await self.get(path="/mode")
 
-    async def get_mode_subject(self, *, subject: str) -> Result:
-        return await self.get(path=f"/mode/{quote_plus(subject)}")
+    async def put_mode(self, *, json: JsonData, force: bool | None = None) -> Result:
+        path = "/mode"
+        if force is not None:
+            path = f"{path}?force={str(force).lower()}"
```
aindriu-aiven, line 254: do you need to check if force is true or false or will the api return/ignore this if incorrect?

src/karapace/api/routers/requests.py
muralibasani: Yes mode `before` runs the validator on the raw JSON value before pydantic coerces it to the field's declared int | None

src/karapace/core/config.py
```diff
@@ -186,6 +186,10 @@ class Config(BaseSettings):
     kafka_retriable_errors_silenced: bool = True
     use_protobuf_formatter: bool = False
     waiting_time_before_acting_as_master_ms: int = 5000
+    # Set false to reject every mode change.
+    mode_mutability: bool = True
+    # Set false to keep schema ids one to one with content during an import.
+    allow_duplicate_schema_ids: bool = True
```
aindriu-aiven, line 192: how do duplicates work, like does it allow an existing id to be saved while adding a new one with the same id or does it allow multiple schemas to be imported while allowing them all to keep their schema ids?

src/karapace/core/in_memory_database.py
```diff
@@ -275,8 +301,27 @@ def set_subject_compatibility(self, *, subject: Subject, compatibility: str) ->
...
+        if subject in self.subjects:
+            self.subjects[subject].mode = mode
+
+    def delete_subject_mode(self, *, subject: Subject) -> None:
+        if subject in self.subjects:
+            self.subjects[subject].mode = None
+
     def find_schema(self, *, schema_id: SchemaId) -> TypedSchema | None:
-        return self.schemas[schema_id]
+        return self.schemas.get(schema_id)
```
aindriu-aiven, line 324: Should we do get or default here and return something expected if the schema id is not found?

README.rst
```diff
@@ -586,6 +622,14 @@ Keys to take special care are the ones needed to configure Kafka and advertised_
    * - ``topic_name``
      - ``_schemas``
      - The name of the Kafka topic where to store the schemas.
+   * - ``allow_duplicate_schema_ids``
+     - ``true``
+     - Whether one schema may be imported under more than one schema id. Set to ``false`` to refuse an import whose content is already
```
aindriu-aiven, line 627: ok this clears up the duplicate schema ids nice feature

src/karapace/core/client.py
muralibasani: Yes here https://github.com/Aiven-Open/karapace/pull/1374/changes#diff-31a4fe49026bcd80632cf996d7a8edaad295abff495583631071882deff111cbR518 there is a check which validates the bool value.

And ?force=garbage is rejected by FastAPI

src/karapace/core/in_memory_database.py
muralibasani: It actually returns None as default. The default comes from dict.get.
It used to be return self.schemas[schema_id].

website/docs/import-mode.md
```diff
@@ -0,0 +1,550 @@
...
+```bash
+SR=http://localhost:8081
+CT='Content-Type: application/vnd.schemaregistry.v1+json'
+```
+
+A migration into one subject is three requests:
+
+```bash
+# 1. put the subject into import mode. It does not have to exist yet
+curl -s -X PUT -H "$CT" -d '{"mode":"IMPORT"}' $SR/mode/my-subject
```
aindriu-aiven, line 34: We should consider a bash script (in the future not here) to handle changing modes.

Also potentially to read from one schema reg and import to another.

website/docs/import-mode.md
```diff
@@ -0,0 +1,550 @@
...
+# {"mode":"IMPORT"}
+
+# 2. register each schema with the id and version it had in the source registry
+curl -s -X POST -H "$CT" \
+  -d '{"schema":"{\"type\":\"string\"}","id":1001,"version":5}' \
+  $SR/subjects/my-subject/versions
+# {"id":1001}
+
+# 3. return the subject to normal operation
+curl -s -X DELETE $SR/mode/my-subject
```
aindriu-aiven, line 44: I'm a little surprised that it isn't 
`curl -s -X PUT -H "$CT" -d '{"mode":"READWRITE"}' $SR/mode/my-subject`

website/docs/import-mode.md
```diff
@@ -0,0 +1,550 @@
...
+## Quick start
+
+The examples below use two shell variables:
+
+```bash
+SR=http://localhost:8081
+CT='Content-Type: application/vnd.schemaregistry.v1+json'
+```
+
+A migration into one subject is three requests:
```
aindriu-aiven, line 32: Ok I get that this is removing the subject level mode and thats why we have the delete on Line 44.
Maybe we could alter slightly the below or something similar?
```suggestion
# Subject Level Import
A migration into one subject is three requests:
```

website/docs/import-mode.md
```diff
@@ -0,0 +1,550 @@
...
+# {"id":1001}
+```
+
+Compared to `READWRITE`, import mode changes four things:
+
+- Content-based deduplication is bypassed, so an identical schema does not collapse onto an
+  existing registration.
+- Version auto-increment is bypassed, so version numbers may have gaps and need not start
+  at 1.
+- Compatibility checks are skipped entirely. The subject's compatibility level is ignored for
```
aindriu-aiven, line 212: Given the subject or the registry depending needs to be empty to use import mode, compatibility shouldn't be an issue anyway right?

tests/integration/test_schema_registry_mode.py
```diff
@@ -46,3 +125,809 @@ async def test_subject_mode(registry_async_client: Client) -> None:
...
+    original_id = res.json()["id"]
+
+    res = await registry_async_client.put_mode_subject(subject=subject, json={"mode": "IMPORT"}, force=True)
+    assert res.status_code == 200
+    assert res.json() == {"mode": str(Mode.import_mode)}
+
+    res = await registry_async_client.get_subjects_versions(subject=subject)
+    assert res.json() == [1]
+    res = await registry_async_client.get_subjects_subject_version(subject=subject, version=1)
+    assert res.json()["id"] == original_id
```
aindriu-aiven, line 359: i think you need to assert 

```suggestion
    assert res.json()["id"] == original_id
    assert json.loads(res.json()["schema"]) == json.loads(schema)
```

tests/unit/test_import_mode.py
```diff
@@ -0,0 +1,1134 @@
...
+        reader, db = self._make_reader()
+        db.insert_subject(subject=Subject("my-topic"))
+        db.set_subject_mode(subject=Subject("my-topic"), mode=Mode.import_mode)
+        KafkaSchemaReader._handle_msg_mode(reader, {"keytype": "MODE", "subject": "my-topic", "magic": 0}, None)
+        assert db.get_subject_mode(subject=Subject("my-topic")) is None
+
+    def test_auto_inserts_subject_if_missing(self) -> None:
+        """A MODE message for an unknown subject should create it."""
+        reader, db = self._make_reader()
+        KafkaSchemaReader._handle_msg_mode(
```
aindriu-aiven, line 127: I think the db is shared in these tests so you might need to assert it doesnt exist first here.

aindriu-aiven (commented):
Looks really good I think I had a few small comments but nothing too major.

website/docs/import-mode.md
muralibasani: Deleting will also put a tombstone, which falls back to global -> READWRITE.
However, updated the docs to put READWRITE, and also documenting what DELETE does.

tests/unit/test_import_mode.py
muralibasani: Actually it is not shared across test, but added an assert for that pre-condition assert.

website/docs/import-mode.md
muralibasani: Updated docs.  comp check compares against whatever was imported earlier in the same session

muralibasani:
> Looks really good I think I had a few small comments but nothing too major.

Thank you for the detailed review. Pushed a few changes.

